PR #34, opened 2025-02-12
feat: implement opentelemetry tracing subscriber

zsoltkacsandi:


data-plane/python-bindings/src/lib.rs
```diff
@@ -127,14 +132,18 @@ impl PyService {
...
+        let otel_guard = agp_tracing::TracingConfiguration::default()
+            .with_log_level(log_level)
+            .setup_tracing_subscriber();
+
+        let mut global_guard = TRACING_GUARD.lock().unwrap();
+        *global_guard = Some(otel_guard);
+
+        Ok(())
+    });
 }
```
msardara: This was an issue also before, but since we are here let's fix it.

This function is not idempotent, meaning if called twice, the second time [it'll panic](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/util/trait.SubscriberInitExt.html#method.init).

We can use something like [OnceCell](https://docs.rs/tokio/latest/tokio/sync/struct.OnceCell.html) to prevent this:

```rust
static TRACING_GUARD: OnceCell<agp_tracing::OtelGuard> = OnceCell::const_new();

async fn init_tracing_impl(log_level: String) {
    let _ = TRACING_GUARD.get_or_init(|| async {
        let otel_guard = agp_tracing::TracingConfiguration::default()
            .with_log_level(log_level)
            .setup_tracing_subscriber();

        otel_guard
    }).await;
}

#[pyfunction]
#[pyo3(signature = (log_level="info".to_string(),))]
fn init_tracing(py: Python, log_level: String) {
    let _ = pyo3_async_runtimes::tokio::future_into_py(py, async move {
        Ok(init_tracing_impl(log_level).await)
    });
}
```

msardara (approved):
Neat, LGTM!

Time allowing we could add an integration test in CSIT and get the exported traces, but this is a plus.

Only little detail, as we are modifying the tracing let's make the function resilient if called more than once by the python bindings.

PR #64, opened 2025-02-24
refactor: add gateway config struct for Python bindings

zsoltkacsandi:

### DIFF
--- a/data-plane/rust-tooling/toolchain.yml
+++ b/data-plane/rust-tooling/toolchain.yml
@@ -92,9 +92,10 @@ tasks:
     desc: "Install the components"
     internal: true
     cmds:
-      - cargo install {{.COMPONENTS}} --locked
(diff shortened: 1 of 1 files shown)


data-plane/python-bindings/src/lib.rs
```diff
@@ -29,6 +29,204 @@ use agp_service::{Service, ServiceError};
...
+    tls_ca_path: Option<String>,
+    tls_ca_pem: Option<String>,
+    tls_cert_path: Option<String>,
+    tls_key_path: Option<String>,
+    tls_cert_pem: Option<String>,
+    tls_key_pem: Option<String>,
+    basic_auth_username: Option<String>,
+    basic_auth_password: Option<String>,
+}
+
```
micpapal, line 70: For this struct I think we should have also getters and setters. The attribute     #[pyo3(get, set)] should be enough.

PR #66, opened 2025-02-25
feat: add metrics

zsoltkacsandi:

### DIFF
--- a/.github/actions/setup-rust/action.yaml
+++ b/.github/actions/setup-rust/action.yaml
@@ -39,6 +39,7 @@ runs:
       if: runner.os == 'Linux'
       shell: bash
       run: |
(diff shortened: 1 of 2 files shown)


data-plane/gateway/datapath/src/message_processing.rs
```diff
@@ -630,6 +652,7 @@ impl PubSubService for MessageProcessor {
             connection.remote_addr(),
             connection.local_addr()
         );
+        info!(counter.num_active_connections = 1);
```
micpapal: I think that this +1 should go also in the register_local_connection that creates the connection between the app and the local forwarder. There in fact there is a call to process_stream and closing that connection you may go at -1 in the counter

PR #72, opened 2025-03-06
feat: notify local app if a message is not processed correctly

micpapal:

### DIFF
--- a/data-plane/gateway/datapath/src/message_processing.rs
+++ b/data-plane/gateway/datapath/src/message_processing.rs
@@ -737,21 +737,8 @@ impl MessageProcessor {
                                             }
                                         } else {
                                             error!("error receiving messages {:?}", e);
-                                            let connection = self_clone.forwarder().get_connection(conn_index);
-                                            match connection {
-                                                Some(conn) => {
-                                                    debug!("try to notify the error to the remote end");
-                                                    if let Channel::Server(tx) = conn.channel() {
-                                                        if tx.send(Err(e)).await.is_err() {
-                                                            debug!("unable to notify the error to the remote end");
-                                                        }
-                                                    }
-                                                }
-                                                None => {
-                                                    error!("connection {:?} not found", conn_index);
-                                                }
-                                            }
                                         }
+                                        break;
                                     }
                                 }
                             }


data-plane/gateway/datapath/src/message_processing.rs
```diff
@@ -716,16 +683,80 @@ impl MessageProcessor {
...
+                                                                &msg_source.unwrap(),
+                                                                &dest.agent_class,
+                                                                Some(dest.agent_id),
+                                                                HashMap::new(), 1, "",
+                                                                Vec::new());
+
+                                                            err_message.metadata.insert(MetadataType::Error.to_string(), e.to_string());
+                                                            if let Channel::Server(tx) = conn.channel() {
+                                                                if tx.send(Ok(err_message)).await.is_err() {
+                                                                    debug!("Unable to notify the error to the remote end");
```
msardara:
```suggestion
                                                                    debug!("unable to notify the error to local app");
```

data-plane/gateway/datapath/src/message_processing.rs
```diff
@@ -716,16 +683,80 @@ impl MessageProcessor {
...
+                                        if let Err(e) = self_clone.handle_new_message(conn_index, msg).await {
+                                            error!("error processing incoming messages {:?}", e);
+                                            // If the message is coming from a local app, notify it
+                                            if is_local {
+                                                let connection = self_clone.forwarder().get_connection(conn_index);
+                                                match connection {
+                                                    Some(conn) => {
+                                                        debug!("try to notify local application");
+                                                        if msg_source.is_none() || msg_name.is_none() {
+                                                            debug!("Unable to notify the error to the remote end");
```
msardara:
```suggestion
                                                            debug!("unable to notify the error to local app");
```

data-plane/gateway/datapath/src/message_processing.rs
```diff
@@ -716,16 +683,80 @@ impl MessageProcessor {
...
+                                                        error!("connection {:?} not found", conn_index);
+                                                    }
+                                                }
+                                            }
+                                        }
+                                    }
+                                    Err(e) => {
+                                        if let Some(io_err) = MessageProcessor::match_for_io_error(&e) {
+                                            if io_err.kind() == std::io::ErrorKind::BrokenPipe {
+                                                info!("Connection {:?} closed by peer", conn_index);
```
msardara:
```suggestion
                                                info!("connection {:?} closed by peer", conn_index);
```

data-plane/gateway/datapath/src/message_processing.rs
```diff
@@ -716,16 +683,80 @@ impl MessageProcessor {
...
+                            Some(result) => {
+                                match result {
+                                    Ok(msg) => {
+                                        // save message source to use in case of error
+                                        let mut msg_source = None;
+                                        let mut msg_name = None;
+                                        if is_local {
+                                            msg_source = get_source(&msg);
+                                            msg_name = get_name(&msg);
+                                        }
```
msardara, line 697: I think we can define msg_source and msg_name just in case of error. WDYT?

data-plane/gateway/datapath/src/message_processing.rs
```diff
@@ -716,16 +683,80 @@ impl MessageProcessor {
...
+                                        if let Some(io_err) = MessageProcessor::match_for_io_error(&e) {
+                                            if io_err.kind() == std::io::ErrorKind::BrokenPipe {
+                                                info!("Connection {:?} closed by peer", conn_index);
+                                            }
+                                        } else {
+                                            error!("error receiving messages {:?}", e);
+                                            let connection = self_clone.forwarder().get_connection(conn_index);
+                                            match connection {
+                                                Some(conn) => {
+                                                    debug!("try to notify the error to the remote end");
```
msardara: Wouldn't this trigger a ping-pong of the error between the 2 ends?

data-plane/gateway/datapath/src/message_processing.rs
micpapal: This is not an error for the local application but rather an error occurring upon reception; it originates from the stream itself. The broken pipe error is treated as a special case to detect the end of a connection. In other instances, we attempt to inform the other side that an unexpected event has occurred.

data-plane/gateway/datapath/src/message_processing.rs
micpapal: yes. the problem is that message is moved into handle_new_message and in case of error the message is not accessible anymore

data-plane/gateway/datapath/src/message_processing.rs
micpapal: in this case the remote end is always the service of the app, so this should not happen

micpapal: in case of messages coming from remote this was a problem. now we drop the connection in case of errors

PR #88, opened 2025-03-17
feat: new message format

micpapal:


data-plane/gateway/datapath/pubsub/proto/v1/pubsub.proto
```diff
@@ -9,39 +9,78 @@ service PubSubService {
...
-  optional uint64 id = 2;
+// Service.id meaning according to ControlType
+// CTRL_FNF = nonce
+// CTRL_REQUEST = nonce
+// CTRL_REPLY = nonce of the associated CTRL_REQUEST
+// CTRL_STREAM = sequential incremental ID
+// CTRL_RTX_REQUEST = nonce
+// CTRL_RTX_REPLY = nonce of the associated CTRL_RTX_REQUEST
+
+// Srervice.stream meaning according to ControlType
```
msardara:
```suggestion
// Service.stream meaning according to ControlType
```

data-plane/gateway/datapath/pubsub/proto/v1/pubsub.proto
```diff
@@ -9,39 +9,78 @@ service PubSubService {
...
+// CTRL_STREAM = stream for which the id field is valid
+// CTRL_RTX_REQUEST = stream for this CTRL_RTX_REQUEST
+// CTRL_RTX_REPLY = stream for this CTRL_RTX_REPLY
+// None in all the other cases
+
+// Srervice.rtx meaning according to ControlType
```
msardara:
```suggestion
// Service.rtx meaning according to ControlType
```

data-plane/gateway/datapath/pubsub/proto/v1/pubsub.proto
```diff
@@ -9,39 +9,78 @@ service PubSubService {
...
-  AgentId source = 1;
-  AgentId name = 2;
+  AGPHeader header = 1;
+  ServiceHeader control = 2;
   uint32 fanout = 3;
   Content msg = 4;
 }
 
-message AgentGroup {
+// recvFrom = connection form where the sub/ubsub is supposed to be received
```
msardara:
```suggestion
// recvFrom = connection from where the sub/ubsub is supposed to be received
```

data-plane/gateway/datapath/src/message_processing.rs
```diff
@@ -239,9 +241,15 @@ impl MessageProcessor {
...
-        msg: Message,
+        mut msg: Message,
         out_conn: u64,
     ) -> Result<(), Box<dyn std::error::Error>> {
+        // clear header
+        let err = clear_agp_header(&mut msg);
+        if err.is_err() {
+            error!("an error occurred while cleaning the AGP header");
+        }
+
```
msardara: What do you think of:

```suggestion
        clear_agp_header(&mut msg)?;
```

And then we log errors in a single point in the caller? At the moment we print the error in each place it can occur, but it is probably better to print the error once in the caller.

micpapal: actually I notice the error is not return here, just printed. I will fix it and return a proper error

PR #90, opened 2025-03-17
feat: propagate context to enable distributed tracing

zsoltkacsandi:

### DIFF
--- a/data-plane/python-bindings/examples/agent.py
+++ b/data-plane/python-bindings/examples/agent.py
@@ -26,9 +26,9 @@ def format_message(message1, message2):
     return f"{color.BOLD}{color.CYAN}{message1.capitalize()}{color.END}\t {message2}"
 
 
-async def run_client(local_id, remote_id, message, address, iterations):
+async def run_client(local_id, remote_id, message, address, iterations, enable_opentelemetry: bool):
     # init tracing
-    agp_bindings.init_tracing(log_level="debug", enable_opentelemetry=True)
+    agp_bindings.init_tracing(log_level="debug", enable_opentelemetry=enable_opentelemetry)
 
     # Split the IDs into their respective components
     try:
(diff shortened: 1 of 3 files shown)


data-plane/python-bindings/examples/agent.py
```diff
@@ -28,7 +28,7 @@ def format_message(message1, message2):
 
 async def run_client(local_id, remote_id, message, address, iterations):
     # init tracing
```
micpapal, line 30: can you add here another flag to enable/disable OTel? Now it is always ON by default

data-plane/python-bindings/examples/gateway.py
```diff
@@ -14,7 +14,7 @@
 
 async def run_server(address: str):
     # init tracing
-    agp_bindings.init_tracing(log_level="debug")
+    agp_bindings.init_tracing(log_level="debug", enable_opentelemetry=True)
```
micpapal: same as for the client. it would be nice to have a flag to enable/disable OTel from the command

data-plane/gateway/datapath/src/message_processing.rs
```diff
@@ -239,16 +286,62 @@ impl MessageProcessor {
...
+                    let span = tracing::span!(
+                        tracing::Level::DEBUG,
+                        "send_message_to_local",
+                        instance_id = %INSTANCE_ID.as_str(),
+                        connection_id = out_conn,
+                        message_type = match &msg.message_type {
+                            Some(PublishType(_)) => "publish",
+                            Some(SubscribeType(_)) => "subscribe",
+                            Some(UnsubscribeType(_)) => "unsubscribe",
+                            None => "unknown"
```
msardara: Let's create a helper function for this match

data-plane/python-bindings/examples/agent.py
zsoltkacsandi: Sure. I've added it.

data-plane/gateway/datapath/src/message_processing.rs
zsoltkacsandi: Done.

PR #97, opened 2025-03-18
feat(tables): do not require Default/Clone traits for elements stored in pool

msardara:
## Motivation

At the moment, the pool implementation in the gateway requires the underlying contained type to implement
the `Default` and `Clone` traits. While this is acceptable in general, it does not work well with types like Box,
that cannot be cloned and mey not implement the Default trait.

## Solution

Enable any type to be stored in the pool data structure.

### DIFF
--- a/data-plane/gateway/datapath/src/tables/pool.rs
+++ b/data-plane/gateway/datapath/src/tables/pool.rs
@@ -85,7 +85,8 @@ impl<T> Pool<T> {
         // If length is equal to capacity, resize the pool
         if self.len == self.capacity {
             // Resize the pool
-            self.pool.resize_with(2 * self.capacity, || MaybeUninit::uninit());
+            self.pool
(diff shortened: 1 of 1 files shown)


data-plane/gateway/datapath/src/tables/pool.rs
```diff
@@ -98,51 +101,72 @@ where
...
+            self.insert_at(element, index)
+                .then_some(true)
+                .expect("insert_at failed");
 
-            if index > self.max_set {
-                self.max_set = index;
-            }
-
-            Some(index)
+            index
```
micpapal, line 110: the update for max_set is missing. in case of resize is it possible that index > max_set

msardara: The update should have been done in insert_at at this point I think 🤔

PR #106, opened 2025-03-19
feat(data-plane/service): first draft of session layer

msardara:
BREAKING CHANGE: This PR breaks the python binding interface.

## Motivation

We need a reliability layer that is able to handle message recovery in case the message sent across
multiple gateways is lost.

## Solution

This PRs provides a first implementation of the session layer and the fire and forget delivery, that does not ensure any reliability.

### DIFF
--- a/data-plane/gateway/service/src/fire_and_forget.rs
+++ b/data-plane/gateway/service/src/fire_and_forget.rs
@@ -4,6 +4,7 @@
 use crate::session::SessionType;
 use crate::session::{Common, Error, Id, Info, MessageDirection, Session, SessionDirection, State};
 
+use rand::Rng;
 use tonic::Status;
 
 use agp_datapath::messages::utils;
@@ -73,6 +74,10 @@ impl Session for FireAndForget {
                 })
             }
             MessageDirection::South => {
+                // add a nonce to the message
+                let header = utils::get_session_header_as_mut(&mut message).unwrap();
+                header.message_id = rand::rng().random();
+
                 let tx = self.common.tx_gw();
                 Box::pin(async move {
                     tx.send(Ok(message))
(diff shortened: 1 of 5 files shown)


data-plane/gateway/service/src/session_layer.rs
```diff
@@ -0,0 +1,353 @@
...
+    async fn handle_message_from_app(
+        &self,
+        mut message: Message,
+        direction: MessageDirection,
+        session_id: Option<Id>,
+    ) -> Result<(), Error> {
+        // if the session ID is not specified, return an error
+        if session_id.is_none() {
+            return Err(Error::MissingSessionId("None".to_string()));
+        }
```
micpapal: session_id is mandatory, remove the option

data-plane/gateway/service/src/session_layer.rs
```diff
@@ -0,0 +1,353 @@
...
+        if let Some(session) = self.pool.read().await.get(&id) {
+            // pass the message to the session
+            let ret = session.on_message(message, direction).await;
+            return ret;
+        }
+
+        // if the session is not found and the direction is North, create a new session
+        if direction != MessageDirection::North {
+            return Err(Error::SessionNotFound(id.to_string()));
+        }
```
micpapal: This check can be removed, this error should never happen.

data-plane/gateway/service/src/fire_and_forget.rs
```diff
@@ -0,0 +1,137 @@
...
+
+                let tx = self.common.tx_app();
+                Box::pin(async move {
+                    tx.send((message, info))
+                        .await
+                        .map_err(|e| Error::AppTransmission(e.to_string()))
+                })
+            }
+            MessageDirection::South => {
+                let tx = self.common.tx_gw();
```
micpapal, line 81: add a nonce as message id

PR #117, opened 2025-03-24
feat: add timers for rtx

micpapal:

### DIFF
--- a/data-plane/gateway/service/src/timer.rs
+++ b/data-plane/gateway/service/src/timer.rs
@@ -32,7 +32,7 @@ impl Timer {
         }
     }
 
-    async fn start<T: TimerObserver + Send + Sync + 'static>(&self, observer: Arc<T>) {
+    fn start<T: TimerObserver + Send + Sync + 'static>(&self, observer: Arc<T>) {
         let timer_id = self.timer_id;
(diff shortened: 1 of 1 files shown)


data-plane/gateway/service/src/timer.rs
```diff
@@ -0,0 +1,224 @@
...
+    fn new(timer_id: u32, duration: u32, max_retries: u32) -> Self {
+        Timer {
+            timer_id,
+            duration,
+            max_retries,
+            cancellation_token: CancellationToken::new(),
+        }
+    }
+
+    async fn start<T: TimerObserver + Send + Sync + 'static>(&self, observer: Arc<T>) {
```
msardara: This does not need to be async. Any reason to have it async?

data-plane/gateway/service/src/timer.rs
```diff
@@ -0,0 +1,224 @@
...
+                    _ = cancellation_token.cancelled() => {
+                        observer.on_stop(timer_id).await;
+                        break;
+                    },
+                }
+            }
+        });
+    }
+
+    async fn stop(&self) {
```
msardara: This can be not async as well.

PR #124, opened 2025-03-28
feat: request/reply session type

msardara:
## Motivation

Request/Response session type.

This is a first draft and requires extensive testing. Unit tests will be added in a second PR.

## Solution

Introduce the Request/Response session type:
 - in a request/reply session, if a reply is not received before the timeout, an error is returned to the application
 - application can now receive either messages or errors: an example of an error is a timeout on a request for which a reply was not received before the timeout
 - multiple messages can be sent in parallel - multiple timers will be set, one for each request
 - for each timeout that triggers, an error is sent to the application

### DIFF
--- a/data-plane/gateway/service/src/request_response.rs
+++ b/data-plane/gateway/service/src/request_response.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::collections::HashMap;
+use std::num::TryFromIntError;
 use std::sync::Arc;
 
 use async_trait::async_trait;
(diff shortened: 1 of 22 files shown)


data-plane/gateway/service/src/request_response.rs
```diff
@@ -0,0 +1,349 @@
...
+
+    pub(crate) async fn send_message_with_timer(
+        &self,
+        message: SessionMessage,
+    ) -> Result<(), SessionError> {
+        // get message id
+        let message_id = message.info.message_id.expect("message id not found");
+
+        // create new timer
+        let timer = timer::Timer::new(message_id, 1000, 0);
```
micpapal: use the duration in RequestResponseConfiguration

micpapal (approved):

PR #125, opened 2025-03-28
perf: improve connection pool performance

zsoltkacsandi:
## Motivation

The goal was to measure the performance of the connection pool, and address the potential performance issues.

## Solution

Baseline performance test:

```
pool insert             time:   [334.62 µs 335.07 µs 335.63 µs]
Found 6 outliers among 100 measurements (6.00%)
  1 (1.00%) high mild
  5 (5.00%) high severe

pool lookup             time:   [1.0598 µs 1.0604 µs 1.0610 µs]
Found 12 outliers among 100 measurements (12.00%)
  5 (5.00%) high mild
  7 (7.00%) high severe

pool grow               time:   [336.44 µs 337.22 µs 338.02 µs]
Found 4 outliers among 100 measurements (4.00%)
  4 (4.00%) high mild

pool capacity           time:   [898.70 ns 899.33 ns 900.28 ns]
Found 5 outliers among 100 measurements (5.00%)
  2 (2.00%) high mild
  3 (3.00%) high severe 
```

It turned out that the machine level code for `MaybeUninit::assume_init` produced by the LLVM optimizer is suboptimal, that led to a 200-300 ns overhead for the lookup operation:

https://github.com/rust-lang/rust/issues/74267

The implementation was changed to use `Option` type instead of `MaybeUninit`, as a result we were able to
- address the performance degradation caused by the poorly optimized code
- get rid of the unsafe code.

The use of `Option` type also made it possible to remove the bitmap lookup operation from the `get` and `get_mut` methods, resulting in an additional ~200ns performance improvement.

```
pool insert             time:   [257.49 µs 258.16 µs 258.79 µs]
                        change: [-12.161% -7.0263% -2.8698%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild

pool lookup             time:   [486.44 ns 489.29 ns 493.89 ns]
                        change: [-7.1040% -3.9066% -1.2306%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high severe

pool grow               time:   [253.72 µs 254.16 µs 254.63 µs]
                        change: [-2.3813% -1.3961% -0.6362%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 4 outliers among 100 measurements (4.00%)
  3 (3.00%) high mild
  1 (1.00%) high severe

pool capacity           time:   [311.65 ps 312.07 ps 312.56 ps]
                        change: [-0.0158% +0.3079% +0.6408%] (p = 0.06 > 0.05)
                        No change in performance detected.
Found 4 outliers among 100 measurements (4.00%)
  4 (4.00%) high mild
```

### DIFF
--- a/data-plane/Cargo.lock
+++ b/data-plane/Cargo.lock
@@ -73,6 +73,7 @@ dependencies = [
  "agp-tracing",
  "bit-vec",
  "bytes",
+ "criterion",
  "drain",
  "h2",
  "opentelemetry",
@@ -202,6 +203,12 @@ checksum = "819e7219dbd41043ac279b19830f2efc897156490d7fd6ea916720117ee66311"
 dependencies = [
  "libc",
 ]
+
+[[package]]
+name = "anes"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
 name = "anstream"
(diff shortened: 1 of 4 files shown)


msardara (approved):
LGTM!

micpapal (approved):

PR #131, opened 2025-03-28
refactor: improve utils classes and simplify message processor

msardara:
## Motivation

The utils helper functions were becoming too many and difficult to manage. Also
the return types were ofter Option<T>, forcing the caller to check for None values and
so add more complexity to the code, making it rather unreadable.

## Solution

Use the rust traits and extend protobuf-generated classes with helpers method returning
the actual value and not an Option of it. Also add functions to validate messages, so that the
check for missing fields is performed only at the beginning of the message processing loop.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ helm pull ghcr.io/agntcy/agp/helm/agp:latest
 pip install agp-bindings
 ```
 
-#dd# Copyright Notice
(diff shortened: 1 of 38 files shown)


micpapal (commented):
lgtm

micpapal (approved):

PR #132, opened 2025-03-31
feat: streaming session type

micpapal:


data-plane/gateway/service/src/receiver_buffer.rs
```diff
@@ -48,114 +44,114 @@ impl Default for ReceiverBuffer {
...
+                msg_id
+            );
+            trace!(
+                "buffer status: last sent {}, first entry {}, len {}",
+                self.last_sent,
+                self.first_entry,
+                self.buffer.len()
+            );
+            // check if the msg_id fits inside the buffer range
+            if msg_id <= (self.last_sent + (self.buffer.len() - self.first_entry)) {
```
msardara, line 114: Should we create a helper for computing this?

data-plane/gateway/service/src/receiver_buffer.rs
```diff
@@ -48,114 +44,114 @@ impl Default for ReceiverBuffer {
...
+                error!("the message does not contain a valid id, drop it");
+                return Err(ReceiverBufferError::ProcessingError(e.to_string()));
             }
+            Ok(msg_id) => msg_id as usize,
+        };
+
+        debug!("Received message id {}", msg_id);
+        // no loss detected, return message
+        // if this is the first packet received (case last_sent == usize::MAX) we consider it
+        // valid one and the buffer is initilaized accordingly. in this way a stream can start from
```
msardara:
```suggestion
        // valid one and the buffer is initialized accordingly. in this way a stream can start from
```

data-plane/gateway/service/src/streaming.rs
```diff
@@ -0,0 +1,994 @@
...
+                    tx_gw,
+                    tx_app,
+                ),
+                buffer: Arc::new(RwLock::new(Endpoint::Receiver(recv))),
+                tx,
+            };
+
+            stream.process_message(rx, timer_rx);
+            stream
+        }
```
msardara:
```suggestion
        let buffer = match session_direction {
            SessionDirection::Sender => {
                let prod = Producer {
                    buffer: ProducerBuffer::with_capacity(500),
                    next_id: 0,
                };
                Arc::new(RwLock::new(Endpoint::Producer(prod)))
            }
            SessionDirection::Receiver => {
                let observer = RtxTimerObserver { channel: timer_tx };

                let recv = Receiver {
                    config: session_config.clone(),
                    buffer: ReceiverBuffer::default(),
                    timer_observer: Arc::new(observer),
                    rtx_map: HashMap::new(),
                    timers_map: HashMap::new(),
                };

                Arc::new(RwLock::new(Endpoint::Receiver(recv)))
            }
            _ => {
                error!("invalid session direction");
                Arc::new(RwLock::new(Endpoint::Unknown))
            }
        };

        let stream = Streaming {
            common: Common::new(
                id,
                session_direction,
                SessionConfig::Streaming(session_config),
                tx_gw,
                tx_app,
            ),
            buffer: buffer,
            tx,
        };

        stream.process_message(rx, timer_rx);
        stream
```

data-plane/gateway/service/src/streaming.rs
```diff
@@ -0,0 +1,994 @@
...
+            _ => {
+                panic!("unable to parse streaming configuration");
+            }
+        };
+        let mut producer_name: Option<Agent> = None;
+        //let state = self.common.state().clone();
+        let mut timer_rx_closed = false;
+        tokio::spawn(async move {
+            debug!("starting message processing on session {}", session_id);
+            loop {
```
msardara, line 200: We need a cancellation token and to pass the drain signal to trigger the loop stop.

data-plane/gateway/service/src/streaming.rs
```diff
@@ -0,0 +1,994 @@
...
+                                                trace!("received message from the gataway on producer session {}", session_id);
+                                                // received a message from the GW
+                                                // this must be an RTX message otherwise drop it
+                                                match get_session_header_type(&msg){
+                                                    Ok(session_type) => {
+                                                        if session_type != service_type_to_int(SessionHeaderType::RtxRequest) {
+                                                            error!("received invalid packet type on producer session {}: not RTX request", session_id);
+                                                            continue;
+                                                        }
+                                                        session_type
```
msardara: This session_type does not look like it is stored anywhere. Is this intentional?

data-plane/gateway/service/src/streaming.rs
```diff
@@ -0,0 +1,994 @@
...
+                                break;
+                            }
+                            Some(result) => {
+                                debug!("got a message in process message");
+                                if result.is_err() {
+                                    error!("error receiving a message on session {}, drop it", session_id);
+                                    continue;
+                                }
+                                let (mut msg, direction) = result.unwrap();
+                                match &mut *buffer_clone.write().await {
```
msardara, line 215: I think the write lock on the buffer is not required here.

Rust will complain if you don't use it as you are using an Arc, but as the buffer is not really shared (it looks like is used only in this loop) we can remove the Arc and instantiate it here in process_message

data-plane/gateway/service/src/streaming.rs
```diff
@@ -0,0 +1,994 @@
...
+                                                    timer.stop();
+                                                    continue;
+                                                }
+                                            };
+
+                                            if send_gw.send(Ok(rtx.clone())).await.is_err() {
+                                                error!("error sending RTX for id {} on receiver session {}", msg_id, session_id);
+                                            }
+                                        } else {
+                                            trace!("pacekt {} lost, not retries left", msg_id);
```
msardara, line 465:
```suggestion
                                            trace!("packet {} lost, not retries left", msg_id);
```

data-plane/gateway/service/src/streaming.rs
```diff
@@ -0,0 +1,994 @@
...
+#[allow(dead_code)]
+impl Streaming {
+    pub(crate) fn new(
+        id: Id,
+        session_config: StreamingConfiguration,
+        session_direction: SessionDirection,
+        tx_gw: GwChannelSender,
+        tx_app: AppChannelSender,
+    ) -> Streaming {
+        let (tx, rx) = mpsc::channel(128);
```
msardara, line 133: TODO for later: we might want to make size of the queues configurable.

data-plane/gateway/service/src/streaming.rs
```diff
@@ -0,0 +1,994 @@
...
+        let session_id = self.common.id();
+        let send_gw = self.common.tx_gw();
+        let send_app = self.common.tx_app();
+        let (source, max_retries, timeout) = match self.common.session_config() {
+            SessionConfig::Streaming(streaming_configuration) => (
+                streaming_configuration.source,
+                streaming_configuration.max_retries.unwrap_or(0),
+                streaming_configuration
+                    .timeout
+                    .unwrap_or(std::time::Duration::new(0, 0)),
```
msardara, line 189: Let's set these 0s as default values of `StreamingConfiguration` for max_retries and timeout, and let's remove the Option. WDYT?

data-plane/gateway/service/src/receiver_buffer.rs
micpapal: yes makes sense. I will improve this in the next PR

data-plane/gateway/service/src/streaming.rs
micpapal: Agree! i will modify also this one in the next PR

msardara (approved):

PR #147, opened 2025-04-03
feat: add configuration for control API

zsoltkacsandi:
# Description

references.

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/config/base/client-config.yaml
+++ b/data-plane/config/base/client-config.yaml
@@ -13,7 +13,8 @@ runtime:
 
 services:
   gateway/0:
-    clients:
-      - endpoint: "http://localhost:46357"
-        tls:
-          insecure: true
+    pubsub:
+      clients:
+        - endpoint: "http://localhost:46357"
+          tls:
+            insecure: true
--- a/data-plane/config/base/server-config.yaml
+++ b/data-plane/config/base/server-config.yaml
@@ -13,9 +13,10 @@ runtime:
 
 services:
   gateway/0:
-    server:
-      endpoint: "0.0.0.0:46357"
-      tls:
-        insecure: true
+    pubsub:
+      server:
+        endpoint: "0.0.0.0:46357"
+        tls:
+          insecure: true
 
-    clients: []
+      clients: []
--- a/data-plane/config/basic-auth/client-config.yaml
+++ b/data-plane/config/basic-auth/client-config.yaml
@@ -13,11 +13,12 @@ runtime:
 
 services:
   gateway/0:
-    clients:
-      - endpoint: "https://localhost:46357"
-        tls:
-          ca_file: "./config/crypto/server-ca-cert.pem"
-        auth:
-          basic:
-            username: username
-            password: ${env:PASSWORD}
+    pubsub:
+      clients:
+        - endpoint: "https://localhost:46357"
+          tls:
+            ca_file: "./config/crypto/server-ca-cert.pem"
+          auth:
+            basic:
+              username: username
+              password: ${env:PASSWORD}
--- a/data-plane/config/basic-auth/server-config.yaml
+++ b/data-plane/config/basic-auth/server-config.yaml
@@ -13,14 +13,15 @@ runtime:
 
 services:
   gateway/0:
-    server:
-      endpoint: "0.0.0.0:46357"
-      tls:
-        cert_file: "./config/crypto/server-cert.pem"
-        key_file: "./config/crypto/server-key.pem"
-      auth:
-        basic:
-          username: username
-          password: ${env:PASSWORD}
+    pubsub:
+      server:
+        endpoint: "0.0.0.0:46357"
+        tls:
+          cert_file: "./config/crypto/server-cert.pem"
+          key_file: "./config/crypto/server-key.pem"
+        auth:
+          basic:
+            username: username
+            password: ${env:PASSWORD}
 
-    clients: []
+      clients: []
--- a/data-plane/config/mtls/client-config.yaml
+++ b/data-plane/config/mtls/client-config.yaml
@@ -13,9 +13,10 @@ runtime:
 
 services:
   gateway/0:
-    clients:
-      - endpoint: "https://localhost:46357"
-        tls:
-          ca_file: "./config/crypto/server-ca-cert.pem"
-          cert_file: "./config/crypto/client-cert.pem"
-          key_file: "./config/crypto/client-key.pem"
+    pubsub:
+      clients:
+        - endpoint: "https://localhost:46357"
+          tls:
+            ca_file: "./config/crypto/server-ca-cert.pem"
+            cert_file: "./config/crypto/client-cert.pem"
+            key_file: "./config/crypto/client-key.pem"
--- a/data-plane/config/mtls/server-config.yaml
+++ b/data-plane/config/mtls/server-config.yaml
@@ -13,12 +13,13 @@ runtime:
 
 services:
   gateway/0:
-    server:
-      endpoint: "0.0.0.0:46357"
-      tls:
-        cert_file: "./config/crypto/server-cert.pem"
-        key_file: "./config/crypto/server-key.pem"
+    pubsub:
+      server:
+        endpoint: "0.0.0.0:46357"
+        tls:
+          cert_file: "./config/crypto/server-cert.pem"
+          key_file: "./config/crypto/server-key.pem"
 
-        client_ca_file: "./config/crypto/client-ca-cert.pem"
+          client_ca_file: "./config/crypto/client-ca-cert.pem"
 
-    clients: []
+      clients: []
--- a/data-plane/config/reference/config.yaml
+++ b/data-plane/config/reference/config.yaml
@@ -51,131 +51,372 @@ services:
   # Configuration for the gateway service.
   # This is the main service that will handle the incoming requests.
   gateway/0:
-    # GRPC Server configuration
-    server:
-      # The endpoint to listen on.
-      # No default, this must be specified.
-      endpoint: "0.0.0.0:46357"
+    # pubsub API configuration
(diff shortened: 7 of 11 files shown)


control-plane/internal/cmd/route/route.go
```diff
@@ -0,0 +1,212 @@
...
+	cmd.AddCommand(newAddCmd(opts))
+	cmd.AddCommand(newDelCmd(opts))
+
+	return cmd
+}
+
+func newAddCmd(opts *options.CommonOptions) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "add <company/namespace/agentname> via <host:port>",
+		Short: "Add a route to the gateway",
```
micpapal, line 43: should we add an optional agent id here in the route name?

control-plane/internal/cmd/route/route.go
```diff
@@ -0,0 +1,212 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			routeID := args[0]
+			viaKeyword := strings.ToLower(args[1])
+			endpoint := args[2]
+
+			if viaKeyword != "via" {
+				return fmt.Errorf("invalid syntax: expected 'via' keyword, got '%s'", args[1])
+			}
+
+			company, namespace, agentName, agentID, err := parseRoute(routeID)
```
micpapal, line 55: if the agent id is not there parseRoute is failing, so we need to add it

control-plane/internal/cmd/route/route.go
```diff
@@ -0,0 +1,212 @@
...
+
+			return sendConfigMessage(cmd.Context(), serverAddr, msg)
+		},
+	}
+	return cmd
+}
+
+func newDelCmd(opts *options.CommonOptions) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "del <company/namespace/agentname> via <host:port>",
```
micpapal: here there is the same problem I think

control-plane/internal/controller/client.go
```diff
@@ -0,0 +1,52 @@
...
+	msg *grpcapi.ControlMessage,
+) (*grpcapi.ControlMessage, error) {
+	//TODO(zkacsand): make the timeout configurable
+	opCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer cancel()
+
+	conn, err := grpc.NewClient(
+		serverAddr,
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+	)
```
micpapal, line 29: here we should add the possibility to use basic security features

data-plane/gateway/controller/src/service.rs
```diff
@@ -0,0 +1,326 @@
...
+                                .connections
+                                .read()
+                                .get(&route.connection_id)
+                                .cloned()
+                                .unwrap();
+                            let source = Agent::from_strings(
+                                route.company.as_str(),
+                                route.namespace.as_str(),
+                                "gateway",
+                                0,
```
micpapal, line 123: Why the type is hard-coded? it may be ok but we should put this in a common place (e.g. encoders) as a constant value. Or simply read it as for the others components of the name

micpapal (changes requested):

control-plane/internal/controller/client.go
zsoltkacsandi: I agree, it's planned to implement it with the config file for the agpctl, so the users won't need to provide the common parameters as flags every time.

data-plane/gateway/controller/src/service.rs
zsoltkacsandi: Based on a discussion with @msardara, this can be any arbitrary string, it does not matter. The only reason that something is here, because the messaging layer is not prepared to accept messages something that is not an agent, so a fake agent and agent type has to be passed here.

control-plane/internal/cmd/route/route.go
zsoltkacsandi: I've added it.

control-plane/internal/controller/client.go
zsoltkacsandi: I agree, it's planned to implement it with the config file for the agpctl, so the users won't need to provide the common parameters as flags every time.

data-plane/gateway/controller/src/service.rs
zsoltkacsandi: Based on a discussion with [@msardara](https://github.com/msardara), this can be any arbitrary string, it does not matter. The only reason that something is here, because the messaging layer is not prepared to accept messages something that is not an agent, so a fake agent and agent type has to be passed here. I will make it a constant.

micpapal: yes in this case the source doesn't make much sense because it is not an agent I agree. We can use a default name for this but we can modify it later.

control-plane/Taskfile.yaml
```diff
@@ -115,8 +115,41 @@ tasks:
         vars:
           CMD: "echo ${PWD} && go env && go mod download"
 
+  control-plane:proto:generate:
+    desc: Generate Go protobuf and gRPC code
+    cmds:
+      - >
+        protoc --proto_path=../proto \
```
msardara: We download protoc (and we could also download includes as well) as part of the tools here: https://github.com/agntcy/agp/blob/main/tasks/tools.yaml#L55-L75, so you can add tools:protoc as dependency of this task.

Later you can access it at `{{.REPO_ROOT}}/.tools/protoc`. Thsi will remove the dependency on external protoc and allows to fix a version.

data-plane/gateway/controller/src/service.rs
```diff
@@ -0,0 +1,326 @@
...
+                                    Ok(msg) => {
+                                        if let Err(e) = svc.handle_new_message(msg, tx.clone()).await {
+                                            error!("error processing incoming control message: {:?}", e);
+                                        }
+                                    }
+                                    Err(e) => {
+                                        error!("error receiving control messages: {:?}", e);
+                                    }
+                                }
+                            }
```
msardara: We could remove one indentation level in this way:

```suggestion
                        match next {
                            Some(Ok(msg)) => {
                                if let Err(e) = svc.handle_new_message(msg, tx.clone()).await {
                                    error!("error processing incoming control message: {:?}", e);
                                }
                            }
                            Some(Err(e)) => {
                                error!("error receiving control messages: {:?}", e);
                            }
                            None => {
                                debug!("end of stream");
                                break;
                            }
                        }
```

data-plane/gateway/controller/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+tokio = { version = "1.0", features = ["macros", "rt-multi-thread"] }
+tokio-stream = "0.1"
+tokio-util = "0.7.14"
+tonic = "0.13"
+tracing = "0.1.41"
+uuid = { version = "1.15.1", features = ["v4"] }
+
+[build-dependencies]
+protoc-bin-vendored = "3.1.0"
+tonic-build = { version = "0.13", features = ["prost"] }
```
msardara:
```suggestion
[dependencies]
agp-config = { workspace = true }
agp-datapath = { workspace = true }
bytes = { workspace = true }
parking_lot = { workspace = true }
prost = { workspace = true }
thiserror = { workspace = true }
tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
tokio-stream = { workspace = true }
tokio-util = { workspace = true }
tonic = { workspace = true }
tracing = { workspace = true }
uuid = { workspace = true, features = ["v4"] }

[build-dependencies]
protoc-bin-vendored = { workspace = true }
tonic-build = { workspace = true, features = ["prost"] }
```

data-plane/Cargo.toml
```diff
@@ -34,7 +35,8 @@ edition = "2024"
 [workspace.dependencies]
 # Local dependencies
 agp-config = { path = "gateway/config", version = "0.1.7" }
-agp-datapath = { path = "gateway/datapath", version = "0.6.0" }
+agp-controller = { path = "gateway/controller", version = "0.1.0" }
+agp-datapath = { path = "gateway/datapath", version = "0.5.0" }
```
msardara:
```suggestion
agp-datapath = { path = "gateway/datapath", version = "0.6.0" }
```

data-plane/gateway/datapath/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "agp-datapath"
-version = "0.6.0"
+version = "0.5.0"
```
msardara:
```suggestion
version = "0.6.0"
```

data-plane/gateway/gateway/src/config.rs
```diff
@@ -159,9 +159,9 @@ mod tests {
     use super::*;
     use tracing_test::traced_test;
 
-    #[test]
+    #[tokio::test]
     #[traced_test]
-    fn test_load_config() {
+    async fn test_load_config() {
```
msardara: Maybe we can remov the async from this function now? I don't see any await so we should be good to run it as normal test.

zsoltkacsandi: Yes, with the lazy loading in the controller, we don't need this anymore. I removed it.

data-plane/gateway/controller/src/service.rs
zsoltkacsandi: I've factored this out into a constant.

micpapal (approved):

control-plane/Taskfile.yaml
zsoltkacsandi: Done. :)

PR #152, opened 2025-04-04
test(python-bindings): streaming and pubsub sessions

msardara:


data-plane/python-bindings/src/pysession.rs
```diff
@@ -0,0 +1,182 @@
...
+
+    #[getter]
+    pub fn topic(&self) -> Option<PyAgentType> {
+        self.topic.clone()
+    }
+
+    #[getter]
+    pub fn max_retries(&self) -> u32 {
+        self.streaming_configuration.max_retries
+    }
```
micpapal, line 181: the getter for the the timeout duration is missing. should we add also the setters? I am not sure it will be really useful because once the session is started those values cannot be changed anymore.

data-plane/python-bindings/agp_bindings/__init__.py
```diff
@@ -270,7 +347,7 @@ async def remove_route(
         name = PyAgentType(organization, namespace, agent)
         await remove_route(self.svc, self.conn_id, name, id)
 
-    async def subscribe(self, organization, namespace, agent, id=None):
+    async def subscribe(self, organization: str, namespace: str, agent: str, id=None):
```
micpapal, line 350: here id should be optional to be consistent with the other functions.

data-plane/python-bindings/tests/test_bindings.py
```diff
@@ -107,8 +83,8 @@ async def test_gateway_wrapper(server):
     # Connect to the service and subscribe for the local name
     _ = await gateway1.connect()
 
-    # subscribe to the service
-    await gateway1.subscribe(org, ns, agent1)
+    # # subscribe to the service
+    # await gateway1.subscribe(org, ns, agent1)
```
micpapal, line 87: this I think can be removed

data-plane/python-bindings/tests/test_request_reply.py
```diff
@@ -48,8 +24,8 @@ async def test_request_reply_base(server):
     # Connect to the service and subscribe for the local name
     _ = await gateway1.connect()
 
-    # subscribe to the service
-    await gateway1.subscribe(org, ns, agent1)
+    # # subscribe to the service
+    # await gateway1.subscribe(org, ns, agent1)
```
micpapal, line 28: also this can be removed

micpapal (approved):
only minor comments, lgtm

PR #195, opened 2025-05-01
docs: update readme

91pavan:
# Description

references.

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [x] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/python-bindings/README.md
+++ b/data-plane/python-bindings/README.md
@@ -53,14 +53,7 @@ import agp_bindings
 
 async def run_server(address: str, enable_opentelemetry: bool):
     # init tracing
-    agp_bindings.init_tracing(
-        {
-            "log_level": "debug"
-            "opentelemetry": {
-                "enabled": enable_opentelemetry
-            }
-        }
-    )
+    agp_bindings.init_tracing(log_level="debug", enable_opentelemetry=enable_opentelemetry)
 
     global gateway
     # create new gateway object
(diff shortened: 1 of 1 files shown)


msardara (commented):
Thanks for the PR @91pavan! The examples here are made for the next to-be-released agp-bingings version. If you want to check the examples related to the current version  of the bindings (v.0.3.0), you can check them out at the tag agp-bindings-v0.3.0 (https://github.com/agntcy/agp/tree/agp-bindings-v0.3.0/data-plane/python-bindings/examples), and there indeed the examples are correct for the corresponding bindings version.

We changed this API to be able to automatically convert the tracing configuration provided by the python API into the corresponding rust configuration (https://github.com/agntcy/agp/blob/main/data-plane/gateway/tracing/src/lib.rs#L35-L50), without having to update the API each time we want to add a new parameter. In the new to-be-released version, we will also be able to set the address of the otel collector to use, in this way:

```python
    agp_bindings.init_tracing(
        {
            "log_level": "debug",
            "opentelemetry": {
                "enabled": enable_opentelemetry,
                "grpc": {
                   "endpoint": "http://1.2.3.4:4317",
                   "tls": {
                      "insecure": True
                    }
                 }
            }
        }
    )
```

Which is the same configuration you would provide with configuration file when running the gateway via the docker image:

```yaml
tracing:
  log_level: debug
  display_thread_names: true
  display_thread_ids: true
  opentelemetry:
     enabled: true
     grpc:
        endpoint: http://1.2.3.4:4317
        tls:
           insecure: true

...
```

The GRPC configuration to provide is the same defined in https://github.com/agntcy/agp/blob/main/data-plane/config/reference/config.yaml#L178-L289, both for the config file and for the python bindings.

The new agp-bingings version will be released as soon as we'll get the new version of the rust opentelemetry-otlp package (it should be soon): https://github.com/open-telemetry/opentelemetry-rust/tree/main/opentelemetry-otlp

PR #265, opened 2025-05-15
feat(control-plane): list subscriptions on control-plane

zsoltkacsandi:
# Description

```
❯ go run cmd/main.go route list -s 127.0.0.1:46358
org/default/b id=0 local=[] remote=[0]
org/default/a id=0 local=[] remote=[2] 
```

## Type of Change

- [ ] Bugfix
- [X] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


data-plane/gateway/controller/src/service.rs
```diff
@@ -190,9 +192,57 @@ impl ControllerService {
...
+                                    let resp = ControlMessage {
+                                        message_id: uuid::Uuid::new_v4().to_string(),
+                                        payload: Some(
+                                            crate::api::proto::api::v1::control_message::Payload::SubscriptionListResponse(
+                                                SubscriptionListResponse { entries: entries_to_send }
+                                            ),
+                                        ),
+                                    };
+
+                                    let _ = tx.try_send(Ok(resp));
```
msardara: hmm, let's try not to send anything while keeping the table lock. Probably better filling the vec and the use something like [chunks](https://doc.rust-lang.org/std/primitive.slice.html#method.chunks) to operate on it. What do you think?

Also, I'd like to resolve the connection into something more clear than the connection id itself, so if we fill the vec now we can later resolve the connection ids by looking them up in the connection table and putting more info, like the endpoint or the local application name (not availabe now, but coming soon).

msardara (approved):

PR #307, opened 2025-06-03
feat: implement MLS

zsoltkacsandi:
# Description

references.

## Type of Change

- [ ] Bugfix
- [X] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/core/mls/src/interceptor.rs
+++ b/data-plane/core/mls/src/interceptor.rs
@@ -147,7 +147,7 @@ impl SessionInterceptor for MlsInterceptor {
 
             if is_member {
                 debug!("Decrypting message for group member");
-                mls_guard.decrypt_message(payload)
+                mls_guard.decrypt_message(&self.group_id, payload)
             } else {
                 warn!(
                     "Not a group member but received encrypted message, passing through unchanged"
--- a/data-plane/core/mls/src/mls.rs
+++ b/data-plane/core/mls/src/mls.rs
@@ -180,37 +180,24 @@ impl Mls {
         Self::map_mls_error(encrypted_msg.to_bytes())
     }
 
-    pub fn decrypt_message(&mut self, encrypted_message: &[u8]) -> Result<Vec<u8>, MlsError> {
-        match self.process_message(encrypted_message)? {
-            Some(data) => Ok(data),
-            None => Err(MlsError::Mls(
+    pub fn decrypt_message(
+        &mut self,
+        group_id: &[u8],
+        encrypted_message: &[u8],
+    ) -> Result<Vec<u8>, MlsError> {
+        let group = self
+            .groups
+            .get_mut(group_id)
+            .ok_or_else(|| MlsError::Mls(format!("Group not found: {:?}", group_id)))?;
+
+        let message = Self::map_mls_error(MlsMessage::from_bytes(encrypted_message))?;
+
+        match Self::map_mls_error(group.process_incoming_message(message))? {
+            ReceivedMessage::ApplicationMessage(app_msg) => Ok(app_msg.data().to_vec()),
+            _ => Err(MlsError::Mls(
                 "Message was not an application message".to_string(),
             )),
         }
-    }
-
-    pub fn process_message(&mut self, message: &[u8]) -> Result<Option<Vec<u8>>, MlsError> {
-        let mls_message = Self::map_mls_error(MlsMessage::from_bytes(message))?;
-
-        for group in self.groups.values_mut() {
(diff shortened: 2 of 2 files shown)


data-plane/core/mls/src/interceptor.rs
```diff
@@ -0,0 +1,258 @@
...
+impl MlsInterceptor {
+    pub fn new(mls: Arc<Mutex<Mls>>, group_id: Vec<u8>) -> Self {
+        Self { mls, group_id }
+    }
+}
+
+impl SessionInterceptor for MlsInterceptor {
+    fn on_msg_from_app(&self, msg: &mut Message) {
+        warn!("MLS interceptor on_msg_from_app called");
+
```
micpapal: Here these warning messages should become debug!

data-plane/core/mls/src/interceptor.rs
```diff
@@ -0,0 +1,258 @@
...
+            warn!("Message is not publish type, skipping MLS");
+            return;
+        }
+
+        use base64::{Engine as _, engine::general_purpose};
+        let group_id_b64 = general_purpose::STANDARD.encode(&self.group_id);
+        warn!("Added group_id metadata: {}", group_id_b64);
+        msg.metadata
+            .insert("mls_group_id".to_string(), group_id_b64);
+
```
micpapal, line 43: instead of msg.metadata.insert use the function insert_metadata that is defined here https://github.com/agntcy/slim/blob/main/data-plane/core/datapath/src/messages/utils.rs#L559 so that all the modification to the message are in the same place.

data-plane/core/mls/src/interceptor.rs
```diff
@@ -0,0 +1,258 @@
...
+        let payload = match msg.get_payload() {
```
micpapal, line 44: create a constant for the key "mls_group_id". As an example we have one for the identity here https://github.com/agntcy/slim/blob/main/data-plane/core/datapath/src/messages/utils.rs#L36 but probably it is better to define the constant in this file to keep all the MLS related conf together.

data-plane/core/mls/src/interceptor.rs
```diff
@@ -0,0 +1,258 @@
...
+                    );
+                    warn!("   Encrypted length: {} bytes", encrypted_payload.len());
+                }
+
+                if let Some(MessageType::Publish(publish)) = &mut msg.message_type {
+                    if let Some(content) = &mut publish.msg {
+                        content.blob = encrypted_payload;
+                        if was_encrypted {
+                            msg.metadata
+                                .insert("mls_encrypted".to_string(), "true".to_string());
```
micpapal: same comment as before.

data-plane/examples/src/sdk-mock/main.rs
```diff
@@ -77,6 +79,63 @@ async fn main() {
     // wait for the connection to be established
     tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
 
+    //TODO(zkacsand): use hardcoded group identifier for now
+    let group_identifier = "group_123";
```
micpapal: here it would be better to add some new configurations in the main so that we can switch between MLS on/off easily. In this way can do some test comparison

zsoltkacsandi: eventually this will be removed from the main.rs as it will get integrated in the session layer, but I added the mls-group-id option, so it can be turned on/off.

data-plane/core/mls/src/mls.rs
```diff
@@ -0,0 +1,320 @@
...
+            None => Err(MlsError::Mls(
+                "Message was not an application message".to_string(),
+            )),
+        }
+    }
+
+    pub fn process_message(&mut self, message: &[u8]) -> Result<Option<Vec<u8>>, MlsError> {
+        let mls_message = Self::map_mls_error(MlsMessage::from_bytes(message))?;
+
+        for group in self.groups.values_mut() {
```
muscariello: I am not sure if this loop is necessary. The message is cloned as many times as the number of groups on this session. Why? A message must belong to a single group.
I am not sure if it is correct to keep several groups in a session.

zsoltkacsandi: It was left there from a previous version where the encryption/decryption did not happen in the session. I've fixed it.

muscariello (approved):
LGTM

data-plane/core/mls/src/mls.rs
muscariello: Thanks @zsoltkacsandi

Issue #355, opened 2025-06-18
Add file watcher for public/private key from file


project-sync-agntcy:
Add the file watcher in data-plane/core/auth/src/builder.rs to reload the certificates when needed

PR #356, opened 2025-06-18
feat(config): update the public/private key on file change

micpapal:
Ref: #355

# Description

references.

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


data-plane/core/auth/src/builder.rs
```diff
@@ -313,6 +323,34 @@ impl JwtBuilder<state::WithPrivateKey> {
...
+        let key = Self::build_internal(&self.private_key, &self.algorithm)?;
+
+        let encoding_key = Arc::new(RwLock::new(key));
+
+        let mut file_watcher = None;
+        if let Some(path) = self.watcher {
+            let encoding_key_clone = encoding_key.clone();
+            let algorithm_clone = self.algorithm;
+            let mut w = FileWatcher::create_watcher(move |file: &str| {
+                println!("in callback create callback");
```
msardara: We can remove this print

data-plane/core/auth/src/builder.rs
```diff
@@ -313,6 +323,34 @@ impl JwtBuilder<state::WithPrivateKey> {
...
+                let key = std::fs::read_to_string(file).expect("error reading file");
```
msardara, line 400: I think we can just return the error instead of panicking. WDYT?

data-plane/core/auth/src/builder.rs
```diff
@@ -391,55 +482,27 @@ impl JwtBuilder<state::WithPublicKey> {
...
+                let algorithm_clone = self.algorithm;
+                let mut w = FileWatcher::create_watcher(move |file: &str| {
+                    println!("in callback create callback");
+                    let key = std::fs::read_to_string(file).expect("error reading file");
+                    let new_key = Self::build_internal(&Some(key), &algorithm_clone)
+                        .expect("error processing new keye");
+                    let mut key_lock = encoding_key_clone.write();
+                    *key_lock = new_key;
+                });
+                w.add_file(&path).expect("error adding file to the watcher");
```
msardara: This is the same code of line 397. Let's create a function for it.

data-plane/core/auth/src/jwt.rs
```diff
@@ -105,10 +107,11 @@ pub struct Jwt<T> {
     claims: StandardClaims,
     token_duration: Duration,
     validation: Validation,
-    encoding_key: Option<EncodingKey>,
-    decoding_key: Option<DecodingKey>,
+    encoding_key: Arc<RwLock<Option<EncodingKey>>>,
+    decoding_key: Arc<RwLock<Option<DecodingKey>>>,
```
msardara: What about `Option<Arc<RwLock<EncodingKey>>>`? We can use the same type in key_resolver as well.

data-plane/core/auth/src/builder.rs
micpapal: That error is happening inside the callback. I can modify the signature of the callback and return it but nobody will check it

data-plane/core/auth/src/jwt.rs
micpapal: yes it makes sense. let me modify this

data-plane/core/auth/src/builder.rs
micpapal: This can be a bit involved to to because in the callback I don't know which key to build and update. It requires to modify the watcher and make it specific to this case

data-plane/core/auth/src/builder.rs
msardara: Right, and it does not make much sense to return an error there and then panic in the watcher code. We can leave this way then.

Issue #349, opened 2025-06-16
Improve configuration for connections created via control channel


msardara:
At the moment connections created via control channel, e.g. with agpctl, are able to enforce the following parameters:

https://github.com/agntcy/slim/blob/main/data-plane/core/controller/proto/v1/controller.proto#L74-L79

The grpc connections actually support many parameters, that at the moment can be configured only via conf file or python bindings: 

https://github.com/agntcy/slim/blob/main/data-plane/config/reference/config.yaml#L180-L291

We should be able to set the same parameters using the grpc control channel, so that we can use agpctl (or a controller) to configure them.

Ideally these parameters will change over time, when we add new authentication methods or new parameters, so the protofile for the control messages should not hardcode each one of them, but rather transport a generic map that can be then converted by SLIM into a client config: https://github.com/agntcy/slim/blob/main/data-plane/core/controller/src/service.rs#L78

hackeramitkumar:
I would like to work on it. Please assign this to me.

msardara:
Fixed in #360

PR #360, opened 2025-06-21
feat(control-plane): handle all configuration parameters when creating a new connection

hackeramitkumar:
# Description
Issue #349 

## Type of Change
- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


hackeramitkumar:
@msardara I have modified the ConnectionEntry message. Instead of using hardcoded values, I have added a generic message. I also made changes to whatever seemed relevant to me.
Could you please check if anything is still missing, specifically in the controller/src/service.rs file?

msardara (commented):
Thanks for the PR Amit! I have one comment.

The configuration object can be more complex than a string mapping. I think here we can either use the proto struct or directly a Json string.

This would allow to easily serialize and deserialize the configuration object. In particular in the rust code you can use serde_json::from_string and get the configuration object directly.

My idea would be to define a JSON schema and to pass the JSON directly (or the yaml) to slimctl. In this way you don't have to know the parameters in the code, but you just have to validate the schema. Then you pass the JSON string to slim, that can convert it directly into its internal conf struct via serde, ans thé rest of the code should already work as it is

The schema can be put in the config crate and we can version it together with the crate. And this can definitely help you: https://github.com/GREsau/schemars

hackeramitkumar:
@msardara
One question: Should I add ConnectionEntry.rs, define a Rust struct there, and derive JsonSchema (provided by Schemars), so that during the build, we generate the .json schema for the connection entry?

Or should I manually generate the .json schema using Schemars and directly add the .json schema to our config crate? Then, for validation, we can use this same JSON file. This approach seems more straightforward to me and less confusing for the users as well.

msardara:
Hello Amit! Yup we can use the second approach.

You can generate the JSON schema with schemars starting from the config struct ClientConfig, in data-plane/core/config/src/grpc/client.rs. That represents the JSON string you need to transport in the Protobuf.

data-plane/core/controller/src/service.rs
```diff
@@ -49,9 +54,10 @@ impl ControllerService {
             message_processor,
             tx_slim: OnceLock::new(),
             connections: Arc::new(parking_lot::RwLock::new(HashMap::new())),
+            schema_validator: Arc::new(SchemaValidator::new(CLIENT_CONFIG_SCHEMA_JSON).unwrap_or_default()),
         }
     }
-
+    
     async fn handle_new_control_message(
```
hackeramitkumar: @msardara @micpapal 
In this function the Payload can be : 
1. ConfigCommand : So config command don't have any link with ConnectionEntry. In this we will only get list<Connection> and List<Subcription>. So here  we re creating client _config using Connection only. which contains {id,ip,port}. So what do u think will our changes impact this Payload type haneling ?
2. SubscriptionListRequest : In this we will not get any input data. So here we will have connection table. And be have to return SubscriptionListResponse. Which will require list<local_ConnectionEntry> and list<remote_ConnectionEntry>. So here again by using this much inforamtion we have to generate output. So here what all data should I send in connection entry ? Can you tell me for both the Remote and Local cases.
3. ConnectionListRequest: In this also we will not have any input data. So here again we will have connection_table. So using this we have to generate List<ConnectionEntry>. So again what all data will be required in the new connectionEntry.

data-plane/core/controller/proto/v1/controller.proto
```diff
@@ -72,10 +72,7 @@ enum ConnectionType {
 }
 
 message ConnectionEntry {
-  uint64 id = 1;
-  ConnectionType connection_type = 2;
-  string ip = 3;
-  uint32 port = 4;
+  string config_data = 1;
```
hackeramitkumar: In the grpc ClientConfig we don't have id, connection_type. So should we keep this in our config_data ?

data-plane/core/controller/src/service.rs
msardara: @hackeramitkumar 

 > ConfigCommand : So config command don't have any link with ConnectionEntry. In this we will only get list and List. So here we re creating client _config using Connection only. which contains {id,ip,port}. So what do u think will our changes impact this Payload type haneling ?

Yup, the `ConfigurationCommand` is a message of list of connections and subscriptions, as defined in the protobuf. I think we can leave it as it is, and you need only to update the Connection and ConnectionEntry messages in this way:

```diff
diff --git a/data-plane/core/controller/proto/v1/controller.proto b/data-plane/core/controller/proto/v1/controller.proto
index 6aaa06b..f822b87 100644
--- a/data-plane/core/controller/proto/v1/controller.proto
+++ b/data-plane/core/controller/proto/v1/controller.proto
@@ -27,8 +27,7 @@ message ControlMessage {
 
 message Connection {
   string connection_id = 1;
-  string remote_address = 2;
-  int32 remote_port = 3;
+  string config_data = 2;
 }
 
 message Subscription {
@@ -74,8 +73,7 @@ enum ConnectionType {
 message ConnectionEntry {
   uint64 id = 1;
   ConnectionType connection_type = 2;
-  string ip = 3;
-  uint32 port = 4;
+  string config_data = 3;
 }
 
 message ConnectionListRequest {}
```

This means that now, instead of just receiving a remote_ip:remote_port pair, you can receive a json string containing the whole configuration you need to create the connection. Once you have that, the client config can be created simply using serde_json in this way:

```diff
diff --git a/data-plane/core/controller/src/service.rs b/data-plane/core/controller/src/service.rs
index 7fd0233..afdacd9 100644
--- a/data-plane/core/controller/src/service.rs
+++ b/data-plane/core/controller/src/service.rs
@@ -75,11 +75,9 @@ impl ControllerService {
 
                             // connect to an endpoint if it's not already connected
                             if !self.connections.read().contains_key(&client_endpoint) {
-                                let client_config = ClientConfig {
-                                    endpoint: format!("http://{}", client_endpoint),
-                                    tls_setting: TlsClientConfig::default().with_insecure(true),
-                                    ..ClientConfig::default()
-                                };
+                                let client_config = serde_json::from_str::<ClientConfig>(&conn.config_data).map_err(
+                                    |e| ControllerError::ConfigurationError(e.to_string())
+                                )?;
 
                                 match client_config.to_channel() {
                                     Err(e) => {
```

msardara: > SubscriptionListRequest : In this we will not get any input data. So here we will have connection table. And be have to return SubscriptionListResponse. Which will require list<local_ConnectionEntry> and list<remote_ConnectionEntry>. So here again by using this much inforamtion we have to generate output. So here what all data should I send in connection entry ? Can you tell me for both the Remote and Local cases.

Here we are missing the required information in the `Connection` struct in `data-plane/core/datapath/src/connection.rs`. We can add the ClientConfig as member of the Connection struct when we create the connection:

```diff
diff --git a/data-plane/core/datapath/src/message_processing.rs b/data-plane/core/datapath/src/message_processing.rs
index d38c36f..23d1edd 100644
--- a/data-plane/core/datapath/src/message_processing.rs
+++ b/data-plane/core/datapath/src/message_processing.rs
@@ -177,6 +177,7 @@ impl MessageProcessor {
                         .with_local_addr(local)
                         .with_remote_addr(remote)
                         .with_channel(Channel::Client(tx))
+                        .with_config_data(client_config.clone())
                         .with_cancellation_token(Some(cancellation_token.clone()));
 
                     debug!(

```

 The connection struct should be modified to be something like this:

```diff
diff --git a/data-plane/core/datapath/src/connection.rs b/data-plane/core/datapath/src/connection.rs
index ccc8d98..fe5f07d 100644
--- a/data-plane/core/datapath/src/connection.rs
+++ b/data-plane/core/datapath/src/connection.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::net::SocketAddr;
+use slim_config::grpc::client::ClientConfig;
 use tokio::sync::mpsc;
 use tokio_util::sync::CancellationToken;
 use tonic::Status;
@@ -43,6 +44,9 @@ pub struct Connection {
     /// Channel to send messages
     channel: Channel,
 
+    /// Configuration data for the connection.
+    config_data: Option<ClientConfig>,
+
     /// Connection type
     connection_type: Type,
 
@@ -78,6 +82,14 @@ impl Connection {
         Self { channel, ..self }
     }
 
+    /// Set the configuration data for the connection
+    pub(crate) fn with_config_data(self, config_data: Option<ClientConfig>) -> Self {
+        Self {
+            config_data,
+            ..self
+        }
+    }
+
     /// Get the remote address
     pub fn remote_addr(&self) -> Option<&SocketAddr> {
         self.remote_addr.as_ref()
@@ -88,6 +100,10 @@ impl Connection {
         self.local_addr.as_ref()
     }
 
+    pub fn config_data(&self) -> Option<&ClientConfig> {
+        self.config_data.as_ref()
+    }
+
     /// Get the channel
     pub(crate) fn channel(&self) -> &Channel {
         &self.channel

```

And at this point, when listing the connections you just need to convert it to json string:

```diff
diff --git a/data-plane/core/controller/src/service.rs b/data-plane/core/controller/src/service.rs
index 7fd0233..8a29d78 100644
--- a/data-plane/core/controller/src/service.rs
+++ b/data-plane/core/controller/src/service.rs
diff --git a/data-plane/core/controller/src/service.rs b/data-plane/core/controller/src/service.rs
index 7fd0233..95a89f1 100644
--- a/data-plane/core/controller/src/service.rs
+++ b/data-plane/core/controller/src/service.rs
@@ -226,36 +224,23 @@ impl ControllerService {
                                     entry.local_connections.push(ConnectionEntry {
                                         id:   cid,
                                         connection_type: ConnectionType::Local as i32,
-                                        ip:   String::new(),
-                                        port: 0,
+                                        config_data: "{}".to_string(),
                                     });
                                 }
 
                                 for &cid in remote {
                                     if let Some(conn) = conn_table.get(cid as usize) {
-                                        if let Some(sock) = conn.remote_addr() {
-                                            entry.remote_connections.push(ConnectionEntry {
-                                                id:   cid,
-                                                connection_type: ConnectionType::Remote as i32,
-                                                ip:   sock.ip().to_string(),
-                                                port: sock.port() as u32,
-                                            });
-                                        } else {
-                                            entry.remote_connections.push(ConnectionEntry {
-                                                id:   cid,
-                                                connection_type: ConnectionType::Remote as i32,
-                                                ip:   String::new(),
-                                                port: 0,
-                                            });
-                                        }
-                                    } else {
-                                        error!("no connection entry for id {}", cid);
                                         entry.remote_connections.push(ConnectionEntry {
                                             id:   cid,
                                             connection_type: ConnectionType::Remote as i32,
-                                            ip:   String::new(),
-                                            port: 0,
+                                            config_data: match conn.config_data() {
+                                                Some(data) => serde_json::to_string(data)
+                                                    .unwrap_or_else(|_| "{}".to_string()),
+                                                None => "{}".to_string(),
+                                            },
                                         });
+                                    } else {
+                                        error!("no connection entry for id {}", cid);
                                     }
                                 }

```

msardara: > ConnectionListRequest: In this also we will not have any input data. So here again we will have connection_table. So using this we have to generate List. So again what all data will be required in the new connectionEntry.

Again here you can follow the same of point 2:

```diff
diff --git a/data-plane/core/controller/src/service.rs b/data-plane/core/controller/src/service.rs
index 7fd0233..03677e5 100644
--- a/data-plane/core/controller/src/service.rs
+++ b/data-plane/core/controller/src/service.rs
@@ -292,8 +277,11 @@ impl ControllerService {
                                 all_entries.push(ConnectionEntry {
                                     id: id as u64,
                                     connection_type: ConnectionType::Remote as i32,
-                                    ip,
-                                    port,
+                                    config_data: match conn.config_data() {
+                                        Some(data) => serde_json::to_string(data)
+                                            .unwrap_or_else(|_| "{}".to_string()),
+                                        None => "{}".to_string(),
+                                    },
                                 });
                             });
 

```

data-plane/core/config/src/client-config.schema.json
msardara, line 1: We should also commit the `#[derive(JsonSchema)]` directive you likely used to generate the schema, as well as an utility binary that when run outputs the JSON schema to a file.

The validation should be done first in the slimctl CLI, by the cli itself, and then if that succeeds we can send the json to SLIM.

In SLIM in theory we don't need to validate the json schema, as it is implicitly done by serde_json when trying to convert the json string to ClientConfig. The schema is useful for clients like slimctl.

data-plane/core/controller/proto/v1/controller.proto
msardara: Yes, as the id and the connection type are external and taken from the connection itself.

control-plane/slimctl/internal/controller/validator.go
```diff
@@ -0,0 +1,36 @@
+package controller
+
+import (
+	_ "embed"
+	"log"
+
+	"github.com/xeipuuv/gojsonschema"
+)
+
+//go:embed client-config.schema.json
```
msardara: Maybe here we can add a `go:generate`  statement that would either download the schema or get the schema from the config crate, WDYT?

msardara (commented):
Just one small comment about how we are getting the schema file in the control plane part, otherwise it looks really good to me.

Great work @hackeramitkumar!

control-plane/slimctl/internal/controller/validator.go
hackeramitkumar: As of now will add a cp command in the go:generate

msardara (approved):

PR #371, opened 2025-06-25
feat: add the ability to drop messages from the interceptor

zsoltkacsandi:
# Description

Adding the ability from the interceptor to drop a message if an error occurs during the processing.

## Type of Change

- [ ] Bugfix
- [X] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/core/mls/src/interceptor.rs
+++ b/data-plane/core/mls/src/interceptor.rs
@@ -5,7 +5,7 @@ use crate::mls::Mls;
 use parking_lot::Mutex;
 use slim_datapath::api::MessageType;
 use slim_datapath::api::proto::pubsub::v1::Message;
-use slim_service::session::SessionInterceptor;
+use slim_service::{errors::SessionError, session::SessionInterceptor};
 use std::sync::Arc;
 use tracing::{debug, error, warn};
 
@@ -32,11 +32,11 @@ impl MlsInterceptor {
 }
 
 impl SessionInterceptor for MlsInterceptor {
-    fn on_msg_from_app(&self, msg: &mut Message) {
+    fn on_msg_from_app(&self, msg: &mut Message) -> Result<(), SessionError> {
         // Only process Publish message types
         if !msg.is_publish() {
             debug!("Skipping non-Publish message type in encryption path");
-            return;
+            return Ok(());
         }
 
         msg.insert_metadata(METADATA_MLS_GROUP_ID.to_owned(), self.group_id_b64.clone());
@@ -45,7 +45,7 @@ impl SessionInterceptor for MlsInterceptor {
             Some(content) => &content.blob,
             None => {
                 warn!("Message has no payload, skipping MLS processing");
-                return;
+                return Ok(());
             }
         };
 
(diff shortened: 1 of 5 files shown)


data-plane/core/mls/src/interceptor.rs
```diff
@@ -32,11 +32,11 @@ impl MlsInterceptor {
 }
 
 impl SessionInterceptor for MlsInterceptor {
-    fn on_msg_from_app(&self, msg: &mut Message) {
+    fn on_msg_from_app(&self, msg: &mut Message) -> Result<(), SessionError> {
```
micpapal, line 35: Is it possible to use RwLock in the MlsInterceptor struct (line 17) instead of Mutex?

data-plane/core/mls/src/interceptor.rs
```diff
@@ -60,8 +60,13 @@ impl SessionInterceptor for MlsInterceptor {
...
             } else {
-                debug!("Not a group member, sending message unencrypted");
-                (Ok(payload.to_vec()), false)
+                warn!("Not a group member, dropping message");
+                (
+                    Err(crate::errors::MlsError::Mls(
+                        "Not a group member".to_string(),
+                    )),
+                    false,
+                )
```
micpapal: this piece of code is quite convoluted. I think you simply return directly an error at line 60 and 65 and modify the payload in the ok at line 59.

data-plane/core/mls/src/interceptor.rs
```diff
@@ -117,27 +119,31 @@ impl SessionInterceptor for MlsInterceptor {
...
         }
 
         let payload = match msg.get_payload() {
             Some(content) => &content.blob,
             None => {
-                //TODO(zkacsand): DROP
-                warn!("Encrypted message has no payload, dropping message");
-                return;
+                warn!("Encrypted message has no payload");
+                return Err(SessionError::InterceptorError(
```
micpapal, line 130: Is this always an error? At line 48 if a message has no payload you return an Ok. I don't know if this can happen but I think we should return Ok maybe logging the empty message

zsoltkacsandi: The difference between the two, that at line 48 if there is no payload, we cannot encrypt it (we just need to handle the None case), but we won't mark the message as encrypted at line 78. From our perspective, there is nothing to send, nothing to encrypt.

In this case, we get a message that we already detected as encrypted (the encryption metadata was added, that should not happen after line 48) at line 108 with checking it's metadata, but there is no payload, so the decryption operation would fail. This should not happen ideally, but in my opinion it's definitely an error.

data-plane/core/mls/src/interceptor.rs
zsoltkacsandi: The Mls Group (on which we call the encrypt, decrypt and other methods) needs to be mutable, and the operations can mutate it's internal state, so even if we used RwLock instead of mutex, we need to lock it for write, so there won't be any benefit.

data-plane/core/mls/src/interceptor.rs
zsoltkacsandi: I simplified this part of the code for both methods.

data-plane/core/mls/src/interceptor.rs
micpapal: yes I missed the fact that the packet is marked as encrypted in this check. I was looking into this particular case because it may happen that some control messages are without payload but sent as a Publish, but what you said cover this case.

PR #385, opened 2025-07-03
feat: integrate MLS with auth

zsoltkacsandi:
# Description

Please provide a meaningful description of what this change will do, or is for.
Bonus points for including links to related issues, other PRs, or technical
references.

Note that by _not_ including a description, you are asking reviewers to do extra
work to understand the context of this change, which may lead to your PR taking
much longer to review, or result in it not being reviewed at all.

## Type of Change

- [ ] Bugfix
- [X] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [ ] I have read the [contributing guidelines](/agntcy/repo-template/blob/main/CONTRIBUTING.md)
- [ ] Existing issues have been referenced (where applicable)
- [ ] I have verified this change is not present in other open pull requests
- [ ] Functionality is documented
- [ ] All code style checks pass
- [ ] New code contribution is covered by automated tests
- [ ] All new and existing tests pass

PR #386, opened 2025-07-07
363 add mls message types in slim messages

micpapal:
# Description

references.

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


data-plane/core/service/src/session.rs
```diff
@@ -482,78 +506,73 @@ where
...
     ) -> Self {
-        Self {
+        let mls = if msl_enabled {
+            // ?? how to use the right identity here ??
+            let mut name = "/tmp/mls_id_".to_owned();
+            let rnd = rand::random::<u32>();
+            name.push_str(&rnd.to_string());
+            let id_provider =
+                Arc::new(FileBasedIdentityProvider::new(name).expect("error creating id provider"));
+            Some(Arc::new(Mutex::new(Mls::new(rnd.to_string(), id_provider))))
```
msardara: I think this is being addressed in #385

data-plane/core/mls/src/mls.rs
```diff
@@ -133,62 +133,112 @@ impl Mls {
...
 
+        // process the welcome message and connect to the group
         let welcome = Self::map_mls_error(MlsMessage::from_bytes(welcome_message))?;
         let (group, _) = Self::map_mls_error(client.join_group(None, &welcome))?;
 
         let group_id = group.group_id().to_vec();
-        self.groups.insert(group_id.clone(), group);
+        self.group = Some(group);
+
+        use base64::{Engine as _, engine::general_purpose};
```
msardara: Move this with the rest of imports

data-plane/core/service/src/channel_endpoint.rs
```diff
@@ -98,6 +106,216 @@ where
...
+                match lock.process_commit(commit) {
+                    Ok(_) => Ok(()),
+                    Err(e) => {
+                        error!("error processing commit message {}", e.to_string());
+                        Err(ChannelEndpointError::CommitMessage)
+                    }
+                }
+            }
+            None => {
+                error!("MLS not seutp, drop commit message");
```
msardara:
```suggestion
                error!("MLS not setup, drop commit message");
```

data-plane/core/mls/src/mls.rs
```diff
@@ -98,7 +99,10 @@ impl Mls {
             Self::map_mls_error(client.create_group(ExtensionList::default(), Default::default()))?;
 
         let group_id = group.group_id().to_vec();
-        self.groups.insert(group_id.clone(), group);
+        self.group = Some(group);
+
+        use base64::{Engine as _, engine::general_purpose};
```
msardara: import this as part of the other imports

data-plane/core/mls/src/mls.rs
```diff
@@ -133,62 +133,112 @@ impl Mls {
...
 
-        Self::map_mls_error(encrypted_msg.to_bytes())
+        let msg = Self::map_mls_error(encrypted_msg.to_bytes())?;
+        Ok((msg, id.to_string()))
     }
 
     pub fn decrypt_message(
         &mut self,
-        group_id: &[u8],
+        group_id: &String,
```
msardara:
```suggestion
        group_id: &str,
```

msardara (commented):
LGTM, modulo some comments:)

data-plane/core/service/src/session.rs
micpapal: I will leave a comment there and we can replace the code as soon as the other PR is ready

data-plane/core/mls/src/mls.rs
zsoltkacsandi: Are we sure that we always pass valid UTF-8 here, no serialized/binary data, etc? If we don't I would stick with `&[u8]`, otherwise using strings can cause panics.

But if we want to change it, we should change the parameter signature of encrypt_message as well.

micpapal: I think that the group id can be remove completely. now we have only one group so we don't need to match the group id to get the right mls state. we can simply try to decrypt the message using the only available state and return an error if the message cannot be decrypted

PR #413, opened 2025-07-11
397 remove endpoints in mls groups

micpapal:
# Description

references.

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


data-plane/core/mls/src/mls.rs
```diff
@@ -256,7 +265,39 @@ where
...
+
+        let ret = MlsAddMemberResult {
+            welcome_message: welcome,
+            commit_message: commit_msg,
+            member_identity: identifier,
+        };
+        Ok(ret)
+    }
+
+    pub fn remove_member(&mut self, identity: &[u8]) -> Result<Vec<u8>, MlsError> {
```
msardara: Could we set an alias for Vec<u8> to understand the meaning of the result of this api as well?

data-plane/core/mls/src/mls.rs
```diff
@@ -256,7 +265,39 @@ where
...
+        info!("Removing member from the  MLS group");
```
msardara: `debug!`

data-plane/core/service/src/channel_endpoint.rs
```diff
@@ -117,6 +113,11 @@ where
 
     /// last commit id
     last_commit_id: u32,
+
+    /// map of the participants with package keys
+    /// this is used only by the moderator to remove
+    /// participants from the channel
+    participants: HashMap<Agent, Vec<u8>>,
```
msardara: Shouldn't we add it to the ChannelModerator only then?

data-plane/core/mls/src/mls.rs
micpapal: Sure I can do this, it will be more readable

data-plane/core/service/src/channel_endpoint.rs
micpapal: I added it there because it is part of the MLS state. And also in the moderator may not be set in MLS is not used. I may split this state in two struct in the next PR

PR #416, opened 2025-07-15
fix: remove request-reply session type

msardara:
BREAKING CHANGE: remove of one session type

# Description

We remove the request-reply session, as request-reply
semantics can be easily implemented in the application layer
if the application requires a reply.

Most of the time, applications require sessions that can handle
bidirectional streaming (e.g. fire-and-forget for anycast and pubsub for multicast), on top
of which we can easily build-up a request-reply semantic.
No need to implement a specific session type only for this.

We will still keep the high level API request-reply in the python
bindings, but the underlying session will not be a request-reply only session,
but rather a bidirectional streaming session.

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [x] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


data-plane/python-bindings/tests/test_request_reply.py
```diff
@@ -134,13 +38,16 @@ async def test_request_reply(server):
 
     # create request/reply session with default config
     session_info = await slim2.create_session(
-        slim_bindings.PySessionConfiguration.RequestResponse(),
+        slim_bindings.PySessionConfiguration.FireAndForget(
+            timeout=datetime.timedelta(seconds=1), max_retries=3, sticky=False
+        )
     )
```
micpapal, line 44: does it makes sense to still provide the Request/Reply session type at the app level as a wrapper on top of Fnf?

msardara: Yup, I'll think about it when updating the APIs with the new identity/encryption functionalities

micpapal (approved):

PR #431, opened 2025-07-17
fix: skip processing commit for an already removed participant

zsoltkacsandi:
# Description

If the moderator retries the commit message for a removed member, the participants are trying to process it again, even if they have already removed that member.

## Type of Change

- [X] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/core/service/src/channel_endpoint.rs
+++ b/data-plane/core/service/src/channel_endpoint.rs
@@ -291,10 +291,11 @@ where
         let id = match self.participants.get(&name) {
             Some(id) => id,
             None => {
-                error!("the name does not exists in the group");
-                return Err(SessionError::RemoveParticipant(
-                    "participant does not exists".to_owned(),
-                ));
+                debug!(
+                    "participant {} already removed or not part of the group",
+                    name.agent_id()
+                );
+                return Ok(vec![]);
             }
         };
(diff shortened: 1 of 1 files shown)


data-plane/core/service/src/channel_endpoint.rs
```diff
@@ -1063,28 +1064,34 @@ where
...
-                // wait for len + 1 acks because the participant list does not contains
-                // the removed participant anymore
-                let len = self.endpoint.mls_state.as_ref().unwrap().participants.len() + 1;
-
-                // the leave request will be forwarded after all acks are received
-                self.create_timer(commit_id, (len).try_into().unwrap(), commit, Some(msg));
-
-                Ok(())
+                // Check if participant was already removed (empty commit payload)
+                if commit_payload.is_empty() {
```
micpapal, line 1068: this packet is coming from the application so if the participant is not in the group either the removal is still in progress or the app wants to remove a participant that wasn't added yet. 
I think that if the remove_participant fails we should drop the packet and do not forward it. Because in the first case eventually it will be forwarded when all the acks are received. In the second it will be dropped by the receiver.

PR #520, opened 2025-08-05
feat(srpc): server skeleton

msardara:
# Description

Server skeleton for srpc.

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


data-plane/python-bindings/srpc/src/srpc/rpc.py
```diff
@@ -0,0 +1,20 @@
...
+
+class Rpc:
+    """
+    Base class for RPC object. It holds
+    """
+
+    def __init__(
+        self,
+        name: str,
+        handler: dict,
```
Tehsmash: I think this needs to be Callable too.

data-plane/python-bindings/srpc/src/srpc/server.py
```diff
@@ -0,0 +1,138 @@
...
+        self.local_app = asyncio.run(
+            create_local_app(
+                local,
+                slim,
+                enable_opentelemetry=enable_opentelemetry,
+                shared_secret=shared_secret,
+            )
+        )
+
+    async def register_rpc(self, rpc_handler: Rpc):
```
Tehsmash: following the grpc server functions and generated code can we add the service name and method name separately? For example:

```
server.register_rpc(Rpc(service="a2a.v1.A2AService", method="SendMessage", handler=...)
```

Or even closer is to take a map:

```
handlers = {
   "SendMessage": RpcHandler(handler=servicer.send_message, request_serializer=..., response_serializer=..., request_streaming=False, response_streaming=False),
    "SendStreamingMessage": RpcHandler(handler=servicer.send_message, request_serializer=..., response_serializer=..., request_streaming=False, response_streaming=True),
}

server.register_method_handlers("a2a.v1.A2AService",  handlers)
```

data-plane/python-bindings/srpc/src/srpc/server.py
```diff
@@ -0,0 +1,138 @@
...
+        self.handlers = {}
+
+        self.local_app = asyncio.run(
+            create_local_app(
+                local,
+                slim,
+                enable_opentelemetry=enable_opentelemetry,
+                shared_secret=shared_secret,
+            )
+        )
```
Tehsmash: I'm not a big fan of asyncio things in the __init__ for a class, this can cause pain if the asyncio loop has already started etc. I think it would be better to have this in the run() function and then also most the subscribe calls into the run(..) function too.

data-plane/python-bindings/srpc/src/srpc/server.py
```diff
@@ -0,0 +1,138 @@
...
+            rpc_handler.name,
+        )
+        await self.local_app.subscribe(
+            subscription_name,
+        )
+
+        # Register the RPC handler
+        self.handlers[subscription_name] = rpc_handler
+
+    async def run(self):
```
Tehsmash, line 58: lets move all the async setup into here like:

```
async def run(self):
    local_app = await create_local_app(...)

	for subscription_name, handler in self.handlers.items():
    	local_app.subscribe(subscription_name)

	async with local_app:
    	...
```

data-plane/python-bindings/srpc/src/srpc/server.py
```diff
@@ -0,0 +1,138 @@
...
+                            # Call the RPC handler and get the response
+                            response = await rpc_handler.handler(request)
+
+                            # Send the response back to the client
+                            await self.local_app.publish_to(
+                                session_info,
+                                response,
+                            )
+
+                            # TODO(msardara): handle cleanup
```
Tehsmash: Maybe move this into:

```
async def handle_session(self, session_info):
    while True:
        ...
```

data-plane/python-bindings/srpc/src/srpc/server.py
```diff
@@ -0,0 +1,138 @@
...
+                            )
+
+                            # Call the RPC handler and get the response
+                            response = await rpc_handler.handler(request)
+
+                            # Send the response back to the client
+                            await self.local_app.publish_to(
+                                session_info,
+                                response,
+                            )
```
Tehsmash: Need to handle the non-streaming vs streaming case.

I think we can assume that in the streaming case handler is an async generator like:

```
if rpc_handler.response_streaming:
	async for response in rpc_handle.handler(request):
		await local_app.publish_to(session_info, response)
```

micpapal:
do the merge on A2A-hackathon branch not on main

data-plane/python-bindings/srpc/src/srpc/server.py
```diff
@@ -0,0 +1,138 @@
...
+
+                            request = rpc_handler.request_deserializer(msg)
+
+                            format_message_print(
+                                f"{instance}",
+                                f"calling handler {rpc_handler.name} for session {session_id.id}",
+                            )
+
+                            # Call the RPC handler and get the response
+                            response = await rpc_handler.handler(request)
```
Tehsmash: We're missing:

```
response_bytes = rpc_handler.response_serializer(response)

await self.local_app.publish_to(
    session_info,
    response_bytes,
)
```

data-plane/python-bindings/srpc/src/srpc/server.py
```diff
@@ -0,0 +1,138 @@
...
+
+import asyncio
+import datetime
+
+
+import click
+
+import slim_bindings
+
+from .rpc import Rpc
```
Tehsmash: its best practise to always us absolute imports even within the same package like:

```
from srpc.rpc import Rpc
```

data-plane/python-bindings/srpc/src/srpc/server.py
```diff
@@ -0,0 +1,138 @@
...
+
+from .common import (
+    common_options,
+    create_local_app,
+    format_message_print,
+    split_id,
+)
```
Tehsmash, line 11: ditto

data-plane/python-bindings/srpc/src/srpc/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+    )
+
+    try:
+        asyncio.run(server.run())
+    except KeyboardInterrupt:
+        print("Client interrupted by user.")
+
+
+if __name__ == "__main__":
+    server_main()
```
Tehsmash: I wouldn't include an example in the __init__.py, this should really be used to determine what is exposed at the top level of the srpc package to make using the library nicer.

data-plane/python-bindings/srpc/src/srpc/common.py
```diff
@@ -0,0 +1,241 @@
...
+        help="Invite other participants to the pubsub session. Can be specified multiple times.",
+    )(function)
+
+    function = click.option(
+        "--enable-mls",
+        is_flag=True,
+        help="Enable MLS (Message Layer Security) for the pubsub session.",
+    )(function)
+
+    return function
```
Tehsmash, line 196: I would switch to using pydantic-settings rather than click for this, I think it'll make it more re-usable because you can create the settings model separately from the CLI like:

```
from pydantic_settings import BaseSettings

class TlsConfig(BaseModel):
   insecure: bool = False

class SlimConfig(BaseModel):
	endpoint: str
    tls: TlsConfig

class SRPCServerConfig(BaseSettings):
	model_config = SettingsConfigDict(cli_parse_args=True, env_prefix='SLIM_')
   	
	local: str
    slim: SlimConfig
    ...
```

as a user of the library, I can then do:

```
config = SRPCServerConfig()
server = Server(config)
await server.run()
```

data-plane/python-bindings/srpc/src/srpc/server.py
```diff
@@ -0,0 +1,138 @@
...
+    async def run(self):
+        instance = self.local_app.get_id()
+
+        async with self.local_app:
+            # Wait for a message and reply in a loop
+            while True:
+                format_message_print(
+                    f"{instance}",
+                    "waiting for new session to be established",
+                )
```
Tehsmash: Lets use the logging package, I have a cool library I wrote for aether which we can use which makes it possible to add structured fields to the various logs depends on the threads which will make it much nicer to log the session id's etc.

data-plane/python-bindings/srpc/src/srpc/server.py
```diff
@@ -0,0 +1,144 @@
...
+
+                # Send the response back to the client
+                if rpc_handler.response_streaming:
+                    async for response in rpc_handler.handler(request):
+                        await local_app.publish_to(session_info, response)
+                else:
+                    await local_app.publish_to(
+                        session_info,
+                        response,
+                    )
```
Tehsmash:
```
# Call the RPC handler and get the response
if rpc_handler.response_streaming:
	async for response in rpc_handler.handler(request):
		response_bytes = rpc_handler.response_serializer(response)
        await local_app.publish_to(session_info, response_bytes)
	return
                
response = await rpc_handler.handler(request)
response_bytes = rpc_handler.response_serializer(response)    		
await local_app.publish_to(
	session_info,
	response_bytes,
)
```

Tehsmash (changes requested):

Tehsmash (approved):

PR #522, opened 2025-08-05
feat: error handling

msardara:
# Description

Add error handling when calling grpc handler.

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/python-bindings/srpc/pyproject.toml
+++ b/data-plane/python-bindings/srpc/pyproject.toml
@@ -7,7 +7,11 @@ version = "0.1.0"
 description = "RPC over Slim Library"
 requires-python = ">=3.9"
 license = "Apache-2.0"
-dependencies = ["slim-bindings>=0.3.6", "click>=8.1.8", "grpcio>=1.64.0", "grpcio-tools>=1.64.0"]
+dependencies = [
+    "slim-bindings>=0.3.6",
+    "click>=8.1.8",
+    "googleapis-common-protos>=1.70.0",
+]
 
 [tool.uv.sources]
 slim-bindings = { workspace = true }
--- a/data-plane/python-bindings/srpc/src/srpc/rpc.py
+++ b/data-plane/python-bindings/srpc/src/srpc/rpc.py
@@ -2,7 +2,16 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from collections.abc import Awaitable
+from typing import Any
 
+from google.rpc import status_pb2, code_pb2
+
+class ErrorResponse(Exception):
+    def __init__(self, code, message, details=None):
+        self.code = code
+        self.message = message
+        self.details = details
+        super().__init__(message)
 
 class Rpc:
     """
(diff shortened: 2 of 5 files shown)


data-plane/python-bindings/srpc/src/srpc/server.py
```diff
@@ -108,17 +108,22 @@ async def handle_session(self, session_info, local_app):
...
 
                     return
 
-                response = await rpc_handler.handler(request, context)
+                response = await rpc_handler.call_handler(request, context)
                 response_bytes = rpc_handler.response_serializer(response)
                 await local_app.publish_to(
                     session_info,
                     response_bytes,
+                    metadata={"code": 200},
```
micpapal: do we want 200 or should we keep always 0 like in streaming case with no error?

data-plane/python-bindings/srpc/pyproject.toml
```diff
@@ -7,7 +7,11 @@ version = "0.1.0"
 description = "RPC over Slim Library"
 requires-python = ">=3.9"
 license = "Apache-2.0"
-dependencies = ["slim-bindings>=0.3.6", "click>=8.1.8", "grpcio>=1.64.0", "grpcio-tools>=1.64.0"]
+dependencies = [
+    "slim-bindings>=0.3.6",
```
micpapal: should we go for version >=0.4.0?

data-plane/python-bindings/srpc/src/srpc/server.py
msardara: Right, let's put 0

data-plane/python-bindings/srpc/src/srpc/rpc.py
```diff
@@ -26,3 +38,25 @@ def __init__(
...
+                code=e.code, message=e.message, details=e.details
+            )
+            code = e.code
+        except Exception:
+            response = status_pb2.Status(
+                code=code_pb2.UNKNOWN, message="Internal Error", details=None
+            )
+            code = code_pb2.UNKNOWN
+
+        return code, response
```
Tehsmash, line 62: Need to make sure we're handling the streaming case too where handler is an async generator. 

``` 
async def call_handler(self, *args, **kwargs):
        """
        Call the handler with the given arguments.
        """

        try:
			if not self.response_streaming:
				response = await self.handler(*args, **kwargs)
           		yield 0, response
			else:
				async for response in await self.handler(*args, **kwargs):
					yield 0, response
			return
        except ErrorResponse as e:
            response = status_pb2.Status(
                code=e.code, message=e.message, details=e.details
            )
            code = e.code
        except Exception:
            response = status_pb2.Status(
                code=code_pb2.UNKNOWN, message="Internal Error", details=None
            )
            code = code_pb2.UNKNOWN

        yield code, response
```

In the server.py it can always be handled as a generator:

```
async for code, response in rpc_handler.call_handler(...):
	...
```

I would also handle the serialization of the request and things in this function too instead of in server.py. So it becomes:

```
async def call_handler(self, request_raw, context):
        """
        Call the handler with the given arguments.
        """

        try:
			request = self.request_deserializer(request_raw)

			# If not streaming, then this function yields once then returns.
            # For streaming it yields each response from the handler
			if not self.response_streaming:
				response = await self.handler(request, context)
				response_bytes = self.response_serializer(response)
           		yield 0, response_bytes
			else:
				async for response in await self.handler(request, context):
					response_bytes = self.response_serializer(response)
					yield 0, response_bytes
			return
```

Issue #569, opened 2025-08-18
Original message id check for Acks


janosSarusiKis:
In the control plane implement a check for the original message id wherever the ACK message is expected. 

Needs more specification.

PR #583, opened 2025-08-21
feat: add original messageID check for Ack

gomba999:
When waitiing for Ack message check for the messageID so that it's returned to the correct caller.


muscariello (approved):

lgecse:
fixes #569

Issue #585, opened 2025-08-24
slimctl version throwing the segmentation fault error


hackeramitkumar:
``slimctl version`` cli cmd is showing segmation fault error. Ideally it should print some version.

lgecse:
@hackeramitkumar please self-assing the issue to yourself.

hackeramitkumar:
@lgecse 
I think I don't have access to issue assignment.

PR #586, opened 2025-08-24
(bug): fixed the slimctl version cmd crash

hackeramitkumar:
# Description

Slimctl version cmd is crashing so as part of this PR we have initialized a Logger and added additional null check for avoiding the crash.

Fixes: #585

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/control-plane/slimctl/internal/cmd/version/version.go
+++ b/control-plane/slimctl/internal/cmd/version/version.go
@@ -66,14 +66,22 @@ func NewVersionCmd(opts *options.CommonOptions) *cobra.Command {
 
 // Run executes the business logic for the version command.
 func Run(v *version, opts *options.CommonOptions, _ *cobra.Command) error {
-	opts.Logger.Info(
-		"Version",
-		zap.String("SemVersion", v.SemVersion),
-		zap.String("GitCommit", v.GitCommit),
-		zap.String("BuildData", v.BuildDate),
-		zap.String("GoVersion", v.GoVersion),
-		zap.String("Compiler", v.Compiler),
-		zap.String("Platform", v.Platform),
+	if opts != nil && opts.Logger != nil {
+		opts.Logger.Info(
+			"Version",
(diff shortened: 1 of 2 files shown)


control-plane/slimctl/internal/cmd/version/version.go
```diff
@@ -66,14 +66,22 @@ func NewVersionCmd(opts *options.CommonOptions) *cobra.Command {
...
-	opts.Logger.Info(
-		"Version",
-		zap.String("SemVersion", v.SemVersion),
-		zap.String("GitCommit", v.GitCommit),
-		zap.String("BuildData", v.BuildDate),
-		zap.String("GoVersion", v.GoVersion),
-		zap.String("Compiler", v.Compiler),
-		zap.String("Platform", v.Platform),
+	if opts != nil && opts.Logger != nil {
+		opts.Logger.Info(
```
hackeramitkumar, line 70: @msardara , @muscariello 
Why are we using Logger.Info here? If we use this, it will also print some irrelevant information.
For example:
```{"level":"info","ts":1756024891.910124,"caller":"version/version.go:70","msg":"Version","SemVersion":"0.1.0-rc0-387-g66aff1c","GitCommit":"66aff1c","BuildDate":"2025-08-24T08:41:16Z","GoVersion":"go1.23.6","Compiler":"gc","Platform":"darwin/arm64"}``` 

Here, level, ts, and caller should not be printed on the console. So I think we should use fmt.Printf(...) instead.
Any opinions ?

msardara: @hackeramitkumar we can go with fmt.Printf(...) for the version print

amitami2: @msardara updated

Issue #601, opened 2025-09-01
SLIM nodes should update Controller about new and existing subsciptions


sancyx:
SLIM nodes should update the Controller about newly created or deleted subscriptions via Southbound API. Control Plane need this to keep track of connected clients.
Current ConfigCommand can be used to send subscriptions, eventually if possible we should add a flag to the Subscription struct in case the client is a Moderator in a Group conversation.

hackeramitkumar:
@sancyx ,
Can you assign this to me?

sancyx:
Related PR https://github.com/agntcy/slim/pull/611 has been merged.

PR #611, opened 2025-09-03
feat: notify controller with new subscriptions

sancyx:
# Description

Send a ConfigCommand message to Controller whenever a new subscription is created or deleted on SLIM node.

Related issue: https://github.com/agntcy/slim/issues/601

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/core/controller/src/config.rs
+++ b/data-plane/core/controller/src/config.rs
@@ -51,7 +51,7 @@ impl Config {
     }
 
     /// Create a ControlPlane service instance from this configuration
-    pub async fn into_service(
+    pub fn into_service(
         &self,
         id: ID,
         rx_drain: drain::Watch,
--- a/data-plane/core/controller/src/service.rs
+++ b/data-plane/core/controller/src/service.rs
@@ -50,7 +50,7 @@ struct ControllerServiceInternal {
     connections: Arc<parking_lot::RwLock<HashMap<String, u64>>>,
 
     /// channel to send messages into the datapath
-    tx_slim: parking_lot::RwLock<mpsc::Sender<Result<PubsubMessage, Status>>>,
+    tx_slim: mpsc::Sender<Result<PubsubMessage, Status>>,
 
     /// channels to send control messages
     tx_channels: parking_lot::RwLock<TxChannels>,
@@ -79,6 +79,10 @@ pub struct ControlPlane {
 
     /// controller
     controller: ControllerService,
+
+    /// channel to receive message from the datapath
+    /// to be used in listen_from_data_plan
+    rx_slim_option: Option<mpsc::Receiver<Result<PubsubMessage, Status>>>,
 }
 
 /// ControllerServiceInternal implements Drop trait to cancel all running listeners and
@@ -112,7 +116,7 @@ impl ControlPlane {
         drain_rx: drain::Watch,
         message_processor: Arc<MessageProcessor>,
     ) -> Self {
-        let (tx, _) = mpsc::channel(128);
+        let (_, tx_slim, rx_slim) = message_processor.register_local_connection(true);
 
         ControlPlane {
             servers,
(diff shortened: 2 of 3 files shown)


data-plane/core/controller/src/config.rs
```diff
@@ -51,7 +51,7 @@ impl Config {
     }
 
     /// Create a ControlPlane service instance from this configuration
-    pub fn into_service(
+    pub async fn into_service(
```
msardara: Does this need to be async? I don't see any await in the function 🤔

data-plane/core/controller/src/service.rs
```diff
@@ -115,8 +112,7 @@ impl ControlPlane {
         drain_rx: drain::Watch,
         message_processor: Arc<MessageProcessor>,
     ) -> Self {
-        // create local connection with the message processor
-        let (_, tx_slim, rx_slim) = message_processor.register_local_connection();
+        let (tx, _) = mpsc::channel(128);
```
msardara: We could keep the creation of the local connection here, and store the rx_slim into an Option. Then we move it out of option in run(). What do you think?

Like:
```rust
let (_, tx_slim, rx_slim) = message_processor.register_local_connection();
let rx_option = Some(rx_slim);

...

let rx = rx_option.take();
```

data-plane/core/controller/src/service.rs
```diff
@@ -50,10 +50,7 @@ struct ControllerServiceInternal {
     connections: Arc<parking_lot::RwLock<HashMap<String, u64>>>,
 
     /// channel to send messages into the datapath
-    tx_slim: mpsc::Sender<Result<PubsubMessage, Status>>,
-
-    /// channel to receive messages from the datapath
-    _rx_slim: mpsc::Receiver<Result<PubsubMessage, Status>>,
+    tx_slim: parking_lot::RwLock<mpsc::Sender<Result<PubsubMessage, Status>>>,
```
msardara: We should be able to remove the Lock here and copy the sender directly.

data-plane/core/controller/src/service.rs
```diff
@@ -171,9 +166,118 @@ impl ControlPlane {
...
         }
 
+        // create local connection with the message processor
+        let (_, tx_slim, rx_slim) = self
+            .controller
+            .inner
+            .message_processor
+            .register_local_connection(true);
+        let inner = self.controller.inner.clone();
+        *inner.tx_slim.write() = tx_slim;
```
msardara: If we register the local connection in the new() method we can remove the lock here.

data-plane/core/datapath/src/message_processing.rs
```diff
@@ -679,11 +702,13 @@ impl MessageProcessor {
         client_config: Option<ClientConfig>,
         cancellation_token: CancellationToken,
         is_local: bool,
+        from_control_plane: bool,
     ) -> tokio::task::JoinHandle<()> {
         // Clone self to be able to move it into the spawned task
         let self_clone = self.clone();
         let token_clone = cancellation_token.clone();
         let client_conf_clone = client_config.clone();
+        let tx_cp: Option<Sender<Result<Message, Status>>> = self.get_tx_control_plane();
```
msardara, line 711: This is retrieved outside of the tokio::spawn, so maybe we can remove the lock and just copy the tx channel

micpapal: The lock there is used only to be able to modify tx_control_plane inside the Arc<MessageProcessorInternal>. In fact the lock is read only in that line and then the tx is cloned and used directly. At the moment I don't see another way to modify that after the creation of the MessageProcessorInternal

Issue #629, opened 2025-09-06
SLIM node ID should be unique in a deployment


sancyx:
Controller uses SLIM node ID sent at node registration to identify SLIM nodes. In case there are multiple SLIM nodes in a deployment, like multiple replicas (Statefulset or Daemonset) they should have unique ID's.

PR #630, opened 2025-09-06
SLIM node ID should be unique in a deployment

sancyx:
# Description

SLIM nodes checks for SLIM_SVC_ID_SUFFIX env var and appends to it's  ID when they register to the Controller.

Add new flag to helm deployment: slim.daemonset. If set to true there will be a SLIM node deployed to each k8s node.
In case  slim.daemonset = false (default) there will be slim.replicaCount number of SLIM nodes  deployed as StatefulSet.
In case of a Daemonset will be set to the name of the node, while in case of SatefulSet will be set with name of the pod running the node, thus SLIm nodes will register with unique ID in the cluster.

Related issue: https://github.com/agntcy/slim/issues/629

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/charts/slim/templates/service.yaml
+++ b/charts/slim/templates/service.yaml
@@ -11,6 +11,7 @@ metadata:
     {{- include "slim.labels" . | nindent 4 }}
 spec:
   type: {{ .Values.slim.service.type }}
+  clusterIP: None
   ports:
     - port: {{ .Values.slim.service.data.port }}
       targetPort: data-plane
--- a/charts/slim/templates/service.yaml
+++ b/charts/slim/templates/service.yaml
@@ -11,7 +11,9 @@ metadata:
     {{- include "slim.labels" . | nindent 4 }}
 spec:
   type: {{ .Values.slim.service.type }}
+  {{- if eq .Values.slim.service.headless true }}
   clusterIP: None
+  {{- end }}
   ports:
     - port: {{ .Values.slim.service.data.port }}
       targetPort: data-plane
--- a/data-plane/core/controller/src/service.rs
+++ b/data-plane/core/controller/src/service.rs
@@ -895,9 +895,9 @@ mod tests {
     async fn test_end_to_end() {
         // Create an ID for slim instance
         let id_server =
-            ID::new_with_name(Kind::new("slim").unwrap(), "test_server_instance").unwrap();
+            ID::new_with_name(Kind::new("slim").unwrap(), "test-server-instance").unwrap();
         let id_client =
-            ID::new_with_name(Kind::new("slim").unwrap(), "test_client_instance").unwrap();
+            ID::new_with_name(Kind::new("slim").unwrap(), "test-client-instance").unwrap();
 
         // Create a server configuration
         let server_config = ServerConfig::with_endpoint("127.0.0.1:50051")
--- a/data-plane/core/service/src/lib.rs
+++ b/data-plane/core/service/src/lib.rs
@@ -410,11 +410,7 @@ impl ComponentBuilder for ServiceBuilder {
         name: &str,
         config: &Self::Config,
     ) -> Result<Self::Component, ComponentError> {
-        let node_name = if config.node_id.is_some() {
-            config.node_id.clone().unwrap()
-        } else {
-            name.to_string()
-        };
+        let node_name = config.node_id.unwrap_or(name.to_string());
         let id = ID::new_with_name(ServiceBuilder::kind(), &node_name)
             .map_err(|e| ComponentError::ConfigError(e.to_string()))?;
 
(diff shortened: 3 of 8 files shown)


data-plane/core/controller/src/service.rs
```diff
@@ -698,7 +698,16 @@ impl ControllerService {
...
-                    node_id: this.inner.id.to_string(),
+                    node_id: {
+                        let mut node_id = this.inner.id.to_string();
+                        if let Ok(suffix) = std::env::var("SLIM_SVC_ID_SUFFIX") {
+                            if !suffix.is_empty() {
+                                node_id.push('-');
+                                node_id.push_str(&suffix);
+                            }
+                        }
+                        node_id
```
msardara: For the node_id, I was thinking that we could modify the service configuration to be able to specify the ID in this way:

```yaml
tracing:
  log_level: debug
  display_thread_names: true
  display_thread_ids: true

runtime:
  n_cores: 0
  thread_name: "slim-data-plane"
  drain_timeout: 10s

services:
  slim/0:
    id: ${env:SLIM_SVC_ID}
    dataplane:
      servers: []
      clients: []
    controller:
      servers: []
```

Notice that specifying a configuration parameter by referring to an external env variable is already supported. The other option is `${file:/path/to/file}`, but I think we don't need it here.

tiswanso:
Offhand it seems strange to make this deployed as a daemonset as a first class supported method in the chart.  I'd think you'd just want to use the default scheduler behavior for statefulsets (default chooses different nodes) or if you want to be more strict, podAntiAffinity rules.

sancyx:
> Offhand it seems strange to make this deployed as a daemonset as a first class supported method in the chart. I'd think you'd just want to use the default scheduler behavior for statefulsets (default chooses different nodes) or if you want to be more strict, podAntiAffinity rules.

By default daemonset is set to false, so it's defaulting to StateFulset.

data-plane/core/service/src/lib.rs
```diff
@@ -406,7 +410,12 @@ impl ComponentBuilder for ServiceBuilder {
         name: &str,
         config: &Self::Config,
     ) -> Result<Self::Component, ComponentError> {
-        let id = ID::new_with_name(ServiceBuilder::kind(), name)
+        let node_name = if config.node_id.is_some() {
+            config.node_id.clone().unwrap()
+        } else {
+            name.to_string()
+        };
```
msardara:
```suggestion
        let node_name = config.node_id.unwrap_or(name.to_string());
```

charts/slim/templates/service.yaml
```diff
@@ -11,6 +11,7 @@ metadata:
     {{- include "slim.labels" . | nindent 4 }}
 spec:
   type: {{ .Values.slim.service.type }}
+  clusterIP: None
```
msardara, line 15: Why do we need a headless service here?

tiswanso:
> > Offhand it seems strange to make this deployed as a daemonset as a first class supported method in the chart. I'd think you'd just want to use the default scheduler behavior for statefulsets (default chooses different nodes) or if you want to be more strict, podAntiAffinity rules.
> 
> By default daemonset is set to false, so it's defaulting to StateFulset.

Understood but my primary concern is having to support (test and maintain) the daemonset deployment type.  Also, from a perception point of view, daemonset deployments it typically implies usage of or dependency on k8s node specific resources (which I would NOT think a SLIM server would need to be tied to).  IMHO, with `podAntiAffinity` it's clearer that the requirement is tied to not have server instances for the same deployment on the same node rather than scaling and dependency for each k8s node in the cluster.

charts/slim/templates/service.yaml
sancyx: Well since it's recommended in case of Statefulset and basically this way a client receives a list of ip's which are direct pod ip's to nodes. Indeed in this case there's no load balancing. May be we can make this configurable as service.type and dont set to headless by default.

sancyx:
> > > Offhand it seems strange to make this deployed as a daemonset as a first class supported method in the chart. I'd think you'd just want to use the default scheduler behavior for statefulsets (default chooses different nodes) or if you want to be more strict, podAntiAffinity rules.
> > 
> > 
> > By default daemonset is set to false, so it's defaulting to StateFulset.
> 
> Understood but my primary concern is having to support (test and maintain) the daemonset deployment type. Also, from a perception point of view, daemonset deployments it typically implies usage of or dependency on k8s node specific resources (which I would NOT think a SLIM server would need to be tied to). IMHO, with `podAntiAffinity` it's clearer that the requirement is tied to not have server instances for the same deployment on the same node rather than scaling and dependency for each k8s node in the cluster.

I would guess that if somebody wants to deploy a SLIM node to each cluster node, he will deploy also clients as Deamonset so that they can connect to the local address. Any other cases probably Statefulset is the way to go.

Issue #607, opened 2025-09-02
Automatically handle SPIRE operations from SLIM node


msardara:
We are currently handling the certificate/jwt retrieval and rotation using the spire-helper, that must run as side car container together with SLIM node.

[SPIRE Helper](https://github.com/spiffe/spiffe-helper)

[SPIRE Helper Sidecar](https://github.com/agntcy/slim/blob/main/charts/slim/templates/deployment.yaml#L39-L54)

This works but:
 - the spire helper should always run with slim-node, and this is not always possible, in particular in non k8s envs
 - if slim starts before the helper, the certs/jwt is not there and SLIM node will panic

It would be nicer if SLIM were able to directly perform the operations performed by the spiffe helper:
 - retrieve SVID certificates / JWT from the SPIFFE Workload API
 - rotate them

This would make easier both the deployment of SLIM and the management of race condition when SLIm starts with the spire-helper sidecar.

We can start checking this repo: https://github.com/maxlambrecht/rust-spiffe

msardara:
@hackeramitkumar would you like to pick this one?

hackeramitkumar:
Yes, I can take this. Please assign this to me.

PR #646, opened 2025-09-09
feat: implementation of Spire for fetching the certificates/token directly from SPIFFE Workload API

hackeramitkumar:
# Description
fixes : #607 

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/Cargo.lock
+++ b/data-plane/Cargo.lock
@@ -1893,7 +1893,7 @@ version = "9.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5a87cc7a48537badeae96744432de36f4be2b4a34a05a5ef32e9dd8a1c169dde"
 dependencies = [
- "base64 0.22.1",
+ "base64",
  "js-sys",
  "pem",
  "ring",
@@ -1996,12 +1996,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "34080505efa8e45a4b816c349525ebe327ceaa8559756f0356cba97ef3bf7432"
 
 [[package]]
-name = "lru-slab"
-version = "0.1.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "112b39cec0b298b6c1999fee3e31427f74f676e4cb9879ed1a121b43661a4154"
-
-[[package]]
 name = "logos"
 version = "0.15.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2034,6 +2028,12 @@ checksum = "605d9697bcd5ef3a42d38efc51541aa3d6a4a25f7ab6d1ed0da5ac632a26b470"
 dependencies = [
  "logos-codegen",
 ]
+
+[[package]]
+name = "lru-slab"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "112b39cec0b298b6c1999fee3e31427f74f676e4cb9879ed1a121b43661a4154"
 
 [[package]]
 name = "macro_rules_attribute"
@@ -2866,28 +2866,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "prost-build"
-version = "0.14.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ac6c3320f9abac597dcbc668774ef006702672474aad53c6d596b62e487b40b1"
-dependencies = [
- "heck 0.4.1",
- "itertools 0.10.5",
- "log",
- "multimap",
- "once_cell",
- "petgraph",
- "prettyplease 0.2.37",
- "prost 0.14.1",
- "prost-types 0.14.1",
- "pulldown-cmark",
- "pulldown-cmark-to-cmark",
- "regex",
- "syn 2.0.106",
- "tempfile",
-]
-
-[[package]]
 name = "prost-derive"
 version = "0.14.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2908,8 +2886,8 @@ checksum = "3c9ae1e4084d9737646934f5ac36a8d77f9a82f962f43e267a382fde4f2a903d"
 dependencies = [
  "logos",
  "miette",
- "prost 0.14.1",
- "prost-types 0.14.1",
+ "prost",
+ "prost-types",
 ]
 
 [[package]]
@@ -2993,9 +2971,9 @@ checksum = "8555716f64c546306ddf3383065dc40d4232609e79e0a4c50e94e87d54f30fb4"
 dependencies = [
  "bytes",
  "miette",
- "prost 0.14.1",
+ "prost",
  "prost-reflect",
- "prost-types 0.14.1",
+ "prost-types",
  "protox-parse",
  "thiserror 2.0.17",
 ]
@@ -3008,33 +2986,13 @@ checksum = "072eee358134396a4643dff81cfff1c255c9fbd3fb296be14bdb6a26f9156366"
 dependencies = [
  "logos",
  "miette",
- "prost-types 0.14.1",
+ "prost-types",
  "thiserror 2.0.17",
 ]
 
 [[package]]
 name = "pulldown-cmark"
 version = "0.13.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1e8bbe1a966bd2f362681a44f6edce3c2310ac21e4d5067a6e7ec396297a6ea0"
-dependencies = [
- "bitflags 2.9.4",
- "memchr",
- "unicase",
-]
-
-[[package]]
-name = "pulldown-cmark-to-cmark"
-version = "21.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e5b6a0769a491a08b31ea5c62494a8f144ee0987d86d670a8af4df1e1b7cde75"
-dependencies = [
- "pulldown-cmark",
-]
-
-[[package]]
-name = "pyo3"
-version = "0.24.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1e8bbe1a966bd2f362681a44f6edce3c2310ac21e4d5067a6e7ec396297a6ea0"
 dependencies = [
@@ -3907,9 +3865,9 @@ dependencies = [
  "jsonwebtoken",
  "log",
  "pkcs8",
- "prost 0.14.1",
- "prost-build 0.14.1",
- "prost-types 0.14.1",
+ "prost",
+ "prost-build",
+ "prost-types",
  "protox",
  "serde",
  "serde_json",
@@ -3919,10 +3877,10 @@ dependencies = [
  "tokio",
  "tokio-stream",
  "tokio-util",
- "tonic 0.14.2",
+ "tonic",
  "tonic-prost",
  "tonic-prost-build",
- "tower 0.5.2",
+ "tower",
  "url",
  "x509-parser",
  "zeroize",
@@ -4375,49 +4333,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "tonic"
-version = "0.14.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "eb7613188ce9f7df5bfe185db26c5814347d110db17920415cf2fbcad85e7203"
-dependencies = [
- "async-trait",
- "axum",
- "base64 0.22.1",
- "bytes",
- "h2 0.4.12",
- "http 1.3.1",
- "http-body 1.0.1",
- "http-body-util",
- "hyper 1.7.0",
- "hyper-timeout",
- "hyper-util",
- "percent-encoding",
- "pin-project",
- "socket2 0.6.0",
- "sync_wrapper 1.0.2",
- "tokio",
- "tokio-stream",
- "tower 0.5.2",
- "tower-layer",
- "tower-service",
- "tracing",
-]
-
-[[package]]
-name = "tonic-build"
-version = "0.14.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4c40aaccc9f9eccf2cd82ebc111adc13030d23e887244bc9cfa5d1d636049de3"
-dependencies = [
- "prettyplease 0.2.37",
- "proc-macro2",
- "prost-build 0.13.5",
- "prost-types 0.13.5",
- "quote",
- "syn 2.0.106",
-]
-
-[[package]]
 name = "tonic-build"
 version = "0.14.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4432,33 +4347,6 @@ dependencies = [
 [[package]]
 name = "tonic-prost"
 version = "0.14.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "66bd50ad6ce1252d87ef024b3d64fe4c3cf54a86fb9ef4c631fdd0ded7aeaa67"
-dependencies = [
- "bytes",
- "prost 0.14.1",
- "tonic 0.14.2",
-]
-
-[[package]]
-name = "tonic-prost-build"
-version = "0.14.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b4a16cba4043dc3ff43fcb3f96b4c5c154c64cbd18ca8dce2ab2c6a451d058a2"
-dependencies = [
- "prettyplease 0.2.37",
- "proc-macro2",
- "prost-build 0.14.1",
- "prost-types 0.14.1",
- "quote",
- "syn 2.0.106",
- "tempfile",
- "tonic-build 0.14.2",
-]
-
-[[package]]
-name = "tonic-tls"
-version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "66bd50ad6ce1252d87ef024b3d64fe4c3cf54a86fb9ef4c631fdd0ded7aeaa67"
 dependencies = [
--- a/data-plane/integrations/mcp-proxy/Cargo.lock
+++ b/data-plane/integrations/mcp-proxy/Cargo.lock
@@ -84,7 +84,7 @@ dependencies = [
  "hyper-util",
  "lazy_static",
  "parking_lot",
- "prost 0.13.5",
+ "prost",
  "protoc-bin-vendored",
  "regex",
  "rustls",
@@ -122,7 +122,7 @@ dependencies = [
  "drain",
  "h2",
  "parking_lot",
- "prost 0.13.5",
+ "prost",
  "protoc-bin-vendored",
  "rand 0.9.2",
  "serde",
@@ -152,7 +152,7 @@ dependencies = [
  "h2",
  "opentelemetry",
  "parking_lot",
- "prost 0.13.5",
+ "prost",
  "protoc-bin-vendored",
  "rand 0.9.2",
  "thiserror 2.0.16",
@@ -734,15 +734,6 @@ name = "colorchoice"
 version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b05b61dc5112cbb17e4b6cd61790d9845d13888356391624cbe7e41efeac1e75"
-
-[[package]]
-name = "concurrent-queue"
-version = "2.5.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4ca0197aee26d1ae37445ee532fefce43251d24cc7c166799f4d46817f1d3973"
-dependencies = [
- "crossbeam-utils",
-]
 
 [[package]]
 name = "const-oid"
@@ -2548,17 +2539,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7231bd9b3d3d33c86b58adbac74b5ec0ad9f496b19d22801d773636feaa95f3d"
 dependencies = [
  "bytes",
- "prost-derive 0.13.5",
-]
-
-[[package]]
-name = "prost"
-version = "0.14.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7231bd9b3d3d33c86b58adbac74b5ec0ad9f496b19d22801d773636feaa95f3d"
-dependencies = [
- "bytes",
- "prost-derive 0.14.1",
+ "prost-derive",
 ]
 
 [[package]]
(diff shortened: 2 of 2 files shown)


data-plane/core/auth/src/spiffe.rs
```diff
@@ -0,0 +1,357 @@
...
+            refresh_interval: None,
+        }
+    }
+}
+
+/// SPIFFE certificate and JWT provider that automatically rotates credentials
+#[derive(Clone)]
+pub struct SpiffeProvider {
+    config: SpiffeConfig,
+    x509_source: Arc<RwLock<Option<Arc<X509Source>>>>,
```
msardara: I think we can remove the lock (and likely the Arc as well), as I see the only method we use of X509Source is get_svid, and it takes an immutable reference. So it should be safe to be called from different thread.

data-plane/core/auth/src/spiffe.rs
```diff
@@ -0,0 +1,357 @@
...
+            .get_svid()
+            .map_err(|e| AuthError::ConfigError(format!("Failed to get X509 SVID: {}", e)))?
+            .ok_or_else(|| AuthError::ConfigError("No X509 SVID available".to_string()))?;
+
+        debug!("Retrieved X509 SVID with SPIFFE ID: {}", svid.spiffe_id());
+        Ok(svid)
+    }
+
+    /// Get a JWT SVID token using the WorkloadApiClient
+    pub async fn get_jwt_svid(&self) -> Result<JwtSvid, AuthError> {
```
msardara: I wonder if we can implement something like the X509Source also for the JWT.

The X509Source internally manages the rotation and when we call get_svid we always get a valid svid.

We could create a similar JWTSource and replicate the behavior by getting inspired from X509Source.

```rust
let svid = jwt_source
            .get_svid()
            .map_err(|e| AuthError::ConfigError(format!("Failed to get JWT SVID: {}", e)))?
            .ok_or_else(|| AuthError::ConfigError("No JWT SVID available".to_string()))?;
```

The nice thing is that get_svid() is a sync method, so we can use it in the sync provider. All the async refresh would be managed inside jwt_source.

data-plane/core/auth/src/spiffe.rs
```diff
@@ -0,0 +1,357 @@
...
+
+        // Convert private key to PEM format
+        let engine = base64::engine::general_purpose::STANDARD;
+        let key_pem = format!(
+            "-----BEGIN PRIVATE KEY-----\n{}\n-----END PRIVATE KEY-----",
+            engine.encode(private_key.as_ref())
+        );
+
+        Ok(key_pem)
+    }
```
msardara: It looks like this code is unused

data-plane/core/auth/src/spiffe.rs
```diff
@@ -0,0 +1,357 @@
...
+
+        // The X509Source handles rotation automatically
+        // We just need to log when credentials are updated
+        info!("SPIFFE credential rotation is handled automatically by X509Source");
+        Ok(())
+    }
+}
+
+impl TokenProvider for SpiffeProvider {
+    fn get_token(&self) -> Result<String, AuthError> {
```
msardara: If we create a JWT source, here we could simply call `self.jwt_source.get_svid()`, which is a sync method.

If the token is not available yet we either retry or return an error.

data-plane/core/auth/src/spiffe.rs
```diff
@@ -0,0 +1,357 @@
...
+        let mut client_guard = self.client.write().await;
+        let client_ref = client_guard.as_mut().ok_or_else(|| {
+            AuthError::ConfigError("WorkloadApiClient not initialized".to_string())
+        })?;
+
+        client_ref
+            .fetch_jwt_bundles()
+            .await
+            .map_err(|e| AuthError::ConfigError(format!("Failed to fetch JWT bundles: {}", e)))
+    }
```
msardara, line 535: Instead of getting the bundles for each verification, which is expensive, we could start an internal task that updates the bundles, by using the stream_jwt_bundle. The bundle needs to be placed in an `Arc<RwLock<Option<JwtBundleSet>>>`:

```rust
let bundles = self.bundles.clone();
tokio::spawn(async move {
    let client = WorkloadApiClient::default().await.map_err(|e| {
        AuthError::ConfigError(format!("Failed to initialize WorkloadApiClient: {}", e))
    })?;
    let mut jwt_bundle_stream = client.stream_jwt_bundles().await?;
    while let Some(jwt_bundle_update) = jwt_bundle_stream.next().await {
        match jwt_bundle_update {
            Ok(update) => {
                *bundles.write() = Some(update);
            }
            Err(e) => {
                // handle the error
            }
        }
    }
```

And then in the verification:

```rust
impl Verifier for SpiffeJwtVerifier {
    async fn verify(&self, token: impl Into<String> + Send) -> Result<(), AuthError> {
        let token_str = token.into();
        let jwt_bundles = self.bundles.read().clone();

        match jwt_bundles {
            Some(bundles) => {
                // Parse and validate the JWT token
                JwtSvid::parse_and_validate(&token_str, &bundles, &self.audiences)
                    .map_err(|e| AuthError::TokenInvalid(format!("JWT validation failed: {}", e)))?;

                debug!("Successfully verified JWT token");
                Ok(())
            }
            None => {
                // manually call jwt_bundle_stream.next().await...
            }
        }
    }
```

This will work also in the try_verify, as it does not need to be async.

data-plane/core/auth/src/spiffe.rs
```diff
@@ -0,0 +1,357 @@
...
+//! This module provides direct integration with SPIFFE Workload API to retrieve
+//! X.509 SVID certificates and JWT tokens.
+
+use async_trait::async_trait;
+use base64::Engine;
+use serde::de::DeserializeOwned;
+use serde_json;
+use std::sync::Arc;
+use std::time::Duration;
+use tokio::sync::RwLock;
```
msardara: I'd suggest you use a parking_lot lock, as it is more performant. You need to pay attention not to call an async function while holding the lock.

data-plane/core/auth/src/spiffe.rs
```diff
@@ -0,0 +1,357 @@
...
+        debug!("Successfully verified JWT token");
+        Ok(())
+    }
+
+    fn try_verify(&self, _token: impl Into<String>) -> Result<(), AuthError> {
+        // SPIFFE verification requires async operations, so this is not supported
+        Err(AuthError::ConfigError(
+            "SPIFFE JWT verification requires async context. Use verify() instead".to_string(),
+        ))
+    }
```
msardara: fn try_verify(&self, _token: impl Into<String>) -> Result<(), AuthError> {
        // Try to get the bundle from self.bundles.read().
        // If empty, you cannot initiate an async operation so you can return an error.
    }

data-plane/core/auth/src/spiffe.rs
```diff
@@ -0,0 +1,357 @@
...
+
+    async fn get_claims<Claims>(&self, token: impl Into<String> + Send) -> Result<Claims, AuthError>
```
msardara, line 571: Also these 2 functions can be rewritten with the same logic of verify and try_verify.

data-plane/core/auth/src/spiffe.rs
hackeramitkumar: added JwtSource implementation

data-plane/core/auth/src/spiffe.rs
hackeramitkumar: done

data-plane/core/auth/src/spiffe.rs
hackeramitkumar: done

data-plane/core/auth/src/spiffe.rs
hackeramitkumar: made changes

data-plane/core/auth/src/spiffe.rs
hackeramitkumar: done

data-plane/core/auth/src/spiffe.rs
hackeramitkumar: done

data-plane/core/auth/src/spiffe.rs
amitami2: done

data-plane/core/auth/src/spiffe.rs
amitami2: done

data-plane/core/auth/src/spiffe.rs
```diff
@@ -0,0 +1,647 @@
...
+                        tokio::time::sleep(sleep_dur).await;
+                    }
+                    Err(err) => {
+                        tracing::warn!(error=%err, "jwt_source: failed to fetch JWT SVID; backing off");
+                        tokio::time::sleep(backoff).await;
+                        backoff = (backoff * 2).min(cfg.max_retry_backoff);
+                    }
+                }
+            }
+        });
```
msardara, line 285: We need to check that this task terminates when the client gets dropped. Otherwise we need to use a [cancellation token](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html) for it.

Also, the sleep between retries should be done in a [tokio::select](https://docs.rs/tokio/latest/tokio/macro.select.html), together with the cancellation token, as we might want to close the task while it is sleeping, and with the current structure the cancellation token won't respond until the sleep ends.

There are [many examples](https://github.com/agntcy/slim/blob/main/data-plane/core/service/src/app.rs#L231-L301) in the code with this pattern.

Summarizing, you need a select between 2 futures:
 - the tokio::sleep, after which you can retrieve the token
 - the cancellation token, after which you can close the task. You can insert a `cancellation_token.cancel();` directly in the Drop trait to close it automatically when the JwtSource gets dropped.

data-plane/core/auth/src/spiffe.rs
```diff
@@ -0,0 +1,647 @@
...
+    // We don't know the exact type of `expiry()` (string formatting earlier), so use a defensive approach.
+    // Try to parse the `expiry().to_string()` as an epoch seconds integer; if that fails default.
+    let default = Duration::from_secs(30);
+    let expiry_str = svid.expiry().to_string();
+    if let Ok(epoch) = expiry_str.parse::<u64>()
+        && let Ok(now_secs) = std::time::SystemTime::now().duration_since(std::time::UNIX_EPOCH)
+        && epoch > now_secs.as_secs()
+    {
+        let remaining = Duration::from_secs(epoch - now_secs.as_secs());
+        return remaining.saturating_sub(buffer).max(Duration::from_secs(5));
```
msardara: I'd renew the token before 5s. Let's say we renew at 2/3 of its duration, so it works both for super short lived tokens with seconds of duration and long lived tokens.

data-plane/core/auth/src/spiffe.rs
```diff
@@ -0,0 +1,647 @@
...
+
+        // Start background task that maintains an in-memory cache of JWT bundles using the streaming API.
+        let bundles_cache = self.bundles.clone();
+        tokio::spawn(async move {
+            // Exponential backoff parameters
+            let mut backoff_secs: u64 = 1;
+            const MAX_BACKOFF: u64 = 30;
+            loop {
+                // Create a fresh client each attempt (avoids mut borrow conflicts with other operations)
+                match WorkloadApiClient::default().await {
```
msardara: I think we don't need to create a client for each attempt. We can just create one client inside the task and use it as local variable, thus making the borrow checker happy.

data-plane/core/auth/src/spiffe.rs
```diff
@@ -0,0 +1,647 @@
...
+            WorkloadApiClient::new_from_path(&path).await
+        } else {
+            WorkloadApiClient::default().await
+        };
+        let mut client = client_res.map_err(|e| {
+            AuthError::ConfigError(format!(
+                "Failed to create WorkloadApiClient for JwtSource: {}",
+                e
+            ))
+        })?;
```
msardara: We can move the client initialization work inside the tokio task directly.

data-plane/core/auth/src/spiffe.rs
```diff
@@ -0,0 +1,647 @@
...
+            const MAX_BACKOFF: u64 = 30;
+            loop {
+                // Create a fresh client each attempt (avoids mut borrow conflicts with other operations)
+                match WorkloadApiClient::default().await {
+                    Ok(mut streaming_client) => {
+                        tracing::debug!("spiffe_jwt: starting JWT bundle stream");
+                        match streaming_client.stream_jwt_bundles().await {
+                            Ok(mut stream) => {
+                                backoff_secs = 1; // reset backoff after successful connection
+                                while let Some(next_item) = stream.next().await {
```
msardara, line 457: Same comment of before, we need to have a way to terminate this task. We can use the cancellation token + tokio::sleep + tokio::select here as well.

data-plane/core/config/src/auth/spiffe.rs
```diff
@@ -0,0 +1,276 @@
...
+
+    /// Target SPIFFE ID for JWT tokens (optional)
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub target_spiffe_id: Option<String>,
+
+    /// JWT audiences for token requests
+    #[serde(default = "default_audiences")]
+    pub jwt_audiences: Vec<String>,
+
+    /// Certificate refresh interval (optional, defaults to 1/3 of certificate lifetime)
```
msardara: Where does it defaul to 1/3 of the cert lifetime?

data-plane/core/config/src/auth/spiffe.rs
```diff
@@ -0,0 +1,276 @@
...
+    #[serde(default)]
+    #[schemars(with = "Option<String>")]
+    pub refresh_interval: Option<Duration>,
+
+    /// Enable X.509 SVID certificate authentication
+    #[serde(default = "default_true")]
+    pub enable_x509: bool,
+
+    /// Enable JWT SVID token authentication
+    #[serde(default = "default_true")]
```
msardara: Let's rename this `default_enable_jwt`

data-plane/core/config/src/auth/spiffe.rs
```diff
@@ -0,0 +1,276 @@
...
+
+    /// Certificate refresh interval (optional, defaults to 1/3 of certificate lifetime)
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(default)]
+    #[schemars(with = "Option<String>")]
+    pub refresh_interval: Option<Duration>,
+
+    /// Enable X.509 SVID certificate authentication
+    #[serde(default = "default_true")]
+    pub enable_x509: bool,
```
msardara, line 34: Let's rename this `default_enable_x509`

data-plane/core/config/src/auth/spiffe.rs
```diff
@@ -0,0 +1,276 @@
...
+    /// Create a SPIFFE JWT verifier from this configuration
+    pub async fn create_jwt_verifier(&self) -> Result<SpiffeJwtVerifier, AuthError> {
+        let config = SpiffeVerifierConfig {
+            socket_path: self.socket_path.clone(),
+            jwt_audiences: self.jwt_audiences.clone(),
+        };
+        let verifier = SpiffeJwtVerifier::new(config);
+        verifier.initialize().await?;
+        Ok(verifier)
+    }
```
msardara, line 127: Consider that we may want to reuse the same config to get a x509 provider/verifier. We can likely add them in another PR where we configure TLS with this.

data-plane/core/config/src/auth/spiffe.rs
```diff
@@ -0,0 +1,276 @@
...
+                "JWT authentication is disabled".to_string(),
+            ));
+        }
+
+        // This is a placeholder - in practice, we'd need to create the verifier asynchronously
+        // The actual implementation would need to be done in an async context
+        Err(AuthError::ConfigError(
+            "SPIFFE server layer creation requires async context".to_string(),
+        ))
+    }
```
msardara, line 166: I think here we should add a method to the provider/verifier traits to allow for async initialization inside the https://github.com/agntcy/slim/blob/main/data-plane/core/auth/src/jwt_middleware.rs, so that we can instantiate them without using async and we async-initialize them the first time we use them. We can handle this separately, the purpose of the PR is to have a first spire integration working.

data-plane/core/auth/src/spiffe.rs
msardara: We could also use `tokio::time::interval` in place of sleep

data-plane/core/config/src/auth/spiffe.rs
amitami2: This I am also planning to take in another PR.

data-plane/core/config/src/auth/spiffe.rs
amitami2: For now in this PR I have added new as sync and the initialization will be asyn but this async func is not part of trait till now. So let's take that a separate task. Since that will also impact our other auth providers.

data-plane/core/config/src/auth/spiffe.rs
amitami2: done

data-plane/core/config/src/auth/spiffe.rs
amitami2: done

data-plane/core/config/src/auth/spiffe.rs
amitami2: removed refresh_interval from the config. By default we will use 2/3 of lifetime as refresh interval

data-plane/core/auth/src/spiffe.rs
amitami2: done

data-plane/core/auth/src/spiffe.rs
amitami2: done

data-plane/core/auth/src/spiffe.rs
amitami2: done

data-plane/core/auth/src/spiffe.rs
amitami2: done

data-plane/core/auth/src/spiffe.rs
```diff
@@ -0,0 +1,1096 @@
...
+                }
+            };
+
+            let mut backoff = cfg.min_retry_backoff;
+
+            // Use interval for timing - start with refresh interval or default
+            // Start with a default interval, but will be dynamically adjusted based on token lifetime
+            let initial_duration = Duration::from_secs(30);
+            let mut interval = tokio::time::interval(initial_duration);
+            interval.tick().await; // consume the first immediate tick
```
msardara: Is there a reason why we are consuming the first tick outside the loop?

data-plane/core/auth/src/spiffe.rs
```diff
@@ -0,0 +1,1096 @@
...
+            return Err(AuthError::ConfigError(
+                "Empty certificate chain".to_string(),
+            ));
+        }
+
+        // Convert the first certificate to PEM format
+        let cert_der = &cert_chain[0];
+        let engine = base64::engine::general_purpose::STANDARD;
+        let cert_pem = format!(
+            "-----BEGIN CERTIFICATE-----\n{}\n-----END CERTIFICATE-----",
```
msardara: I am not sure this will endup in a valid PEM format. In PEM encoding, each line should consists of exactly 64 characters, as specified in [RFC7468](https://www.rfc-editor.org/rfc/rfc7468)

> Generators MUST wrap the base64-encoded lines so that each line
   consists of exactly 64 characters except for the final line, which
   will encode the remainder of the data (within the 64-character line
   boundary), and they MUST NOT emit extraneous whitespace.  Parsers MAY
   handle other line sizes.  These requirements are consistent with PEM
   [[RFC1421](https://www.rfc-editor.org/rfc/rfc1421)].

Fortunately, we already have a function doing this that you can reuse (maybe moving it in a util module) ✅

https://github.com/agntcy/slim/blob/main/data-plane/core/auth/src/testutils.rs#L211-L229

data-plane/core/auth/src/spiffe.rs
```diff
@@ -0,0 +1,1096 @@
...
+
+    /// Get the X.509 private key in PEM format
+    pub fn get_x509_key_pem(&self) -> Result<String, AuthError> {
+        let svid = self.get_x509_svid()?;
+        let private_key = svid.private_key();
+
+        // Convert private key to PEM format
+        let engine = base64::engine::general_purpose::STANDARD;
+        let key_pem = format!(
+            "-----BEGIN PRIVATE KEY-----\n{}\n-----END PRIVATE KEY-----",
```
msardara: Same comment as https://github.com/agntcy/slim/pull/646/files#r2444793662

data-plane/core/auth/src/spiffe.rs
```diff
@@ -0,0 +1,1096 @@
...
+                                }
+                                // Reset backoff on success
+                                backoff = cfg.min_retry_backoff;
+
+                                // Always use automatic 2/3 lifetime calculation
+                                let next_duration = calculate_refresh_interval(&svid);
+
+                                // Reset interval with new duration
+                                interval = tokio::time::interval(next_duration);
+                                interval.tick().await; // consume the first immediate tick
```
msardara: Same question as https://github.com/agntcy/slim/pull/646/files#r2444095702

data-plane/core/auth/src/spiffe.rs
```diff
@@ -0,0 +1,1096 @@
...
+                                tracing::debug!(next_duration_secs = next_duration.as_secs(), "jwt_source: next refresh in {} seconds", next_duration.as_secs());
+                            }
+                            Err(err) => {
+                                tracing::warn!(error=%err, "jwt_source: failed to fetch JWT SVID; backing off");
+                                // Reset interval with backoff duration
+                                interval = tokio::time::interval(backoff);
+                                interval.tick().await; // consume the first immediate tick
```
msardara: Same question as https://github.com/agntcy/slim/pull/646/files#r2444095702

data-plane/core/auth/src/spiffe.rs
hackeramitkumar: added bytes_to_pem in utils mod and now using that in both spiffe and testUtils.

data-plane/core/auth/src/spiffe.rs
hackeramitkumar: done

data-plane/core/auth/src/spiffe.rs
hackeramitkumar: done

data-plane/core/auth/src/spiffe.rs
hackeramitkumar: done

data-plane/core/auth/src/spiffe.rs
hackeramitkumar: I have removed additional ticks. I got confused with this line: https://docs.rs/tokio/latest/tokio/time/index.html#:~:text=Creates%20new%20Interval%20that%20yields%20with%20interval%20of%20period.%20The%20first%20tick%20completes%20immediately. Now I realized that we can remove that, because we don't care about the initial tick since we don't have any token yet.

msardara (approved):
LGTM. Great job @hackeramitkumar!

PR #652, opened 2025-09-10
ci: publish Rust coverage to Codecov using Taskfile

### LINKED ISSUE #649: Coverage report in GH Actions
Generate a report in GH actions and publish repo in codedov.

muscariello:
Adds a GitHub Action to run Rust coverage with the Taskfile and upload results to Codecov. Uses cargo llvm-cov and the lcov.info output. Closes #649 if merged.

### DIFF
--- a/.github/workflows/data-plane-coverage.yaml
+++ b/.github/workflows/data-plane-coverage.yaml
@@ -29,8 +29,8 @@ jobs:
 
       - name: Generate coverage report (lcov)
         run: |
-          ../bin/task data-plane:test:coverage
-        working-directory: ./data-plane
+          cd data-plane
+          task test:coverage
 
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v4
--- a/.github/workflows/data-plane-coverage.yaml
+++ b/.github/workflows/data-plane-coverage.yaml
@@ -34,7 +34,7 @@ jobs:
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v4
         with:
-          files: lcov.info
+          files: data-plane/lcov.info
           flags: rust
           name: rust-data-plane
           fail_ci_if_error: true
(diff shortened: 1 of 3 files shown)


.github/workflows/data-plane-coverage.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v4
+        with:
+          files: ./data-plane/lcov.info
+          flags: rust
+          name: rust-data-plane
+          fail_ci_if_error: true
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
```
msardara: We could incorporate this step in the existing coverage job in https://github.com/agntcy/slim/blob/main/.github/workflows/reusable-rust-build-and-test.yaml#L176-L202

The reusable workflow needs to also be updated to accept the `CODECOV_TOKEN` as secret. Also this step should not run on PRs but only in pushes to main.

muscariello: I need to see this working in the PR first. As soon as the integration works I will remove it I do not like to close and open PR as test.

msardara: Sorry I was not clear :D what I meant is that if we incorporate the step in the reusable-rust-build-and-test, we need to costraint it to pushes to main only, as all the rest of the workflow runs also on PRs.

We can do it by adding a `if: github.event_name == 'push'` to the step once we are sure it works.

muscariello: Codecov produces test coverages on different branches. How do we do that then?

msardara: Yup actually running it on PR is useful as well, I see it can also leave comments regarding how code coverage is affected: https://docs.codecov.com/docs/pull-request-comments

So we can leave the triggers as they are.

muscariello: check the same for agent directory by clicking on the README badge
https://github.com/agntcy/dir

msardara (approved):

Issue #590, opened 2025-08-27
Handle updates from SLIM nodes (affecting subscriptions)


lpuskas:
Design and implement handling of updates related to subscriptions in the controller.
These updates are expected through the SB api.

Describe the approach, eventually create diagrams (alternatively provide a good description of the feature in the PR description)

The feature requires the SLIM node to send the updates to the Controller (Related issue: #TBD)

PR #708, opened 2025-09-17
feat: handle updates from SLIM nodes

sancyx:
# Description

Refactor route handling in Controller and slimctl:

- new RouterService which manages routes either they are created via Northbound API (ex. slimctl) or via Southbound API (slim node subscription updates).
- RouterService starts a number reconciler threads waiting for ReconcileRequest's from a work queue
- whenever a route is created or deleted, first is stored in the db, then a ReconcileRequest with the relevant nodeID is sent to the work queue which is processed by a reconciler. RouteReconciler fetches all routes for a node and sends to the SLIM node in a ConfigCommand. Deleted routes are deleted only when they were successfully sent to the node.
- Node registration also triggers a reconciliation for the given node

Related issue: https://github.com/agntcy/slim/issues/590

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [x] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/core/controller/src/service.rs
+++ b/data-plane/core/controller/src/service.rs
@@ -26,7 +26,7 @@ use crate::api::proto::api::v1::{
 };
 use crate::errors::ControllerError;
 use slim_config::grpc::client::ClientConfig;
-use slim_datapath::api::ProtoMessage as PubsubMessage;
+use slim_datapath::api::ProtoMessage as DataPlaneMessage;
 use slim_datapath::message_processing::MessageProcessor;
 use slim_datapath::messages::Name;
 use slim_datapath::messages::utils::SlimHeaderFlags;
@@ -54,7 +54,7 @@ struct ControllerServiceInternal {
     connections: Arc<parking_lot::RwLock<HashMap<String, u64>>>,
 
     /// channel to send messages into the datapath
-    tx_slim: mpsc::Sender<Result<PubsubMessage, Status>>,
+    tx_slim: mpsc::Sender<Result<DataPlaneMessage, Status>>,
 
     /// channels to send control messages
     tx_channels: parking_lot::RwLock<TxChannels>,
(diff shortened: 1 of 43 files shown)


control-plane/control-plane/internal/config/config.go
```diff
@@ -12,9 +12,10 @@ import (
 
 // Application configuration
 type ControlPlaneConfig struct {
-	Northbound APIConfig `yaml:"northbound"`
-	Southbound APIConfig `yaml:"southbound"`
-	LogConfig  LogConfig `yaml:"logging"`
+	Northbound        APIConfig `yaml:"northbound"`
+	Southbound        APIConfig `yaml:"southbound"`
+	LogConfig         LogConfig `yaml:"logging"`
+	ReconcilerThreads int       `yaml:"reconcilerThreads"` // Number of threads for the route reconciler
```
msardara: Not urgent, but maybe we could use a "ReconcilerConfig" so that we can pass additional parameters if required 🤔

lpuskas (approved):
👍 LGTM!

control-plane/control-plane/internal/services/routes/conn_config.go
msardara, line 1: Not urgent as well, but I I was wondering whether we can generate these types starting from the json schema, using a tool like https://github.com/omissis/go-jsonschema

data-plane/core/controller/src/service.rs
```diff
@@ -96,6 +100,31 @@ impl Drop for ControlPlane {
...
 
+fn from_server_config(server_config: &ServerConfig) -> ConnectionDetails {
+    let group_name = server_config
+        .metadata
+        .as_ref()
+        .and_then(|m| m.get("groupName"))
+        .and_then(|v| match v {
+            MetadataValue::String(s) => Some(s.clone()),
+            _ => None,
+        });
```
msardara: Nice! Very rusty code 🔝

msardara (approved):

control-plane/control-plane/internal/config/config.go
sancyx: created a seprate config stuct for that.

control-plane/control-plane/internal/services/routes/conn_config.go
sancyx: I was looking for something like, this but havent seen this one, will give it a try, definitely would be better to generate.

PR #735, opened 2025-09-23
feat: improve point to point session with sender/receiver buffer

micpapal:
# Description

Improved point-to-point sessions by introducing sender and receiver buffers. This functionality is only used in reliable sessions. 

In Anycast sessions, each side maintains a producer buffer to resend packets if acknowledgments are not received. Out-of-order delivery can occur because messages may be load balanced across multiple instances. 

In Unicast sessions, both producer and receiver buffers are enabled. The producer buffer works the same as in Anycast, while the receiver buffer reorders incoming packets and requests retransmission if packets are missing. 

Also: The term "sticky" has been removed everywhere.

## Type of Change

- [x] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


data-plane/core/service/src/session/point_to_point.rs
```diff
@@ -656,29 +772,280 @@ where
...
+            ProtoSessionType::SessionPointToPoint.into(),
+            ProtoSessionMessageType::P2PAck.into(),
+            message.get_session_header().session_id,
+            message_id,
+            &None,
+            &None,
+        ));
+
+        Message::new_publish_with_headers(slim_header, session_header, "", vec![])
+    }
```
msardara, line 887: Not for this PR, but we should probably consolidate all the creation of these control packet in one place and reuse across sessions.

micpapal: totally agree! I think that most of the code can be shared among the sessions now.

Issue #692, opened 2025-09-15
Add authentication to slimctl


gomba999:
Add support for authentication for the silmctl cli tool to call the controller northbound API via an ingress with basic auth.

PR #763, opened 2025-09-29
feat: add basic auth to slimctl

gomba999:
# Description

Adding basic auth support to slimctl to be used with the control plane or slim nodes.

Related issue: https://github.com/agntcy/slim/issues/692

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/control-plane/common/controlplane/client.go
+++ b/control-plane/common/controlplane/client.go
@@ -5,6 +5,7 @@ package controlplane
 
 import (
 	"context"
+	"encoding/base64"
 	"fmt"
 
 	"google.golang.org/grpc"
@@ -31,9 +32,10 @@ func GetClient(
 		creds = c
 	}
 
-	if opts.BasicAuthKey != "" {
+	if opts.BasicAuthCredentials != "" {
+		encodedAuth := base64.StdEncoding.EncodeToString([]byte(opts.BasicAuthCredentials))
 		md := metadata.New(map[string]string{
-			"authorization": "Basic " + opts.BasicAuthKey,
+			"authorization": "Basic " + encodedAuth,
 		})
 		ctx = metadata.NewOutgoingContext(ctx, md)
 	}
--- a/control-plane/slimctl/cmd/main.go
+++ b/control-plane/slimctl/cmd/main.go
@@ -31,13 +31,13 @@ var k = koanf.New(".")
 func initConfig(opts *options.CommonOptions, flagSet *pflag.FlagSet) error {
 	// defaults
 	defaults := map[string]interface{}{
-		"basic_auth_key": "",
-		"server":         "localhost:50051",
-		"timeout":        "15s",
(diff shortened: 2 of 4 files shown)


control-plane/common/controlplane/client.go
```diff
@@ -4,38 +4,48 @@
...
 		}
 		creds = c
 	}
 
+	if opts.BasicAuthKey != "" {
+		md := metadata.New(map[string]string{
+			"authorization": "Basic " + opts.BasicAuthKey,
+		})
+		ctx = metadata.NewOutgoingContext(ctx, md)
+	}
```
msardara: Probably not for this PR, but we could think of using [client interceptors](https://github.com/grpc/grpc-go/blob/master/examples/features/interceptor/client/main.go#L157) to handle auth and other stuff like rate limiting, logging etc.

control-plane/slimctl/internal/cmd/controller/channel.go
```diff
@@ -21,6 +21,7 @@ func NewChannelCmd(opts *options.CommonOptions) *cobra.Command {
 		Short: "Manage SLIM channels",
 		Long:  `Manage SLIM channels`,
 	}
+	cmd.PersistentFlags().StringP(basicAuthKey, "b", "", "Basic auth key for the control plane authentication")
```
msardara: Any reason not not add it to the root command?

gomba999: My main reason was that I didn't want the users to have the impression that this flag is functional for every command, it only works with commands that call the control plane and not with the node-connect commands. When we'll have a functional ingress for the slim in its chart this can be moved to the root command making things easier.

msardara: We can already configure the controller part of SLIM to accept basic credentials, so maybe it makes sense to have it at the root, so that it can be used out of the box also for node-connect 🤔 WDYT?

gomba999: ahh I see, I wasn't aware of that, I'll move it to the rootcmd then, that'll make things easier.

control-plane/slimctl/cmd/main.go
```diff
@@ -104,6 +106,12 @@ func main() {
 			return nil
 		},
 	}
+	rootCmd.PersistentFlags().StringP(
+		"basic_auth_key",
+		"b",
+		k.String("basic_auth_key"),
+		"Basic auth key for authentication",
+	)
```
msardara, line 114: Should we leave like this and expect the base64 encoded `username:password`, or should we expect it in the form `username:password` and then we do the base64 encoding inside?

msardara (approved):
LGTM, just a comment.

Issue #740, opened 2025-09-23
Notify Controller about deleted connections


sancyx:
At the moment when a client disconnects from a SLIM node, the connection is deleted but the subscription remains there on SLIM node and on the Controller since the Controller is not aware of client disconnect. 
Thus the Controller routing table becomes stale.
SLIM nodes should delete subscriptions related to a dropped connection and notify the Controller sending a ConfigCommand with subscriptions_to_delete param listing the deleted subscriptions.

PR #901, opened 2025-10-30
feat: connection drop controller update

gomba999:
# Description

When a client connection is lost to a slim node, the node updates the controller instance to report which subscriptions need to be removed.

Related issue: https://github.com/agntcy/slim/issues/740

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -589,6 +589,15 @@ impl SubscriptionTable for SubscriptionTableImpl {
...
     }
 
+    fn get_local_subscriptions_on_connection(&self, conn: u64) -> Vec<Name> {
+        let conn_map = self.connections.read();
+        let set = conn_map.get(&conn);
+        if set.is_none() {
+            return Vec::new();
+        }
+        set.unwrap().iter().cloned().collect()
+    }
```
```suggestion
set.map(|s| s.iter().cloned().collect()).unwrap_or_default()
```

data-plane/core/datapath/src/message_processing.rs
```diff
@@ -806,6 +807,28 @@ impl MessageProcessor {
             }
 
             if !connected {
+                // if connection is not local and controller exists, notify about lost subscriptions on the connection
+                if let (false, Some(tx)) = (is_local, tx_cp) {
+                    let subscriptions = self_clone
+                        .forwarder()
+                        .get_local_subscriptions_on_connection(conn_index);
+
+                    if !subscriptions.is_empty() {
```
msardara: I think we can remove this check, if empty the next for loop will not run

data-plane/core/datapath/src/message_processing.rs
```diff
@@ -806,6 +807,28 @@ impl MessageProcessor {
...
+                                "notify control plane about lost subscription: {}",
+                                subscription
+                            );
+                            let msg = Message::new_unsubscribe(
+                                &subscription,
+                                &subscription,
+                                None,
+                                Some(SlimHeaderFlags::default().with_recv_from(conn_index)),
+                            );
+                            let _ = tx.send(Ok(msg)).await;
```
msardara: Let's consider this error and print an error message if it fails

data-plane/core/datapath/src/message_processing.rs
msardara: We are currently doing 2 lookups: one for getting the subscriptions and one for deleting the connection. Maybe we could just return the list of names here and use it to notify the control plane, without calling get_local_subscriptions_on_connection before?

We can still keep the new `get_local_subscriptions_on_connection`, as it might be useful for future ops we might be doing in the future.

gomba999: that was my first idea as well, and I implemented it too but I ended up changing it around, because in `on_connection_drop` both the subscription table and remote subscription tables are cleared and I thought that returning the change from only one of those might not be intuitive. do you think that I should still go ahead and change it back?

msardara: Maybe we could return the 2 and use only the local ones, like:

```rust
let (local_subs, _remote_subs) = self_clone.forwarder().on_connection_drop(conn_index, is_local);
```

data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -589,24 +589,36 @@ impl SubscriptionTable for SubscriptionTableImpl {
...
+
+            for name in &subscriptions_to_remove {
                 let table = self.table.write();
                 debug!("remove subscription {} from connection {}", name, conn);
                 remove_subscription_from_sub_table(name, conn, is_local, table)?;
             }
-        }
+
+            subscriptions_to_remove
+        };
```
msardara: Let's simplify this a bit:

```suggestion
    fn remove_connection(
        &self,
        conn: u64,
        is_local: bool,
    ) -> Result<HashSet<Name>, SubscriptionTableError> {
        let removed_subscriptions = self
            .connections
            .write()
            .remove(&conn)
            .ok_or(SubscriptionTableError::ConnectionIdNotFound)?;

        let mut table = self.table.write();
        for name in &removed_subscriptions {
            debug!("remove subscription {} from connection {}", name, conn);
            remove_subscription_from_sub_table(name, conn, is_local, &mut table)?;
        }

        Ok(removed_subscriptions)
    }
```

You will also need to modify the signature of the function `remove_subscription_from_sub_table` and the code using it a bit:

```diff
diff --git a/data-plane/core/datapath/src/tables/subscription_table.rs b/data-plane/core/datapath/src/tables/subscription_table.rs
index bf6a7d31..0d4f405c 100644
--- a/data-plane/core/datapath/src/tables/subscription_table.rs
+++ b/data-plane/core/datapath/src/tables/subscription_table.rs
@@ -473,7 +473,7 @@ fn remove_subscription_from_sub_table(
     name: &Name,
     conn_index: u64,
     is_local: bool,
-    mut table: RwLockWriteGuard<'_, RawRwLock, HashMap<InternalName, NameState>>,
+    table: &mut RwLockWriteGuard<'_, RawRwLock, HashMap<InternalName, NameState>>,
 ) -> Result<(), SubscriptionTableError> {
     // Convert &Name to &InternalName. This is unsafe, but we know the types are compatible.
     let query_name = unsafe { std::mem::transmute::<&Name, &InternalName>(name) };
@@ -578,9 +578,8 @@ impl SubscriptionTable for SubscriptionTableImpl {
         is_local: bool,
     ) -> Result<(), SubscriptionTableError> {
         {
-            let table = self.table.write();
-
-            remove_subscription_from_sub_table(name, conn, is_local, table)?
+            let mut table = self.table.write();
+            remove_subscription_from_sub_table(name, conn, is_local, &mut table)?
         }
         {
             let conn_table = self.connections.write();
```

msardara (approved):
LGTM, let's just simplify a bit the logic of the subscription table as we are modifying there ✅

Issue #672, opened 2025-09-11
Python bindings and integration coverage


msardara:
Python bindings and integrations like slimrpc and slima2a are currently not considered
in the coverage report.

PR #915, opened 2025-11-05
feat: Adding unit test for slimrpc and include them in coverage report

janosSarusiKis:
# Description

- Adding unit tests to slimrpc and include them in the coverage report.
- Adding  integration test to test more complex functionality 

#672 

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/.cargo/config.toml
+++ b/data-plane/.cargo/config.toml
@@ -10,3 +10,12 @@ rustflags = ["--codegen", "linker=aarch64-linux-gnu-gcc"]
 
 [target.x86_64-unknown-linux-gnu]
 linker="x86_64-linux-gnu-gcc"
+
+[env]
+CC_x86_64_unknown_linux_gnu = "clang-19"
+AR_x86_64_unknown_linux_gnu = "llvm-ar-19"
+STRIP_x86_64_unknown_linux_gnu = "llvm-strip-19"
+
+CC_aarch64_unknown_linux_gnu = "clang-19"
+AR_aarch64_unknown_linux_gnu = "llvm-ar-19"
+STRIP_aarch64_unknown_linux_gnu = "llvm-strip-19"
--- a/data-plane/core/session/src/moderator_task.rs
+++ b/data-plane/core/session/src/moderator_task.rs
@@ -588,7 +588,7 @@ mod tests {
         assert!(!task.task_complete());
 
         task.welcome_start(timer_id + 2)
-            .expect("error on weclome start");
+            .expect("error on welcome start");
         assert!(!task.task_complete());
 
         task.commit_start(timer_id + 3)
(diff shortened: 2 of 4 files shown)


data-plane/python/integrations/slimrpc/tests/test_common.py
```diff
@@ -0,0 +1,161 @@
...
+    def components_strings(self) -> list[str]:
+        return list(self.args)
+
+    def __str__(self) -> str:
+        return "/".join(self.args)
+
+
+@pytest.fixture
+def dummy_pyname_cls() -> type[DummyPyName]:
+    return DummyPyName
```
Tehsmash: No need for a Dummy + MonkeyPatch here, we can just use the slim_bindings.PyName which is the right python type for the functions. Otherwise we're not really testing the functions as they will be used.

data-plane/python/integrations/slimrpc/tests/test_channel.py
```diff
@@ -0,0 +1,12 @@
...
+
+import pytest
+
+from slimrpc import channel
+
+
+def test_compute_deadline(monkeypatch: pytest.MonkeyPatch) -> None:
+    monkeypatch.setattr(channel.time, "time", lambda: 100.0)
+
+    assert channel._compute_deadline(5) == 105.0  # noqa: SLF001
```
Tehsmash: Tbh I think it would be better to test internal functions as part of the larger functions because they may change move around etc. For example we can test the public channel.unary_unary function by mocking the slim local app and then check that local_app.create_session is called, then session.publish is called with the correct metadata including the correct deadline, and then we can mock the response and check that is working as expected too.

For time based tests I would look at pulling in the "freezegun" library which fully mocks the datetime functions in python. This + aiotools.VirtualClock https://github.com/achimnol/aiotools#virtual-clock should allow us to exercise scenario's where the request times out etc without taking the real amount of time.

data-plane/python/integrations/slimrpc/tests/test_integration.py
```diff
@@ -0,0 +1,448 @@
...
+
+    async def delete_session(self, session_id: int) -> None:
+        await self.transport.delete_session(session_id)
+
+    async def subscribe(self, _: FakePyName) -> None:
+        # Subscription handling delegated to FakeTransport
+        return
+
+    def get_id(self) -> str:
+        return str(self.local_name)
```
Tehsmash: Personally I wouldn't try to create fake implementations like this and instead use unittest.mock + MagicMock to create mock versions. This more easily allows us to assert certain functions are called with the correct inputs, and in the right order.

data-plane/python/integrations/slimrpc/tests/test_channel.py
janosSarusiKis: I have added these kind of test in the test_integration.go. Maybe not the best filename, we can change that to a better one.

PR #971, opened 2025-11-19
Spike uniffi bindgen go - udl approach

janosSarusiKis:
# Description

WIP

Spike for testing the uniffi bindgen go pkg to generate go bindings.
This contains the udl approach, the proc macro approach: #979 

These are created to compare the 2 methods.

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/bindings/go/tests/integration_test.go
+++ b/data-plane/bindings/go/tests/integration_test.go
@@ -1,0 +1,279 @@
+package tests
+
+import (
+	"os"
+	"testing"
+	"time"
+
+	slim "github.com/agntcy/slim/bindings/generated/slimbindings"
+)
+
+// TestIntegrationFullFlow tests the complete workflow
+// This test requires a proper SLIM network setup
+func TestIntegrationFullFlow(t *testing.T) {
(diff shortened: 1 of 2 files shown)


data-plane/bindings/common/src/slim_bindings.udl
Tehsmash, line 1: I think we can avoid using the UDL if we use the "proc macro" style e.g. https://mozilla.github.io/uniffi-rs/latest/proc_macro/index.html 

This way we can simply have: `#[uniffi::export]` on the functions which should be exported from the bindings.

data-plane/bindings/common/src/lib.rs
Tehsmash, line 1: Do we need this? I thought the plan was to just annotate / bind the existing types/functions in the binding library without needing wrapper functions/types like we have for python or having specific wrappers for different languages.

janosSarusiKis:
Closing this since we do the proc macro approach

PR #979, opened 2025-11-20
feat: Go bindings generation using uniffi

janosSarusiKis:
# Description

The PR contains:

- Refactoring the bindings rust code to be uniffi compatible. 
- Reorganizing code so core service and the bindings are in a different crate
- Generated go code
- Examples
- Integration test
- Taskfile to build, generate, test and run examples

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/bindings/adapter/src/adapter.rs
+++ b/data-plane/bindings/adapter/src/adapter.rs
@@ -85,9 +85,11 @@ pub fn get_runtime() -> &'static tokio::runtime::Runtime {
 }
 
 /// Initialize the crypto provider
+///
+/// This must be called before any TLS operations. It's safe to call multiple times.
 #[uniffi::export]
 pub fn initialize_crypto_provider() {
-    // Crypto initialization happens automatically in slim_auth
+    slim_config::tls::provider::initialize_crypto_provider();
     // Also initialize the global runtime
     let _ = get_runtime();
 }
--- a/data-plane/python/bindings/src/pyapp.rs
+++ b/data-plane/python/bindings/src/pyapp.rs
@@ -49,7 +49,7 @@ pub struct PyApp {
 struct PyAppInternal {
     /// The adapter instance (uses AuthProvider/AuthVerifier enums internally)
     /// The adapter manages the service internally
-    adapter: Arc<BindingsAdapter>,
+    adapter: BindingsAdapter,
 }
 
 #[gen_stub_pymethods]
@@ -88,7 +88,6 @@ impl PyApp {
                 // IdentityProvider/IdentityVerifier are already AuthProvider/AuthVerifier type aliases
                 // Use BindingsAdapter's complete creation logic
                 BindingsAdapter::new(base_name, provider, verifier, local_service)
-                    .map(Arc::new)
                     .map_err(|e| format!("Failed to create BindingsAdapter: {}", e))
             })
             .map_err(|e: String| PyErr::new::<PyException, _>(e))?;
--- a/data-plane/python/bindings/src/pysession.rs
+++ b/data-plane/python/bindings/src/pysession.rs
@@ -236,7 +236,7 @@ impl PySessionContext {
                 let ctx = PySessionContext {
                     internal: internal_clone,
                 };
-                ctx.publish_to_py_internal(message_ctx, blob, payload_type, metadata)
+                ctx.publish_to_internal(message_ctx, blob, payload_type, metadata)
                     .await
                     .map(PyCompletionHandle::from)
                     .map_err(|e| PyErr::new::<PyException, _>(e.to_string()))
@@ -253,7 +253,7 @@ impl PySessionContext {
             let ctx = PySessionContext {
                 internal: internal_clone,
             };
-            ctx.invite_py_internal(name)
+            ctx.invite_internal(name)
                 .await
                 .map(PyCompletionHandle::from)
                 .map_err(|e| PyErr::new::<PyException, _>(e.to_string()))
@@ -269,7 +269,7 @@ impl PySessionContext {
             let ctx = PySessionContext {
                 internal: internal_clone,
             };
-            ctx.remove_py_internal(name)
+            ctx.remove_internal(name)
                 .await
                 .map(PyCompletionHandle::from)
                 .map_err(|e| PyErr::new::<PyException, _>(e.to_string()))
@@ -349,7 +349,7 @@ impl PySessionContext {
     }
 
     /// Publish a message as a reply to a received message
-    async fn publish_to_py_internal(
+    async fn publish_to_internal(
         &self,
         message_ctx: PyMessageContext,
         blob: Vec<u8>,
@@ -366,7 +366,7 @@ impl PySessionContext {
     }
 
     /// Invite a participant to this session (multicast only)
-    async fn invite_py_internal(&self, name: PyName) -> PyResult<CompletionHandle> {
+    async fn invite_internal(&self, name: PyName) -> PyResult<CompletionHandle> {
         self.internal
             .bindings_ctx
             .invite_internal(&name.into())
@@ -375,7 +375,7 @@ impl PySessionContext {
     }
 
     /// Remove a participant from this session (multicast only)
-    async fn remove_py_internal(&self, name: PyName) -> PyResult<CompletionHandle> {
+    async fn remove_internal(&self, name: PyName) -> PyResult<CompletionHandle> {
         self.internal
             .bindings_ctx
             .remove_internal(&name.into())
--- a/data-plane/bindings/go/generated/slim_uniffi/slim_uniffi.go
+++ b/data-plane/bindings/go/generated/slim_uniffi/slim_uniffi.go
@@ -378,7 +378,7 @@ func uniffiCheckChecksums() {
 		checksum := rustCall(func(_uniffiStatus *C.RustCallStatus) C.uint16_t {
 			return C.uniffi_slim_uniffi_checksum_func_initialize_crypto_provider()
 		})
-		if checksum != 2455 {
+		if checksum != 45020 {
 			// If this happens try cleaning and rebuilding your project
 			panic("slim_uniffi: uniffi_slim_uniffi_checksum_func_initialize_crypto_provider: UniFFI API checksum mismatch")
 		}
@@ -3783,6 +3783,8 @@ func GetVersion() string {
 }
 
 // Initialize the crypto provider
+//
+// This must be called before any TLS operations. It's safe to call multiple times.
 func InitializeCryptoProvider() {
 	rustCall(func(_uniffiStatus *C.RustCallStatus) bool {
 		C.uniffi_slim_uniffi_fn_func_initialize_crypto_provider(_uniffiStatus)


data-plane/bindings/adapter/src/adapter.rs
```diff
@@ -1,96 +1,376 @@
...
-    app: Arc<App<P, V>>,
+// Re-export uniffi for proc macros
+use uniffi;
+
+// ============================================================================
+// UniFFI Type Definitions
+// ============================================================================
+
+/// Global Tokio runtime for async operations
+static GLOBAL_RUNTIME: OnceLock<Arc<tokio::runtime::Runtime>> = OnceLock::new();
```
msardara: Do we need an Arc here? I think we can go with

```suggestion
static GLOBAL_RUNTIME: OnceLock<tokio::runtime::Runtime> = OnceLock::new();
```

As the runtime is static and no one else should own it. This also means that normally we don't need to save it as part of the completion handle or the session context. To keep things as they are, we can save a `&'static tokio::runtime::Runtime` in place of `Arc<tokio::runtime::Runtime>`

data-plane/bindings/adapter/src/adapter.rs
```diff
@@ -1,96 +1,376 @@
...
+
+/// Get or initialize the global Tokio runtime
+///
+/// Configured for FFI workloads with:
+/// - Worker threads: 2x CPU cores (to handle blocking operations better)
+/// - Max blocking threads: 512 (allows high concurrency from FFI calls)
+/// - Named threads for easier debugging
+fn get_runtime() -> &'static Arc<tokio::runtime::Runtime> {
```
msardara:
```suggestion
fn get_runtime() -> &'static tokio::runtime::Runtime {
```

data-plane/bindings/adapter/src/adapter.rs
```diff
@@ -1,96 +1,376 @@
...
 
-    /// Channel receiver for notifications from the app
-    notification_rx: Arc<RwLock<mpsc::Receiver<Result<Notification, SessionError>>>>,
+/// Initialize the crypto provider
+#[uniffi::export]
+pub fn initialize_crypto() {
+    // Crypto initialization happens automatically in slim_auth
+    // Also initialize the global runtime
+    let _ = get_runtime();
 }
```
msardara, line 95: Let's make it explicit and update the function name (calling `initialize_crypto_provider` multiple times is fine):

```suggestion
pub fn initialize_crypto_and_runtime() {
    use slim_config::tls::provider::initialize_crypto_provider;
    
    initialize_crypto_provider();
    get_runtime();
}
```

data-plane/core/service/src/bindings/adapter.rs
```diff
@@ -1,96 +1,376 @@
...
-        app: App<P, V>,
-        notification_rx: mpsc::Receiver<Result<Notification, SessionError>>,
-    ) -> Self {
-        Self {
-            app: Arc::new(app),
-            notification_rx: Arc::new(RwLock::new(notification_rx)),
+/// Get the version of the SLIM bindings
+#[uniffi::export]
+pub fn get_version() -> String {
+    env!("CARGO_PKG_VERSION").to_string()
```
msardara: This is the version of the `agntcy-slim-service` crate, we should probably move the uniffi-bindings to a separate crate and give its own version, which will then define the version of each language binding.

data-plane/bindings/adapter/src/adapter.rs
```diff
@@ -116,418 +396,1096 @@ where
...
+/// ```
+#[derive(uniffi::Object)]
+pub struct FfiCompletionHandle {
+    /// Receiver for the completion result (can only be consumed once)
+    receiver: Arc<
+        parking_lot::Mutex<
+            Option<tokio::sync::oneshot::Receiver<Result<(), slim_session::SessionError>>>,
+        >,
+    >,
+    runtime: Arc<tokio::runtime::Runtime>,
```
msardara:
```suggestion
    runtime: &'static tokio::runtime::Runtime,
```

data-plane/bindings/adapter/Cargo.toml
```diff
@@ -0,0 +1,36 @@
+[package]
+name = "agntcy-slim-uniffi"
```
msardara, line 2: If we call it uniffi, let's also name the folder that way.

data-plane/bindings/adapter/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+edition = { workspace = true }
+license = { workspace = true }
+version = "0.1.0"
```
msardara: This should start with version 0.7.0, to align bindings with the slin binary version.

data-plane/bindings/adapter/src/adapter.rs
```diff
@@ -0,0 +1,1841 @@
...
+pub fn initialize_crypto() {
+    // Crypto initialization happens automatically in slim_auth
+    // Also initialize the global runtime
+    let _ = get_runtime();
+}
+
+/// Get the version of the SLIM bindings
+#[uniffi::export]
+pub fn get_version() -> String {
+    env!("CARGO_PKG_VERSION").to_string()
```
msardara, line 113: Here we might want to give more information such as the commit hash, the date etc.

Check https://github.com/agntcy/slim/blob/main/data-plane/python/bindings/build.rs

data-plane/bindings/adapter/src/adapter.rs
```diff
@@ -0,0 +1,1841 @@
...
+    app: Arc<App<AuthProvider, AuthVerifier>>,
+
+    /// Channel receiver for notifications from the app
+    notification_rx: Arc<RwLock<mpsc::Receiver<Result<Notification, SlimSessionError>>>>,
+
+    /// Service reference for lifecycle management
+    service_ref: ServiceRef,
+
+    /// Tokio runtime for blocking async operations
+    runtime: Arc<tokio::runtime::Runtime>,
```
msardara:
```suggestion
    runtime: &'static tokio::runtime::Runtime,
```

data-plane/bindings/adapter/src/adapter.rs
```diff
@@ -0,0 +1,1841 @@
...
+        completion.await.map_err(|e| SlimError::SessionError {
+            message: format!("Session establishment failed: {}", e),
+        })?;
+
+        // Convert SessionContext to BindingsSessionContext
+        let bindings_ctx = crate::BindingsSessionContext::from(session_ctx);
+
+        Ok(Arc::new(FFISessionContext {
+            inner: bindings_ctx,
+            runtime: Arc::clone(&self.runtime),
```
msardara:
```suggestion
            runtime: self.runtime,
```

data-plane/bindings/adapter/src/adapter.rs
```diff
@@ -0,0 +1,1841 @@
...
+pub fn get_version() -> String {
+    env!("CARGO_PKG_VERSION").to_string()
+}
+
+/// Name type for SLIM (Secure Low-Latency Interactive Messaging)
+#[derive(uniffi::Record)]
+pub struct Name {
+    pub components: Vec<String>,
+    pub id: Option<u64>,
+}
```
msardara, line 132: Not urgent for now, but we could also define an impl block with constructors and useful accessors to prevent users from defining incomplete objects.

data-plane/bindings/adapter/src/adapter.rs
```diff
@@ -0,0 +1,1841 @@
...
+        SlimError::SessionError {
+            message: err.to_string(),
+        }
+    }
+}
+
+/// TLS configuration for server/client
+#[derive(uniffi::Record)]
+pub struct TlsConfig {
+    pub insecure: bool,
```
msardara, line 253: We are missing many parameters for the tls configuration. As follow up, we need to think if there is a clever way to expose all the fields in the configuration structs without redefining everything 🤔

data-plane/bindings/adapter/src/adapter.rs
```diff
@@ -0,0 +1,1841 @@
...
+
+    // Wrap in enum types for flexible auth support
+    let provider = AuthProvider::SharedSecret(shared_secret_impl.clone());
+    let verifier = AuthVerifier::SharedSecret(shared_secret_impl);
+
+    let adapter = tokio::task::spawn_blocking(move || {
+        BindingsAdapter::new(slim_name, provider, verifier, false)
+    })
+    .await
+    .map_err(|e| ServiceError::ConfigError(format!("Task join error: {}", e)))??;
```
msardara:
```suggestion
    let adapter = BindingsAdapter::new(slim_name, provider, verifier, false)?;
```

data-plane/bindings/adapter/src/adapter.rs
```diff
@@ -0,0 +1,1841 @@
...
+/// Create an app with the given name and shared secret (blocking version for FFI)
+///
+/// This is the main entry point for creating a SLIM application from language bindings.
+#[uniffi::export]
+pub fn create_app_with_secret(
+    app_name: Name,
+    shared_secret: String,
+) -> Result<Arc<BindingsAdapter>, SlimError> {
+    let runtime = get_runtime();
+    runtime.block_on(async { create_app_with_secret_async(app_name, shared_secret).await })
```
msardara, line 312: The `create_app_with_secret_async` should not be async I think:

```suggestion
    create_app_with_secret_async(app_name, shared_secret)
```

data-plane/bindings/adapter/src/adapter.rs
```diff
@@ -0,0 +1,1841 @@
...
+}
+
+/// Create an app with the given name and shared secret (async version)
+async fn create_app_with_secret_async(
```
msardara, line 316:
```suggestion
fn create_app_with_secret_async(
```

data-plane/bindings/adapter/src/adapter.rs
```diff
@@ -0,0 +1,1841 @@
...
+    /// - SharedSecret
+    /// - JWT (JwtSigner/JwtVerifier)
+    /// - SPIRE (SpireIdentityManager)
+    /// - StaticToken
+    pub fn new(
+        base_name: SlimName,
+        identity_provider: AuthProvider,
+        identity_verifier: AuthVerifier,
+        use_local_service: bool,
+    ) -> Result<Arc<Self>, SlimError> {
```
msardara: Is there a reason to return an Arc here? If needed we can wrap it in an Arc later, but let's return `Self` itself (pun not intended 🤣) here

data-plane/bindings/adapter/src/adapter.rs
```diff
@@ -0,0 +1,1841 @@
...
+                .ok_or_else(|| SlimError::SessionError {
+                    message: "Session already closed or dropped".to_string(),
+                })?;
+
+        self.app
+            .delete_session(&session_ref)
+            .map(|_| ())
+            .map_err(|e| SlimError::SessionError {
+                message: format!("Failed to delete session: {}", e),
+            })
```
msardara: Here the return of delete_session is a completion handle, which we should await  here in the same way we do for create_session. We could also add an optional timeout as follow up.

data-plane/bindings/adapter/src/adapter.rs
```diff
@@ -0,0 +1,1841 @@
...
+        parking_lot::Mutex<
+            Option<tokio::sync::oneshot::Receiver<Result<(), slim_session::SessionError>>>,
+        >,
+    >,
+    runtime: Arc<tokio::runtime::Runtime>,
+}
+
+impl FfiCompletionHandle {
+    /// Create a new FFI completion handle from a Rust CompletionHandle
+    fn new(handle: slim_session::CompletionHandle, runtime: Arc<tokio::runtime::Runtime>) -> Self {
```
msardara:
```suggestion
    fn new(handle: slim_session::CompletionHandle, runtime: &'static tokio::runtime::Runtime) -> Self {
```

data-plane/bindings/adapter/src/adapter.rs
```diff
@@ -0,0 +1,1841 @@
...
+            receiver
+                .await
+                .map_err(|_| SlimError::InternalError {
+                    message: "Completion sender dropped before result was sent".to_string(),
+                })?
+                .map_err(|e| SlimError::SessionError {
+                    message: e.to_string(),
+                })
+        })
+    }
```
msardara, line 923: Instead of duplicating the code, we can maybe just block_on the async version.

data-plane/bindings/adapter/src/adapter.rs
```diff
@@ -0,0 +1,1841 @@
...
+// ============================================================================
+// FFI SessionContext Wrapper
+// ============================================================================
+
+/// FFISessionContext represents an active session (FFI-compatible wrapper)
+#[derive(uniffi::Object)]
+pub struct FFISessionContext {
+    /// The inner BindingsSessionContext (public for Python bindings access)
+    pub inner: crate::BindingsSessionContext,
+    runtime: Arc<tokio::runtime::Runtime>,
```
msardara:
```suggestion
    runtime: &'static tokio::runtime::Runtime,
```

data-plane/bindings/adapter/src/adapter.rs
```diff
@@ -0,0 +1,1841 @@
...
+    }
+
+    /// Publish a reply message with delivery confirmation (async version)
+    pub async fn publish_to_with_completion_async(
+        &self,
+        message_context: MessageContext,
+        data: Vec<u8>,
+        payload_type: Option<String>,
+        metadata: Option<std::collections::HashMap<String, String>>,
+    ) -> Result<Arc<FfiCompletionHandle>, SlimError> {
```
msardara: Do we need the Arc here?

data-plane/bindings/adapter/src/adapter.rs
```diff
@@ -0,0 +1,1841 @@
...
+/// Client configuration for connecting to a SLIM server
+#[derive(uniffi::Record)]
+pub struct ClientConfig {
+    pub endpoint: String,
+    pub tls: TlsConfig,
+}
+
+/// Message context for received messages
+#[derive(uniffi::Record)]
+pub struct MessageContext {
```
msardara: Is there a reason not to reuse the original message_context defined in this crate?

data-plane/bindings/adapter/src/adapter.rs
msardara: Given that we don't need to hold a reference to the runtime, do we still have a reason to duplicate the SessionContext here?

data-plane/python/bindings/src/pyapp.rs
```diff
@@ -151,10 +175,17 @@ impl PyApp {
         let internal_clone = self.internal.clone();
 
         pyo3_async_runtimes::tokio::future_into_py(py, async move {
+            // Convert to FFI ServerConfig
+            let ffi_config = slim_uniffi::ServerConfig {
+                endpoint: config.endpoint,
+                tls: slim_uniffi::TlsConfig {
+                    insecure: config.tls_setting.insecure,
+                },
+            };
```
msardara, line 190: In the current python bindings, we get a python dict as input that we convert into the original rust grpc configuration via serde, which can be used to spin up a server given the whole configuration.

The ffi version works but does not expose all the required config fields, we need to think how to do this. Possible options I can think about:

 - Replicate all fields in the new struct (costly but allow to expose a native object to the foreign language)
 - Get the config in json format and try to convert it with serde_json (we don't need to care about syncing the options, but caller should provide a json string)

data-plane/bindings/adapter/src/adapter.rs
```diff
@@ -0,0 +1,1841 @@
...
+            id: Some(name.id()),
+        }
+    }
+}
+
+/// Session type enum
+#[derive(uniffi::Enum)]
+pub enum SessionType {
+    PointToPoint,
+    Multicast,
```
micpapal: minor issue. we need to rename Multicast to Group as now this is visible to the app developer

data-plane/bindings/adapter/Cargo.toml
janosSarusiKis: As discussed in meeting, we do the renaming in a follow-up PR.

data-plane/bindings/adapter/src/adapter.rs
janosSarusiKis: fixed the wait and will do the timeout in the follow up

data-plane/python/bindings/src/pyapp.rs
janosSarusiKis: I added all the fields to the config

data-plane/bindings/adapter/src/adapter.rs
janosSarusiKis: will do this in a follow up PR

data-plane/bindings/adapter/src/adapter.rs
msardara: As second thought, some AuthProvider might start a tokio task, which is not an async call but it requires to be run inside a runtime (unless the spawn is called as member function of a specific runtime).

data-plane/python/bindings/src/pysession.rs
```diff
@@ -361,34 +360,31 @@ impl PySessionContext {
...
             .bindings_ctx
-            .publish_to(&ctx, blob, payload_type, metadata)
+            .publish_to_internal(&ctx, blob, payload_type, metadata)
             .await
             .map_err(|e| PyErr::new::<PyException, _>(e.to_string()))
     }
 
     /// Invite a participant to this session (multicast only)
-    async fn invite_internal(&self, name: PyName) -> PyResult<CompletionHandle> {
+    async fn invite_py_internal(&self, name: PyName) -> PyResult<CompletionHandle> {
```
msardara:
```suggestion
    async fn invite_internal(&self, name: PyName) -> PyResult<CompletionHandle> {
```

data-plane/python/bindings/src/pysession.rs
```diff
@@ -344,13 +343,13 @@ impl PySessionContext {
...
             .bindings_ctx
-            .publish(&target_name, fanout, blob, conn_out, payload_type, metadata)
+            .publish_internal(&target_name, fanout, blob, conn_out, payload_type, metadata)
             .await
             .map_err(|e| PyErr::new::<PyException, _>(e.to_string()))
     }
 
     /// Publish a message as a reply to a received message
-    async fn publish_to_internal(
+    async fn publish_to_py_internal(
```
msardara:
```suggestion
    async fn publish_to_internal(
```

data-plane/python/bindings/src/pysession.rs
```diff
@@ -361,34 +360,31 @@ impl PySessionContext {
...
             .bindings_ctx
-            .invite(&name.into())
+            .invite_internal(&name.into())
             .await
             .map_err(|e| PyErr::new::<PyException, _>(e.to_string()))
     }
 
     /// Remove a participant from this session (multicast only)
-    async fn remove_internal(&self, name: PyName) -> PyResult<CompletionHandle> {
+    async fn remove_py_internal(&self, name: PyName) -> PyResult<CompletionHandle> {
```
msardara:
```suggestion
    async fn remove_internal(&self, name: PyName) -> PyResult<CompletionHandle> {
```

data-plane/python/bindings/src/pysession.rs
msardara:
```suggestion
    async fn publish_to_internal(
```

data-plane/python/bindings/src/pysession.rs
```diff
@@ -254,7 +253,7 @@ impl PySessionContext {
             let ctx = PySessionContext {
                 internal: internal_clone,
             };
-            ctx.invite_internal(name)
+            ctx.invite_py_internal(name)
```
msardara:
```suggestion
            ctx.invite_internal(name)
```

data-plane/python/bindings/src/pysession.rs
```diff
@@ -270,7 +269,7 @@ impl PySessionContext {
             let ctx = PySessionContext {
                 internal: internal_clone,
             };
-            ctx.remove_internal(name)
+            ctx.remove_py_internal(name)
```
msardara:
```suggestion
            ctx.remove_internal(name)
```

data-plane/python/bindings/src/pysession.rs
```diff
@@ -237,7 +236,7 @@ impl PySessionContext {
                 let ctx = PySessionContext {
                     internal: internal_clone,
                 };
-                ctx.publish_to_internal(message_ctx, blob, payload_type, metadata)
+                ctx.publish_to_py_internal(message_ctx, blob, payload_type, metadata)
```
msardara:
```suggestion
                ctx.publish_to_internal(message_ctx, blob, payload_type, metadata)
```

data-plane/python/bindings/src/pyapp.rs
```diff
@@ -27,23 +26,30 @@ use crate::utils::PyName;
...
-{
-    /// The adapter instance
-    adapter: BindingsAdapter<P, V>,
-
-    /// Reference to the service
-    service: ServiceRef,
+struct PyAppInternal {
+    /// The adapter instance (uses AuthProvider/AuthVerifier enums internally)
+    /// The adapter manages the service internally
+    adapter: Arc<BindingsAdapter>,
```
msardara: The Arc here can likely be removed

PR #987, opened 2025-11-21
fix: remove participants and ack management in controller sender

micpapal:
# Description

This PR fixes a critical issue where removing participants from group sessions would hang indefinitely and never complete. The problem occurred due to improper acknowledgment tracking and missing cleanup logic in the controller sender.

Root Causes Fixed
- **Incorrect ID reset logic in acknowledgment tracking**: The controller sender was incorrectly calling reset_id() on participant names for all message types when tracking pending acknowledgments. This should only be done for DiscoveryReply messages, causing acks from removed participants to be mismatched and never cleared from the pending list.

- **Missing receiver-side cleanup**: When removing an endpoint, only the sender was being cleaned up. The receiver's message buffer for that endpoint was not being cleared, which could cause issues if the same participant rejoined later (messages would be incorrectly dropped as duplicates).

- **Potential panic on task completion failure**: The moderator could attempt to access current_task when it was None during failure handling.

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


data-plane/core/session/src/session_receiver.rs
```diff
@@ -320,6 +320,13 @@ impl SessionReceiver {
             .await
     }
 
+    pub fn remove_endpoint(&mut self, endpoint: &Name) {
+        // remove the buffer raleted to an endpoint so that if it is added again
```
msardara:
```suggestion
        // remove the buffer related to an endpoint so that if it is added again
```

data-plane/core/session/src/session_receiver.rs
```diff
@@ -320,6 +320,13 @@ impl SessionReceiver {
...
+        // the messages will not be dropped as duplicated
+        tracing::info!("remove endpoint on the receiver {}", endpoint);
```
msardara:
```suggestion
        tracing::debug!("remove endpoint on the receiver {}", endpoint);
```

data-plane/core/session/src/session_moderator.rs
```diff
@@ -191,20 +191,23 @@ where
...
+                            ),
+                            ModeratorTask::Update(task) => signal_failure(
+                                &mut task.ack_tx,
+                                "failed to update state of the participant",
+                            ),
+                            ModeratorTask::Close(task) => {
+                                signal_failure(&mut task.ack_tx, "failed to close the session")
+                            }
+                        };
+                    }
```
msardara: Let's add some accessor in the moderator task like this:

```rust
impl ModeratorTask {
    pub(crate) fn ack_tx_take(
        &mut self,
    ) -> Option<oneshot::Sender<Result<(), SessionError>>> {
        match self {
            ModeratorTask::Add(t) => t.ack_tx.take(),
            ModeratorTask::Remove(t) => t.ack_tx.take(),
            ModeratorTask::Close(t) => t.ack_tx.take(),
            ModeratorTask::Update(t) => t.ack_tx.take(),
        }
    }

    pub(crate) fn failure_message<'a>(&self, add_msg: &'a str) -> &'a str {
        match self {
            ModeratorTask::Add(_) => add_msg,
            ModeratorTask::Remove(_) => "failed to remove a participant from the group",
            ModeratorTask::Update(_) => "failed to update state of the participant",
            ModeratorTask::Close(_) => "failed to close the session",
        }
    }
}
```

And simplify this with:


```suggestion
                    if let Some(task) = self.current_task.as_mut() {
                        if let Some(ack_tx) = task.ack_tx_take() {
                            let _ = ack_tx.send(Err(SessionError::ModeratorTask(
                                task.failure_message(message).to_string(),
                            )));
                        }
                    }
```

We can also delete the signal_failure at line 179 I think.

data-plane/core/session/src/controller_sender.rs
```diff
@@ -94,7 +89,13 @@ impl ControllerSender {
                 }
                 let mut missing_replies = HashSet::new();
                 let mut name = message.get_dst();
-                name.reset_id();
+                if message.get_session_message_type()
+                    == slim_datapath::api::ProtoSessionMessageType::DiscoveryRequest
+                {
+                    // the discovery reply should be sent to an unknown destination id.
```
msardara:
```suggestion
                    // the discovery request should be sent to an unknown destination id.
```

data-plane/core/session/src/session_moderator.rs
```diff
@@ -803,18 +806,12 @@ where
...
-        //for (k, v) in self.group_list.iter() {
-        //    let name = k.clone().with_id(*v);
-        //    participants.push(name);
-        //}
-        let participants = self.group_list.keys().cloned().collect();
+        let mut participants = vec![];
+        for (k, v) in self.group_list.iter() {
+            let name = k.clone().with_id(*v);
+            participants.push(name);
+        }
```
msardara:
```suggestion
        let participants: Vec<_> = self
            .group_list
            .iter()
            .map(|(k, v)| k.clone().with_id(*v))
            .collect();
```

data-plane/core/session/src/controller_sender.rs
```diff
@@ -108,7 +109,9 @@ impl ControllerSender {
                 let mut missing_replies = HashSet::new();
                 for n in &payload.participants {
                     let name = Name::from(n);
-                    missing_replies.insert(name);
+                    if name != self.local_name {
+                        missing_replies.insert(name);
+                    }
                 }
```
msardara:
```suggestion
                let missing_replies: HashSet<Name> = payload
                    .participants
                    .iter()
                    .map(Name::from)
                    .filter(|n| *n != self.local_name)
                    .collect();
```

msardara (changes requested):
LGTM, with small nits ✅

Issue #970, opened 2025-11-19
Make backoff retry configurable


gomba999:
Make the backoff retry configurable via the config file for the data plane.

PR #991, opened 2025-11-24
feat: make backoff retry configurable

gomba999:
# Description

Make backoff retry configurable via the config file, it's used to connect to the controller.

Related issue: https://github.com/agntcy/slim/issues/970

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/config/base/client-config.yaml
+++ b/data-plane/config/base/client-config.yaml
@@ -18,4 +18,8 @@ services:
         - endpoint: "http://localhost:46357"
           tls:
             insecure: true
+          backoff:
+            type: fixed_interval
+            interval: 2000ms
+            max_attempts: 5
       servers: []
--- a/data-plane/core/config/src/backoff/exponential.rs
+++ b/data-plane/core/config/src/backoff/exponential.rs
@@ -1,3 +1,5 @@
+use crate::backoff::default_max_attempts;
+
 use super::Strategy;
 use duration_string::DurationString;
 use schemars::JsonSchema;
@@ -12,7 +14,28 @@ pub struct Config {
     factor: u64,
     #[schemars(with = "String")]
     max_delay: DurationString,
+    #[serde(default = "default_max_attempts")]
+    max_attempts: usize,
+    #[serde(default)]
     jitter: bool,
+}
+
+impl Config {
+    pub fn new(
+        base: u64,
+        factor: u64,
+        max_delay: Duration,
+        max_attempts: usize,
+        jitter: bool,
+    ) -> Self {
+        Config {
+            base,
+            factor,
+            max_delay: max_delay.into(),
+            max_attempts,
+            jitter,
+        }
+    }
 }
 
 impl Default for Config {
@@ -21,6 +44,7 @@ impl Default for Config {
             base: 100,
             factor: 1,
             max_delay: Duration::from_millis(1000).into(),
+            max_attempts: default_max_attempts(),
             jitter: true,
         }
     }
@@ -28,19 +52,12 @@ impl Default for Config {
 
 impl Strategy for Config {
     fn get_strategy(&self) -> Box<dyn Iterator<Item = Duration> + Send> {
-        if self.jitter {
-            Box::new(
-                ExponentialBackoff::from_millis(self.base)
-                    .factor(self.factor)
-                    .max_delay(self.max_delay.into())
-                    .map(jitter),
-            )
-        } else {
-            Box::new(
-                ExponentialBackoff::from_millis(self.base)
-                    .factor(self.factor)
-                    .max_delay(self.max_delay.into()),
-            )
-        }
+        let ret = ExponentialBackoff::from_millis(self.base)
+            .factor(self.factor)
+            .max_delay(self.max_delay.into())
+            .take(self.max_attempts);
+        let jitter_flag = self.jitter;
+
+        Box::new(ret.map(move |d| if jitter_flag { jitter(d) } else { d }))
     }
 }
--- a/data-plane/core/config/src/backoff/fixedinterval.rs
+++ b/data-plane/core/config/src/backoff/fixedinterval.rs
@@ -1,4 +1,5 @@
 use super::Strategy;
+use crate::backoff::default_max_attempts;
 use duration_string::DurationString;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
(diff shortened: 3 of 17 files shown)


data-plane/core/config/src/backoff/exponential.rs
```diff
@@ -0,0 +1,44 @@
...
+    }
+}
+
+impl Strategy for Config {
+    fn get_strategy(&self) -> Box<dyn Iterator<Item = Duration> + Send> {
+        if self.jitter {
+            Box::new(
+                ExponentialBackoff::from_millis(self.base)
+                    .factor(self.factor)
+                    .max_delay(Duration::from_secs(self.max_delay))
```
micpapal: This max_delay should in be in milliseconds as well I think as the default is set to 1000. use duration directly in config so that you don't have any problem

data-plane/core/config/src/backoff/exponential.rs
```diff
@@ -0,0 +1,44 @@
...
+                    .map(jitter),
+            )
+        } else {
+            Box::new(
+                ExponentialBackoff::from_millis(self.base)
+                    .factor(self.factor)
+                    .max_delay(Duration::from_secs(self.max_delay)),
```
micpapal: same here

data-plane/core/config/src/backoff/exponential.rs
```diff
@@ -0,0 +1,44 @@
...
+    factor: u64,
+    max_delay: u64,
+    jitter: bool,
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Config {
+            base: 100,
+            factor: 2,
```
micpapal: I would put 1 as default which is also the default in tokio_retry

data-plane/core/config/src/backoff/fixedinterval.rs
```diff
@@ -0,0 +1,22 @@
+use super::Strategy;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+use std::time::Duration;
+use tokio_retry::strategy::FixedInterval;
+
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, JsonSchema)]
+#[serde(default)]
+pub struct Config {
+    interval: u64,
```
micpapal: Also here is better to use duration

data-plane/core/config/src/backoff/exponential.rs
```diff
@@ -0,0 +1,46 @@
...
+
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, JsonSchema)]
+#[serde(default)]
+pub struct Config {
+    base: u64,
+    factor: u64,
+    #[schemars(with = "String")]
+    max_delay: DurationString,
+    jitter: bool,
+}
```
msardara, line 21: We should add also the max number of attempts (`max_attempts`), which defaults to usize::MAX.

As we aredealing with an iterator, we can use the take() method of the iterator to limit the iterations to max_attempts.


```suggestion
pub struct Config {
    base: u64,
    factor: u64,
    #[serde(default = "default_max_attempts")]
    max_attempts: usize,
    #[schemars(with = "String")]
    max_delay: DurationString,
    jitter: bool,
}

fn default_max_attempts() -> usize {
    usize::MAX
}
```

data-plane/core/config/src/backoff/exponential.rs
```diff
@@ -0,0 +1,46 @@
...
+
+impl Default for Config {
+    fn default() -> Self {
+        Config {
+            base: 100,
+            factor: 1,
+            max_delay: Duration::from_millis(1000).into(),
+            jitter: true,
+        }
```
msardara, line 49:
```suggestion
    fn default() -> Self {
        Config {
            base: 100,
            factor: 1,
            max_attempts: default_max_attempts(),
            max_delay: Duration::from_millis(1000).into(),
            jitter: true,
        }
```

data-plane/core/config/src/backoff/exponential.rs
```diff
@@ -0,0 +1,46 @@
...
+                    .map(jitter),
+            )
+        } else {
+            Box::new(
+                ExponentialBackoff::from_millis(self.base)
+                    .factor(self.factor)
+                    .max_delay(self.max_delay.into()),
+            )
+        }
+    }
```
msardara, line 62:
```suggestion
    fn get_strategy(&self) -> Box<dyn Iterator<Item = Duration> + Send> {
        let ret = ExponentialBackoff::from_millis(self.base)
            .factor(self.factor)
            .max_delay(self.max_delay.into())
            .take(self.max_attempts);
        let jitter_flag = self.jitter;

        Box::new(ret.map(move |d| if jitter_flag { jitter(d) } else { d }))
    }
```

data-plane/core/controller/src/service.rs
```diff
@@ -1512,11 +1510,7 @@ impl ControllerService {
     ) -> Result<mpsc::Sender<Result<ControlMessage, Status>>, ControllerError> {
         info!(%config.endpoint, "connecting to control plane");
 
-        let backoff_strategy = ExponentialBackoff::from_millis(1000)
-            .factor(2)
-            .max_delay(Duration::from_secs(10))
-            .map(jitter);
-
+        let backoff_strategy = config.backoff.get_strategy();
```
msardara: Here I think the caller should not really care about the backoff strategy set the client config. We can probably move this logic inside the grpc channel builder.

msardara (changes requested):

data-plane/config/tls/client-config.yaml
```diff
@@ -15,7 +15,7 @@ services:
   slim/0:
     dataplane:
       clients:
-        - endpoint: "https://localhost:46357"
+        - endpoint: "https://1.2.3.4:46357"
```
msardara:
```suggestion
        - endpoint: "https://localhost:46357"
```

data-plane/config/base/client-config.yaml
```diff
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 tracing:
-  log_level: info
+  log_level: debug
```
msardara:
```suggestion
  log_level: info
```

data-plane/core/controller/src/service.rs
msardara: Calling `config.to_channel()` will now perform the required configured retries, no need to do it here as well

Issue #1005, opened 2025-12-01
Impelement SLIM RPC functionality in data-plane core


sancyx:
Move current SLIM RPC implementation https://github.com/agntcy/slim/tree/main/data-plane/python/integrations/slimrpc from python to rust code base in data-plane.

msardara:
Closing as completed in in #1202

PR #1036, opened 2025-12-08
feat: port slimrpc code to rust lang

sancyx:
# Description

Slimrpc code ported from python to Rust.
Related issue: https://github.com/agntcy/slim/issues/1005

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/bindings/rust/src/slimrpc/channel.rs
+++ b/data-plane/bindings/rust/src/slimrpc/channel.rs
@@ -1,6 +1,40 @@
 // Copyright AGNTCY Contributors (https://github.com/agntcy)
 // SPDX-License-Identifier: Apache-2.0
 
+//! RPC Channel implementation for SlimRPC
+//!
+//! This module implements async bidirectional streaming for RPC calls, inspired by
+//! Go's goroutine-based concurrency model:
+//!
+//! ## Async Streaming Architecture
+//!
+//! ### Client-side (RpcChannel):
+//! - **Unary-Unary**: Send single request, await single response
+//! - **Unary-Stream**: Send single request, receive response stream
+//! - **Stream-Unary**: Send request stream synchronously, await single response
+//! - **Stream-Stream**: Send requests asynchronously in background task (tokio::spawn),
+//!   receive responses concurrently via stream - enables true bidirectional streaming
+//!
+//! ### Server-side (RpcServer):
+//! - Each session is handled in a separate tokio task (like Go goroutines)
+//! - Request streams are read asynchronously using `spawn_request_reader` task
+//! - Response streams are sent asynchronously as they're generated
+//! - No buffering of responses - streaming happens in real-time
+//!
+//! ## Example: Stream-Stream Flow
+//!
+//! ```text
+//! Client:                          Server:
+//!   |                                |
+//!   | spawn send task --------------->|
+//!   | (sends requests async)         | spawn request reader
+//!   |                                | (reads requests async)
+//!   |                                |
+//!   | await response stream <---------|
+//!   | (receives concurrently)        | generate responses
+//!   |                                | (streamed as produced)
+//! ```
+
 use super::common::{service_and_method_to_name, DEADLINE_KEY, MAX_TIMEOUT};
 use super::context::MessageContext;
 use super::error::{Result, SRPCError};
@@ -237,6 +271,7 @@ impl RpcChannel {
     }
 
     /// Internal helper for stream-stream calls
+    /// Sends requests asynchronously while receiving responses concurrently (true bidirectional streaming)
     pub async fn stream_stream(
         &self,
         method: &str,
@@ -248,9 +283,10 @@ impl RpcChannel {
             self.common_setup(method, metadata).await?;
         let deadline = compute_deadline(timeout);
 
-        self.send_stream(request_stream, &session, &service_name, metadata, deadline)
-            .await?;
+        // Send requests asynchronously in background task (like Go goroutine)
+        let _send_handle = self.send_stream_async(request_stream, &session, &service_name, metadata, deadline);
 
+        // Immediately start receiving responses (concurrent with sending)
         let stream = self.receive_stream(session_ctx, deadline).await?;
 
         Ok(stream)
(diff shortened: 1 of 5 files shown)


data-plane/python/integrations/uv.lock
```diff
@@ -1,5 +1,5 @@
 version = 1
-revision = 3
```
msardara, line 2: Let's not lower the revision of uv here.

data-plane/Cargo.toml
```diff
@@ -12,6 +11,7 @@ members = [
     "core/session",
     "core/signal",
     "core/slim",
+    "core/slimrpc",
```
msardara: I think we should move the crate at the same level of `slimrpc-compiler`

data-plane/core/slimrpc/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+agntcy-slim = { workspace = true }
+agntcy-slim-service = { workspace = true }
+agntcy-slim-auth = { workspace = true }
+agntcy-slim-config = { workspace = true }
+agntcy-slim-datapath = { workspace = true }
+agntcy-slim-session = { workspace = true }
+agntcy-slim-signal = { workspace = true }
+
+# Optional: for protobuf support
+prost = { version = "0.12", optional = true }
```
msardara: Also the version of prost can be taken by the workspace.

data-plane/core/slimrpc/Cargo.toml
```diff
@@ -0,0 +1,45 @@
+[package]
+name = "agntcy-slimrpc"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+anyhow = "1.0"
```
msardara: As we are using `thiserror`, we should not use `anyhow`. Anyhow can be used in the program using the crate

data-plane/core/slimrpc/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+async-trait = "0.1"
+clap = { version = "4.5", features = ["derive"] }
+futures = "0.3"
+serde = "1.0"
+serde_yaml = "0.9"
+thiserror = "1.0"
+tokio = { version = "1", features = ["full"] }
+tokio-stream = "0.1"
+tracing = "0.1"
+tracing-subscriber = "0.3"
```
msardara: We have the version for all this dependencies defined in the workspace Cargo.toml, let's use that.

data-plane/core/slimrpc/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+
+# Optional: for protobuf support
+prost = { version = "0.12", optional = true }
+
+[dev-dependencies]
+tokio-test = "0.4"
+
+[features]
+default = []
+protobuf = ["prost"]
```
msardara: Are these features used?

data-plane/core/slimrpc/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+agntcy-slim-config = { workspace = true }
+agntcy-slim-datapath = { workspace = true }
+agntcy-slim-session = { workspace = true }
+agntcy-slim-signal = { workspace = true }
+
+# Optional: for protobuf support
+prost = { version = "0.12", optional = true }
+
+[dev-dependencies]
+tokio-test = "0.4"
```
msardara:
```suggestion
tokio-test = { workspace = true }
```

msardara (changes requested):

data-plane/python/integrations/uv.lock
sancyx: fixed

data-plane/Cargo.toml
sancyx: move to data-plane folder.

data-plane/core/slimrpc/Cargo.toml
sancyx: fixed

data-plane/core/slimrpc/Cargo.toml
sancyx: fixed

data-plane/core/slimrpc/Cargo.toml
sancyx: fixed

data-plane/core/slimrpc/Cargo.toml
sancyx: removed

data-plane/core/slimrpc/Cargo.toml
sancyx: fixed

data-plane/slimrpc/src/server.rs
```diff
@@ -0,0 +1,428 @@
...
+                    .clone();
+
+                let mut response_stream = h.call(request, msg_ctx, session_context).await?;
+                let mut responses = Vec::new();
+
+                while let Some(response) = response_stream.next().await {
+                    responses.push(response?);
+                }
+
+                Ok(responses)
```
Tehsmash: This isn't quite right, what will happen here is that the stream will get drained into responses, and then returned to handle_session, which will then drain it into the SLIM channel.

Instead call_handler should return the stream itself to handle_session, and as items are drained from the stream they should be sent on the SLIM channel.

In python we handled this by having call_handler return a generator of responses regardless of stream or not stream. 

Immediate thought is that call_handler should create a channel, and a background task to do the handling, then return the channel to session_handler or perhaps we can use a futures:Stream.

Honestly I would revisit this whole logic as we're in a different language for example:

session_handler can create two different channels one for  inputs and one for responses, these are passed to call_handler:

For UnaryUnary:
  * session_handler reads one message and push it down the input channel
  * call_hander reads one message from the input channel
  * calls the handler
  * sends the response down the output channel then closes it

For UnaryStream:
  * session_handler reads one message and push it down the input channel
  * call_hander reads one message from the input channel
  * calls the handler
  * drains the response stream down the output channel then closes it

For StreamUnary:
  * session_handler creates a routine/task which continuously reads messages and pushes them down the input channel
  * call_handler passes the stream to the handler
  * waits for one response from the call_handler function
  * sends the response down the output channel then closes it

For StreamStream:
  * session_handler creates a routine/task which continuously reads messages and pushes them down the input channel
   * call_handler passes the stream to the handler
   * drains the response stream down the output channel then closes it

session_handler's main "thread" will drain the output channel, forwarding each message, and then when the channel closes / ends it will send the channel end response to the client.

data-plane/bindings/rust/src/slimrpc/rpc.rs
```diff
@@ -0,0 +1,75 @@
...
+
+#[async_trait]
+pub trait UnaryUnaryHandler<Req, Res>: Send + Sync {
+    async fn call(
+        &self,
+        request: Req,
+        msg_ctx: MessageContext,
+        session_ctx: SessionContext,
+    ) -> Result<Res>;
+}
```
Tehsmash, line 34: In order for this to work with uniffi, we need to make sure that we enable the trait to be implemented by the foreign/binding language explictly: https://mozilla.github.io/uniffi-rs/latest/foreign_traits.html

data-plane/bindings/rust/src/slimrpc/rpc.rs
```diff
@@ -0,0 +1,75 @@
...
+
+#[async_trait]
+pub trait UnaryStreamHandler<Req, Res>: Send + Sync {
+    async fn call(
+        &self,
+        request: Req,
+        msg_ctx: MessageContext,
+        session_ctx: SessionContext,
+    ) -> Result<ResponseStream<Res>>;
+}
```
Tehsmash, line 44: I'm not sure that the function declaration here is correct, in python we had call implement a "generator" essentially the function gets wrapped by the runtime into an iterator class which isn't a concept that exists in rust. This current declaration would require the call function to create and return a stream and trigger downstream a background task to populate the stream.

It would be better for Stream handlers to receive an output stream channel rx from the session handler, so that they can simply call `output_stream.try_send(...)` or add we function on the session_ctx to send responses which then pushes responses onto the channel internally (this is how gRPC implements it BTW).

Tehsmash: For reference this is the gRPC interface for the session context:
https://github.com/grpc/grpc/blob/master/src/python/grpcio/grpc/aio/_base_server.py#L154

read is used to read messages for "Stream****" channels and write is used to response to "***Stream" channels.

msardara:
Closing this PR as done in #1202

PR #1037, opened 2025-12-08
fix(session): correctly remove routes on session close

### LINKED ISSUE #1033: Correctly remove routes on session close
When a session closes, both the moderator and participant remove all subscriptions and routes to the remote node and group. This causes issues when multiple sessions depend on the same entry in the forwarding table. If one session closes while another is still active, the remaining session will fail because the required state is no longer available to support it.

micpapal:
# Description

This PR fixes session route cleanup so that routes are correctly removed when sessions close. In particular the fix introduces a ref count for each route so that if multiple sessions use the same route, the route is removed only if the ref count goes to 0.

The PR improves the testing for P2P application to test 2 sessions in parallel.

fix #1033

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/core/session/src/session_routes.rs
+++ b/data-plane/core/session/src/session_routes.rs
@@ -7,7 +7,7 @@ use std::collections::HashMap;
 use std::sync::Arc;
 
 // each route is composed by a name and the related connection id
-// this allows as to distinguish between multiple routes to the same name.
+// this allows us to distinguish between multiple routes to the same name.
 // the route boolean indicates whether it's a route (true) or a subscription (false)
 #[derive(Hash, Eq, PartialEq, Clone)]
 pub(crate) struct Route {


data-plane/core/session/src/session_routes.rs
```diff
@@ -0,0 +1,392 @@
+// Copyright AGNTCY Contributors (https://github.com/agntcy)
+// SPDX-License-Identifier: Apache-2.0
+
+use parking_lot::RwLock;
+use slim_datapath::messages::Name;
+use std::collections::HashMap;
+use std::sync::Arc;
+
+// each route is compose by a name and the related connection id
```
msardara:
```suggestion
// each route is composed by a name and the related connection id
```

data-plane/core/session/src/session_routes.rs
```diff
@@ -0,0 +1,392 @@
...
+// this allows as to distinguish between multiple routes to the same name.
```
msardara:
```suggestion
// this allows us to distinguish between multiple routes to the same name.
```

msardara (approved):
LGTM

Issue #807, opened 2025-10-08
Ack messages from the control plane


micpapal:
Sessions created by the control plane must be acknowledged back to the control plane to confirm successful creation. This should be done also for invites/removes and channel delete

We need to reuse the same code of sender/receiver created in #800

PR #1050, opened 2025-12-15
feat: send group acknowledge from the session

gomba999:
# Description

Instead of sending back an ack directly from the controller after having received a command from the control plane the moderator will signal back to the controller whether the command execution was successful.

Fix: #807 

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/core/session/src/session_moderator.rs
+++ b/data-plane/core/session/src/session_moderator.rs
@@ -1511,7 +1511,6 @@ mod tests {
             .unwrap();
         let result = moderator.process_control_message(join_msg, None).await;
         assert!(result.is_ok());
-        // println!("{:?}", rx_slim.recv().await.unwrap().unwrap());
         let subscribe_msg = rx_slim.recv().await.unwrap().unwrap();
         assert!(
             matches!(


data-plane/core/session/src/moderator_task.rs
```diff
@@ -158,17 +168,22 @@ pub struct AddParticipant {
     join: State,
     welcome: State,
     commit: State,
+    ack_msg: Message,
```
micpapal: The ack_msg should be always an option as it is used only is the command was triggered by the controller.

data-plane/core/session/src/session_moderator.rs
```diff
@@ -516,23 +510,43 @@ where
...
                 // here we assume that the destination is reachable from the
                 // same connection from where we got the message from the controller
+
+                // src = controller
+                // dst = moderator
+                // payload = participant to add
+
+                // src -> moderator
+                // dst -> participant to add
+
```
micpapal: this comment can be removed

data-plane/core/controller/src/service.rs
```diff
@@ -213,13 +215,16 @@ impl ControlPlane {
             .collect();
 
         let (signal, watch) = drain::channel();
-
+        let controller_name = Name::from_strings(["controller", "controller", "controller"])
+            .with_id(rand::random::<u64>());
+        info!("CREATE CONTROLLER WITH NAME: {}", controller_name);
```
micpapal: change to debug and use lowercase letters

data-plane/core/controller/src/service.rs
```diff
@@ -213,13 +215,16 @@ impl ControlPlane {
             .collect();
 
         let (signal, watch) = drain::channel();
-
+        let controller_name = Name::from_strings(["controller", "controller", "controller"])
```
micpapal: put the string "controller" in a const

data-plane/core/datapath/src/message_processing.rs
```diff
@@ -731,7 +731,11 @@ impl MessageProcessor {
                                         // 3. the control plane exists
                                         if !is_local && !from_control_plane && tx_cp.is_some(){
                                             match msg.get_type() {
-                                                PublishType(_) => {/* do nothing */}
+                                                PublishType(_) => {
+                                                    if msg.get_session_message_type() == GroupAck {
+                                                         let _ = tx_cp.as_ref().unwrap().send(Ok(msg.clone())).await;
```
micpapal: Here I think that you must check the destination name of the message as the group ack may be for a remote node connected to this SLIM instance. You have no guarantees that the destination is the local controller. The best way is to use the subscription table because the controller subscribes for its name

data-plane/core/session/src/session_moderator.rs
```diff
@@ -440,15 +440,39 @@ where
...
+                    Some(c) => {
+                        self.common
+                            .add_route(&message.get_source(), c)
+                            .await
+                            .map_err(|e| self.handle_task_error(e))?;
+                    }
+                    None => {
+                        debug!(
+                            "JoinRequest message missing incoming connection info: {:?}",
+                            message
```
micpapal: here you should return and do not send the ack that will be lost anyway

micpapal (approved):

Issue #1053, opened 2025-12-16
fix(session): errors coming from dataplane should be routed to correct session


msardara:
At the moment errors from dataplane are returned back to the session layer that forwards them to the app.

This works but it is difficult for the app to correlate the error with the session that originated it.

We should re-route the error to the correct session, which will handle it.

PR #1056, opened 2025-12-17
fix(session): route dataplane errors to correct session

msardara:
# Description

Re-route dataplane errors to the correct session, which will handle it.

Fixes: #1053 

## Type of Change

- [x] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/core/service/src/service.rs
+++ b/data-plane/core/service/src/service.rs
@@ -845,15 +845,9 @@ mod tests {
         info!("starting test_error_routing_with_session_context");
 
         // Create the service
-        let tls_config = TlsServerConfig::new().with_insecure(true);
-        let server_config =
-            ServerConfig::with_endpoint("0.0.0.0:12348").with_tls_settings(tls_config);
-        let config = ServiceConfiguration::new().with_server([server_config].to_vec());
-        let service = config
-            .build_server(
-                ID::new_with_name(Kind::new(KIND).unwrap(), "test-error-routing").unwrap(),
-            )
-            .unwrap();
+        let service = Service::new(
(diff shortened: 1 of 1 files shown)


data-plane/core/datapath/src/errors.rs
```diff
@@ -37,10 +39,18 @@ pub enum DataPathError {
...
 
     // Processing
+    #[error("malformed message")]
+    MalformedMessage(#[from] MessageError),
     #[error("message processing error: {0}")]
     ProcessingError(MessageError),
     #[error("error adding connection to connection table")]
     ConnectionTableAddError,
+    #[error("message processing error: {source}")]
+    MessageWithContext {
```
micpapal: This name does not look like an error. Maybe MessageProcessingError?

data-plane/core/session/src/session.rs
```diff
@@ -88,6 +88,15 @@ impl Session {
...
                     .await
             }
+            SessionMessage::MessageError { error } => {
+                // Received error from lower layer. Cancel any pending timers for this message and send error up
+                debug!(
+                    name = %self.local_name,
+                    "received message error: {:?}", error,
+                );
+
+                self.sender.on_slim_failure(error)
```
micpapal, line 98: This error is always for the a message sent by the app right? so there is no need to check the receiver side for this reason I assume

micpapal (approved):

data-plane/core/session/src/session.rs
msardara: Yup, maybe new future errors will require a check in the receiver, bir for now we are good

Issue #1028, opened 2025-12-05
Allow SLIM nodes to communicate via Unix socket


msardara:
Currently, SLIM nodes communicate via gRPC over TCP sockets, using localhost for local connections. While this approach works well for remote clients, it is not optimal when multiple agents run on the same host. In such cases, we can improve efficiency by switching to gRPC over Unix domain sockets for local communication.

Advantages of Using Unix Sockets over TCP/localhost:

- **Lower Latency and Overhead**:
Unix sockets avoid the network stack, resulting in faster data transfer and reduced CPU usage compared to TCP, even when using localhost.

- **Improved Security:**
Unix sockets can leverage filesystem permissions for access control, restricting which processes can connect, and reducing the attack surface compared to TCP ports.

- **Resource Efficiency:**
Unix sockets use fewer system resources and file descriptors than TCP sockets, which is especially beneficial when managing many concurrent local connections.

- **No Network Configuration Needed**:
Unix sockets do not require IP addresses or open ports, simplifying deployment and minimizing potential network conflicts on the host.

Switching to gRPC over Unix domain sockets for intra-host communication will enhance performance, security, and scalability when running multiple SLIM agents locally.

Reference example: https://github.com/hyperium/tonic/tree/master/examples/src/uds

msardara:
@hackeramitkumar would you like to pick it up?

hackeramitkumar:
@msardara sure, I will assign this to me.

PR #1060, opened 2025-12-21
feat: add support of gRPC using the uds for the local communication

hackeramitkumar:
# Description

As part of this change PR added support of UDS(unix domain sockets) for the local communication.
Fixes: #1028 

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/core/config/src/grpc/client.rs
+++ b/data-plane/core/config/src/grpc/client.rs
@@ -6,20 +6,15 @@ use rustls_pki_types::ServerName;
 use tokio_retry::RetryIf;
 
 use display_error_chain::ErrorChainExt;
-#[cfg(target_family = "unix")]
-use hyper_util::rt::TokioIo;
-use std::error::Error as StdErrorTrait;
-#[cfg(target_family = "unix")]
-use std::path::PathBuf;
-#[cfg(target_family = "unix")]
-use std::sync::Arc;
-use std::time::Duration;
-use std::{collections::HashMap, str::FromStr};
-#[cfg(target_family = "unix")]
-use tokio::net::UnixStream;
+use std::{collections::HashMap, error::Error as StdErrorTrait, str::FromStr, time::Duration};
 use tower::ServiceExt;
 #[cfg(target_family = "unix")]
-use tower::service_fn;
+use {
+    hyper_util::rt::TokioIo,
+    std::{path::PathBuf, sync::Arc},
+    tokio::net::UnixStream,
+    tower::service_fn,
+};
 
 use base64::prelude::*;
 use http::header::{HeaderMap, HeaderName, HeaderValue};
@@ -786,7 +781,6 @@ impl ClientConfig {
                         tracing::warn!(error = %err.chain(), "Transport error encountered. Retrying...");
                         true
                     }
-                    #[cfg(target_family = "unix")]
                     ConfigError::UnixSocketConnect(err) => {
                         tracing::warn!(error = %err, "Unix socket connect error encountered. Retrying...");
                         true
--- a/data-plane/core/config/tests/e2e.rs
+++ b/data-plane/core/config/tests/e2e.rs
@@ -99,15 +99,11 @@ mod tests {
     use slim_config::testutils::helloworld::greeter_server::GreeterServer;
     use slim_testing::utils::setup_test_jwt_resolver;
     #[cfg(unix)]
-    use tokio::net::UnixListener;
-    #[cfg(unix)]
-    use tokio::task::yield_now;
-    #[cfg(unix)]
-    use tokio::time::timeout;
-    #[cfg(unix)]
-    use tokio_stream::wrappers::UnixListenerStream;
-    #[cfg(unix)]
-    use tonic::transport::Server;
+    use {
+        tokio::{net::UnixListener, task::yield_now, time::timeout},
+        tokio_stream::wrappers::UnixListenerStream,
+        tonic::transport::Server,
+    };
 
     static TEST_DATA_PATH: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/testdata");
 


hackeramitkumar:
@msardara
On Windows, there are named pipes, which are equivalent to Unix domain sockets. Tokio has an experimental implementation for named pipes, but tonic does not support named pipes yet.

So for now, I suggest we merge this part. Later, I will try to implement a custom tonic::NamedPipeListener to achieve behavior similar to tonic::UnixSocketListener.

data-plane/core/config/src/grpc/client.rs
```diff
@@ -546,56 +557,21 @@ impl ClientConfig {
...
         // Create the channel with or without retry based on lazy flag
-        let channel = if lazy {
-            let connection = self.create_connection(uri, http_connector).await?;
-            self.create_channel_from_connection(builder, connection, tls_config, true)
-                .await?
+        #[cfg(target_family = "unix")]
+        let is_unix = self.endpoint.starts_with("unix://");
+
+        #[cfg(not(target_family = "unix"))]
+        let is_unix = false;
```
Tehsmash: It feels a little bit weird to me to be using the `cfg(target_family= ...` macro here resulting in a configuration of `unix://` calling self.connect_tcp_channel (well actually it'll terminate in "parse_endpoint_uri").

I think it could be nicer to separate the parsing / validation, from the connecting for example, this function would become:

```
let uri = self.parse_endpoint_uri()?;

if uri.scheme == "unix" {
	self.connect_unix_channel(uri).await?
} else if uri.scheme == "http" || uri.scheme == "https" {
	self.connect_tcp_channel(uri).await?
} else {
    return Err(ConfigError::URISchemeUnsupported);
}
```

parse endpoint uri would become something like:

```
fn parse_endpoint_uri(&self) -> Result<Uri, ConfigError> {
	 // Special case unix scheme because it doesn't have an
     // authority in the uri and the Uri parser doesn't like this today
     // so we build our own uri with a fake localhost authority.
	 if self.endpoint.starts_with("unix://") {
		let Some(path) = endpoint.strip_prefix("unix://") else {
            return Err(ConfigError::UnixSocketMissingPath);
        };
		let uri = Uri::builder().scheme("unix").authority("localhost").path_and_query(path).build()?;
		Ok(uri)
	}
	Ok(Uri::from_str(&self.endpoint)?)
}
```

Then we can have:

```
#[cfg(not(target_family = "unix"))]
async fn connect_unix_channel(&self, uri: Uri, _lazy: bool) -> Result<Channel, ConfigError> {
    Err(ConfigError::UnixSocketUnsupported)
}

#[cfg(target_family = "unix")]
async fn connect_unix_channel(&self, uri: Uri, _lazy: bool) -> Result<Channel, ConfigError> {
    <the real implementation>
}
```

Once we've got a uri we can use that in the later functions instead of needing unix_placeholder_uri for the channel builder etx.

hackeramitkumar: Thanks, @Tehsmash , for the suggestions. I have implemented the changes. Please take another look.

data-plane/core/config/src/grpc/client.rs
```diff
@@ -546,56 +557,20 @@ impl ClientConfig {
...
 
-        // Create the channel with or without retry based on lazy flag
-        let channel = if lazy {
-            let connection = self.create_connection(uri, http_connector).await?;
-            self.create_channel_from_connection(builder, connection, tls_config, true)
-                .await?
+        let channel = if uri.scheme_str() == Some("unix") {
+            if cfg!(not(target_family = "unix")) {
+                return Err(ConfigError::UnixSocketUnsupported);
+            }
```
Tehsmash: No need for this check here as the non-unix implementation of "connect_unix_channel" will raise this, and then when we add other targets etc we can do similar.

data-plane/core/config/src/grpc/client.rs
```diff
@@ -610,8 +585,34 @@ impl ClientConfig {
...
-    /// Parses the endpoint string into a URI
+    /// Parses the endpoint string into a URI for TCP/HTTP, Unix domain socket endpoints.
     fn parse_endpoint_uri(&self) -> Result<Uri, ConfigError> {
+        // Special case for the unix scheme because it doesn't have an
+        // authority in the URI and the Uri parser doesn't like this today,
+        // so we build our own URI with a fake localhost authority.
+        if self.endpoint.starts_with("unix://") {
+            let Some(path) = self.endpoint.strip_prefix("unix://") else {
+                return Err(ConfigError::UnixSocketMissingPath);
+            };
```
Tehsmash: As we've already checked "starts_with(...)" we simply remove the prefix safely like:
```
let path = &self.endpoint[7..]
```

data-plane/core/config/src/grpc/client.rs
```diff
@@ -610,8 +585,34 @@ impl ClientConfig {
...
+
+            if path.trim_matches('/').is_empty() {
+                return Err(ConfigError::UnixSocketMissingPath);
+            }
+
+            let normalized = if path.starts_with('/') {
+                path.to_string()
+            } else {
+                format!("/{path}")
+            };
```
Tehsmash: I think we should allow for relative paths i.e. unix://relative.socket vs unix:///etc/foo/mysocket.socket

So we can just do:

```
if path.is_empty() {
	return Err(ConfigError::UnixSocketMissingPath);
}
```

any other issues should be caught when we try to connect to the socket I think.

data-plane/core/config/src/grpc/client.rs
```diff
@@ -732,6 +780,111 @@ impl ClientConfig {
...
     }
 
+    #[cfg(target_family = "unix")]
+    async fn connect_unix_channel(&self, uri: Uri, lazy: bool) -> Result<Channel, ConfigError> {
+        if !self.tls_setting.insecure {
+            // TLS handshakes are unnecessary over local UDS and currently unsupported
+            return Err(ConfigError::UnixSocketTlsUnsupported);
+        }
+
+        let socket_path = Arc::new(Self::parse_unix_socket_path(&self.endpoint)?);
```
Tehsmash: As we have the Uri that we've parsed earlier, there is no need for another function to parse the socket path. We can remove "parse_unix_socket_path" and just use `url.path()` which will also strip off any "query" part of the path.

data-plane/core/config/src/grpc/server.rs
```diff
@@ -318,6 +445,16 @@ impl ServerConfig {
             return Err(ConfigError::MissingEndpoint);
         }
 
+        #[cfg(target_family = "unix")]
+        if self.endpoint.starts_with("unix://") {
+            return self.to_unix_server_future(svc).await;
```
Tehsmash: I'm not a huge fan of the amount of duplicated logic in to_unix_server_future compared to here

`let incoming = UnixListenerStream::new(listener);`

and 

`let incoming = TcpIncoming::bind(addr)?;`

both implement the AsyncRead + AsyncWrite trait so I think it would be nicer if we could just create different "incoming" streams and then use them for the rest of the logic below.

(I know tls isn't supported for unix:// right now, but we can just validate that in the config layer I think)

hackeramitkumar: I think yeah we can do this. It will reduce duplicate code. 
I have slightly modified the implementation. Have a look.

data-plane/core/config/src/grpc/client.rs
hackeramitkumar: removed 'parse_unix_socket_path'.

data-plane/core/config/src/grpc/client.rs
hackeramitkumar: removed the normalization part from this function.

data-plane/core/config/src/grpc/client.rs
hackeramitkumar: removed the check.

data-plane/core/config/src/grpc/client.rs
```diff
@@ -6,9 +6,20 @@ use rustls_pki_types::ServerName;
...
+use std::path::PathBuf;
+#[cfg(target_family = "unix")]
+use std::sync::Arc;
 use std::time::Duration;
 use std::{collections::HashMap, str::FromStr};
+#[cfg(target_family = "unix")]
+use tokio::net::UnixStream;
 use tower::ServiceExt;
+#[cfg(target_family = "unix")]
+use tower::service_fn;
```
msardara:
```suggestion
use std::{collections::HashMap, error::Error as StdErrorTrait, str::FromStr, time::Duration};
use tower::ServiceExt;

#[cfg(target_family = "unix")]
use {
    hyper_util::rt::TokioIo,
    std::{path::PathBuf, sync::Arc},
    tokio::net::UnixStream,
    tower::service_fn,
};
```

data-plane/core/config/src/grpc/client.rs
```diff
@@ -732,6 +741,112 @@ impl ClientConfig {
...
+                            .await
+                            .map_err(Self::map_transport_error)
+                    }
+                },
+                |e: &ConfigError| match e {
+                    ConfigError::TransportError(err) => {
+                        tracing::warn!(error = %err.chain(), "Transport error encountered. Retrying...");
+                        true
+                    }
+                    #[cfg(target_family = "unix")]
```
msardara: Isn't the whole function already compiled out of non unix systems?

data-plane/core/config/tests/e2e.rs
```diff
@@ -98,6 +98,16 @@ mod tests {
...
+    #[cfg(unix)]
+    use tokio::net::UnixListener;
+    #[cfg(unix)]
+    use tokio::task::yield_now;
+    #[cfg(unix)]
+    use tokio::time::timeout;
+    #[cfg(unix)]
+    use tokio_stream::wrappers::UnixListenerStream;
+    #[cfg(unix)]
+    use tonic::transport::Server;
```
msardara:
```suggestion
#[cfg(unix)]
use {
    tokio::{net::UnixListener, task::yield_now, time::timeout},
    tokio_stream::wrappers::UnixListenerStream,
    tonic::transport::Server,
};
```

data-plane/core/config/src/grpc/client.rs
amitami2: Good catch. 
I think we can remove this.

amitami2:
Thanks, @msardara, for the suggestions. I’ve pushed the changes.

PR #1099, opened 2026-01-13
feat(bindings): configuration file support

msardara:
# Description

This PR integrates the configuration file support from the
`slim` crate into the `bindings` crate, enabling complete
configuration management, automatic service initialization,
and graceful shutdown matching the behavior of the main
SLIM application.

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/python/bindings/examples/Dockerfile
+++ b/data-plane/python/bindings/examples/Dockerfile
@@ -25,6 +25,7 @@ DEBIAN_FRONTEND=noninteractive \
         git \
         ca-certificates \
         gcc \
+        make \
         libc-dev
 
 curl -LsSf https://astral.sh/uv/install.sh | sh


data-plane/bindings/adapter/src/config.rs
```diff
@@ -0,0 +1,548 @@
...
+    GLOBAL_STATE.get_or_init(|| {
+        // Load configuration
+        let mut config = ConfigLoader::new(&config_path).expect("Failed to create config loader");
+
+        // Get configurations
+        let runtime_config = config.runtime().clone();
+        let tracing_conf = config.tracing().clone();
+        let service_config = match config.services() {
+            Ok(services) => {
+                if let Some(service) = services.values().next() {
```
micpapal, line 83: This loads only one service even if the config allows to have multiple ones, is this the expected behavior?

msardara: Yes for the moment we support only one global service. If we put a global array services[] then it becomes unclear what service should create an app when we call the NewBindingsAdapter wthout providing a service.

msardara: After discussion we decided to support a set of services as configured in the conf file and we use only the fist one as default service to create the App.

PR #1103, opened 2026-01-13
chore: refactor bindings github workflows

Tehsmash:
# Description

Refactors the bindings CI workflows to:
a) separate releases and PR testing
b) make as much common as possible

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/.github/workflows/release-bindings.yaml
+++ b/.github/workflows/release-bindings.yaml
@@ -5,10 +5,6 @@
 name: Release Bindings
 
 on:
-  pull_request:
-    paths:
-      - '.github/workflows/release-bindings.yaml'
-      - '.github/workflows/reusable-bindings-build-and-test.yaml'
   push:
     tags:
       - 'slim-bindings-libs-*'
@@ -33,7 +29,6 @@ jobs:
     name: Release Libraries
     runs-on: ubuntu-latest
     needs: build-and-test-bindings
-    if: startsWith(github.ref, 'refs/tags/slim-bindings-libs-')
     permissions:
       contents: write
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/release-bindings.yaml
```diff
@@ -7,9 +7,8 @@ name: Release Bindings
 on:
   pull_request:
     paths:
-      - 'data-plane/**'
       - '.github/workflows/release-bindings.yaml'
-
+      - '.github/workflows/reusable-bindings-build-and-test.yaml'
```
msardara: Do we want to keep updates to this file as trigger for the release workflow?

If we keep them I think we should also keep the `if: startsWith(github.ref, 'refs/tags/slim-bindings-libs-')` in the release-go-bindings workflow, otherwise we endup with [these tags](https://github.com/agntcy/slim-bindings-go/releases/tag/1116%2Fmerge) (coming from my PR which is rebased on this one) in the slim-bindings-go repo.

data-plane/bindings/go/Taskfile.yaml
```diff
@@ -120,6 +120,8 @@ tasks:
   generate:
     desc: Generate Go bindings from compiled library
     deps:
+      # TODO(sambetts): Consider making this optional through a wrapper
+      # task+status check to avoid building if we're using prebuilt libs
```
msardara, line 124: +1

.github/workflows/release-bindings.yaml
Tehsmash: What I was hoping is that I can run everything except the final merge so that we could test the workflow (creating a branch, pushing the changes etc) in PRs more easily, but I do see that its missing some clean up to remove the pre-release tag and branch. WDYT do we need this? or just test it with a real release?

msardara: I think it makes sense to run the workflow almost up to completion in PRs as well: we just need to skip the part where we merge the temporary branch to main in the go-bindings repo.

This would also allow to always test the installation/run using the tmp branch + the current lib artifact and then delete it (even though we cannot really test the artifact download script if we don't make a complete release).

PR #1106, opened 2026-01-14
refactor(bindings): allow multiple global services

msardara:
# Description

This PR extends the bindings layer to support multiple
global services instead of just a single global service.
This refactoring enables more flexible service configurations
while maintaining backward compatibility with existing code.

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/integrations/mcp-proxy/src/main.rs
+++ b/data-plane/integrations/mcp-proxy/src/main.rs
@@ -77,7 +77,7 @@ async fn main() {
     let _guard = config.tracing().setup_tracing_subscriber();
 
     let services = config.services().expect("error loading services");
-    let service = services.remove(&svc_id).expect("service not found");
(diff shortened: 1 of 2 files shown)


data-plane/bindings/rust/src/service.rs
```diff
@@ -356,7 +351,7 @@ pub fn create_service_with_config(
 /// across the application. All calls to this function return the same service instance.
 #[uniffi::export]
 pub fn get_global_service() -> Arc<Service> {
-    get_or_init_global_service()
+    crate::config::get_services()[0].clone()
```
micpapal: better to return the first in the vec or use directly the global service in the conf?

msardara: Good point, let's use the global service

PR #1107, opened 2026-01-14
fix(slim-bindings-setup): use ${GOPATH}/.cache/slim-bindings as default libs location

msardara:
# Description

Changes the cache directory location for SLIM bindings
libraries from the system cache directory to `$GOPATH/.cache/slim-bindings`.

### Motivation
Using `$GOPATH/.cache` is more consistent with Go ecosystem
conventions where Go-related artifacts are typically stored within
the GOPATH structure. This makes it easier for developers to:
 - Locate cached libraries alongside other Go dependencies
 - Clean up Go-related files in one location
 - Manage cache as part of their Go workspace

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/Cargo.lock
+++ b/data-plane/Cargo.lock
@@ -90,6 +90,8 @@ dependencies = [
  "agntcy-slim-testing",
  "agntcy-slim-tracing",
  "display-error-chain",
+ "futures",
+ "futures-timer",
  "parking_lot",
  "serde",
  "serde_json",
@@ -1470,6 +1472,12 @@ name = "futures-task"
 version = "0.3.31"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f90f7dce0722e95104fcb095585910c0977252f286e354b5e3bd38902cd99988"
+
+[[package]]
+name = "futures-timer"
+version = "3.0.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f288b0a4f20f9a56b5d1da57e2227c661b7b16168e2f72365f57b63326e29b24"
 
 [[package]]
 name = "futures-util"
--- a/data-plane/Cargo.toml
+++ b/data-plane/Cargo.toml
@@ -65,6 +65,7 @@ display-error-chain = { version = "0.2" }
 drain = { version = "0.2", features = ["retain"] }
 duration-string = { version = "0.5.3", features = ["serde"] }
 futures = "0.3.31"
+futures-timer = "3.0.3"
 h2 = "0.4.7"
 headers = "0.4.1"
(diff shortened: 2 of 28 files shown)


data-plane/bindings/go/release-files/cmd/slim-bindings-setup/main.go
```diff
@@ -68,22 +69,12 @@ func GetTarget(goos, arch string) string {
...
-		switch runtime.GOOS {
-		case "windows":
-			cacheHome = filepath.Join(home, "AppData", "Local")
-		default:
-			cacheHome = filepath.Join(home, ".cache")
-		}
+	gopath := build.Default.GOPATH
+	if gopath == "" {
+		return "", fmt.Errorf("failed to determine GOPATH")
 	}
```
Tehsmash, line 75: We should check the env first I think:

```
cacheHome := os.Getenv("GOPATH")
if cacheHome == "" {
	gopath := build.Default.GOPATH
}
```

data-plane/bindings/go/release-files/cmd/slim-bindings-setup/main.go
```diff
@@ -68,22 +69,12 @@ func GetTarget(goos, arch string) string {
...
 
-	return filepath.Join(cacheHome, cacheDirName), nil
+	return filepath.Join(gopath, ".cache", cacheDirName), nil
```
Tehsmash: Do we need another layer with ".cache", or can we just do `$GOPATH/.slim-bindings` ? or I was thinking we could even do `.cgo-cache` and maybe standardise this technique?

data-plane/bindings/go/release-files/cmd/slim-bindings-setup/main.go
msardara: I think we can skip it as the build package is already checking it: https://github.com/golang/go/blob/master/src/go/build/build.go#L341

data-plane/bindings/go/release-files/cmd/slim-bindings-setup/main.go
msardara: Yes let's go with `.cgo-cache`, I like the idea of standardising it 🔝

data-plane/bindings/go/release-files/cmd/slim-bindings-setup/main.go
Tehsmash: Oh cool! I didn't realise that build.Default was built dynamically like that. Perfect!

PR #1114, opened 2026-01-19
build: (x)build the slimctl tool with zig

lpuskas:
Zig Cross-Compilation Support

### DIFF
--- a/.github/workflows/reusable-go-build-and-test.yaml
+++ b/.github/workflows/reusable-go-build-and-test.yaml
@@ -276,9 +276,6 @@ jobs:
           - os: windows
             arch: amd64
             runner: ubuntu-latest
-          - os: windows
-            arch: arm64
-            runner: ubuntu-24.04-arm
 
     steps:
       - name: Checkout Repo


control-plane/slimctl/internal/manager/slim.go
```diff
@@ -67,10 +64,7 @@ func (s *manager) Start(_ context.Context) error {
 	fmt.Printf("✅ Server app created (ID: %d)\n", app.Id())
```
msardara: For the use case of running a local node we can drop the app creation and just use the server in the global service.

The only thing I'd add is the config file support. It is already available in the bindings, we can probably add a flag in slimctl in a next PR

msardara (approved):

control-plane/Taskfile.yaml
```diff
@@ -146,74 +146,106 @@ tasks:
...
+          unset CXX
+          unset CGO_CFLAGS
+          unset CGO_LDFLAGS
         else
           # Cross-compilation - setup Zig compiler
-          echo "🔨 Cross-compiling slimctl from $CURRENT_OS/$CURRENT_ARCH to {{.TARGETOS}}/{{.TARGETARCH}}"
+          echo "🔨 Cross-compiling slimctl from {{.CURRENT_OS}}/{{.CURRENT_ARCH}} to {{.TARGETOS}}/{{.TARGETARCH}}"
           BUILD_MODE="cross"
-
+
```
msardara: Let's remove these trailing whitespaces

control-plane/slimctl/internal/manager/slim.go
lpuskas: created a task to track the config support #1129 
removed the unnecessary code

Issue #1118, opened 2026-01-20
Add flags to disable send/receive messages in a SLIM app


micpapal:
The are some applications that just need to send data (e.g. otel exporter) or receive data (e.g. otel receiver).
SLIM always setup bidirectional channels where each app can send and receive messages. However this can be problematic, especially for applications that only wants to send data. In fact even if not needed, the application should listen for incoming messages as they will be received because the app is part of a group and messages are broadcasted on the group.

Add config flag to shutdown the receive or send capability or the application so that the application will not need to care about unwanted messages.

PR #1121, opened 2026-01-21
feat(session): Add direction to slim app to control message flow

micpapal:
# Description

Adds direction to applications to disable sending or receiving messages while maintaining protocol compliance.

#### Core Session Layer
- **SessionReceiver**: Added `shutdown_receive` flag
  - When true, ACKs all incoming messages but does not deliver them to the application

  
- **SessionSender**: Added `shutdown_send` flag  
  - When true, accepts messages from application but does not transmit to network
  - Immediately signals success via ack notifiers to prevent application errors

#### Service Layer & Bindings
- **Service API**: New methods for creating apps with direction
  - `create_app_with_direction()` - Blocking version
  - `create_app_with_direction_async()` - Async version

ref: #1118 

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


data-plane/bindings/rust/src/service.rs
```diff
@@ -304,6 +304,45 @@ impl Service {
...
+    /// # Returns
+    /// * `Ok(Arc<App>)` - Successfully created adapter
+    /// * `Err(SlimError)` - If adapter creation fails
+    pub async fn create_app_with_shutdown_flags_async(
+        &self,
+        base_name: Arc<Name>,
+        identity_provider_config: IdentityProviderConfig,
+        identity_verifier_config: IdentityVerifierConfig,
+        shutdown_send: bool,
+        shutdown_receive: bool,
```
msardara: I was thinking we can use an enum for the main interface:

```
enum Direction {
  Send,
  Recv,
  Bidirectional,
  None
}
```

What do you think?

data-plane/core/service/src/app.rs
```diff
@@ -78,6 +79,8 @@ where
         tx_slim: SlimChannelSender,
         tx_app: mpsc::Sender<Result<Notification, SessionError>>,
         storage_path: std::path::PathBuf,
+        shutdown_send: bool,
+        shutdown_receive: bool,
```
msardara: Let's create also 2 constructors for App so that we don't have to modify all the tests.

data-plane/core/session/src/session_receiver.rs
```diff
@@ -106,6 +111,12 @@ impl SessionReceiver {
             None
         };
 
+        if shutdown_receive {
+            tracing::info!(
```
msardara: This looks like a debug log

data-plane/core/session/src/session_sender.rs
```diff
@@ -106,6 +111,12 @@ impl SessionSender {
             None
         };
 
+        if shutdown_send {
+            tracing::info!(
```
msardara: This looks like a debug message

data-plane/core/session/src/session_sender.rs
```diff
@@ -106,6 +111,12 @@ impl SessionSender {
...
+                "creating session sender with shutdown_send=true, no messages will be sent to the network"
+            );
+        }
```
msardara: I was thinking we can just debug log here, independently of the fact that shutdown_send (and shutdown_receive) is set:

```suggestion
        tracing::debug!(
           %shutdown_send, "creating session sender"
        );
```

data-plane/core/session/src/session_sender.rs
```diff
@@ -188,6 +200,16 @@ impl SessionSender {
         mut message: Message,
         ack_tx: Option<oneshot::Sender<Result<(), SessionError>>>,
     ) -> Result<(), SessionError> {
+        // if shutdown_send is true we drop all messages but we signal success
+        // to the application using the ack channel if provided
+        if self.shutdown_send {
+            debug!("sender is shutdown, drop message");
```
msardara: Let's also log the message id:
```suggestion
            debug!(%message.id(), "sender is shutdown, drop message");
```

data-plane/core/session/src/session_sender.rs
```diff
@@ -188,6 +200,16 @@ impl SessionSender {
...
+            if let Some(tx) = ack_tx {
+                let _ = tx.send(Ok(()));
```
msardara: Is it correct not to signal an error here? If the app creates a session with the send disabled and it tries to send something, we should probably signal it.

data-plane/core/session/src/session_receiver.rs
micpapal: I wanted to have it printed so that when the app starts the user will see that the flag is set. I can move to debug of course

data-plane/core/session/src/session_sender.rs
micpapal: same here

data-plane/core/session/src/session_sender.rs
micpapal: I was not sure what to do in this case. I started with an error and then I went back to this to avoid unexpected crashes on the app. But is it also true that this should not happen if the app is configured to do not send messages. Probably the error make more sense. I will go back to that.

data-plane/bindings/rust/src/service.rs
micpapal: This is a good idea, also because the two bool param one after the other are a bit tricky use, you can easily swap the two.

Issue #1138, opened 2026-01-26
add reference count to subscription table


micpapal:
Add a reference count to each subscription. This will prevent to remove routes that are used by other applications.
Also: remove the route_cache in the session layer

PR #1143, opened 2026-01-27
feat: add reference count to subscription table

micpapal:
# Description

Previously, when multiple consumers within a session subscribed to the same name, each subscription would send a separate subscribe message to the data plane. Similarly, each unsubscribe would send an unsubscribe message. This lead to issues where a subscription was removed from the data plane while other consumers still needed it.

This PR implements reference counting in the subscription table to prevent this problem. It also removes the session route cache in the session layer as the ref count is now directly implemented in the subscription table.

Ref: #1138 

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/core/datapath/src/tables/subscription_table.rs
+++ b/data-plane/core/datapath/src/tables/subscription_table.rs
@@ -57,9 +57,7 @@ struct SubscriptionRefs {
 
 impl SubscriptionRefs {
     fn new(conn: u64) -> Self {
-        let refs = HashMap::from([
-            (conn, 1),
-        ]);
+        let refs = HashMap::from([(conn, 1)]);
         SubscriptionRefs { refs }
     }
 


data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -49,6 +49,65 @@ impl ConnId {
...
+#[derive(Debug, Default)]
+struct SubscriptionRefs {
+    // map from connection id to reference counter
+    refs: HashMap<u64, usize>,
+}
+
+impl SubscriptionRefs {
+    fn new(conn: u64) -> Self {
+        let mut refs = HashMap::new();
+        refs.insert(conn, 1);
```
```suggestion
        let refs = HashMap::from([
            (conn, 1),
        ]);
```

PR #1228, opened 2026-02-13
fix: cache usage

gomba999:
# Description

Fix typo in cache usage, disable caching cargo/bin since the swatinem cache used by the rust toolchain has a bug, add that as a directory instead. Disable cache for release builds.

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/.github/workflows/reusable-bindings-build-and-test.yaml
+++ b/.github/workflows/reusable-bindings-build-and-test.yaml
@@ -15,7 +15,7 @@ on:
         description: Git tag name for version extraction (e.g., slim-bindings-1.2.3)
         required: false
         type: string
-        default: ''
+        default: ""
       rust-workspace:
         description: Rust workspace directory for caching
         required: false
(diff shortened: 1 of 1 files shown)


.github/workflows/reusable-rust-build-and-test.yaml
```diff
@@ -129,6 +137,10 @@ jobs:
         uses: ./.github/actions/setup-rust
         with:
           workspace: ${{ inputs.working-directory }}
+          cache-directories: |
+            ~/.cargo/bin
+            ~/.cargo/.crates.toml
+            ~/.cargo/.crates2.json
```
msardara: I'd tend to leave this as default instead of repearing it around, what do you think?

gomba999: My thinking was that without this the install of components with binstall would have to be forced every time so this let's us spare a little more time on the builds.

msardara: 100% agree, I am suggesting to put them as default in the action directly

gomba999: ah sorry I see now, I thought you meant the default as it is defined by the toolchain itself, I can set these to be the default values in the setup-rust workflow

PR #1247, opened 2026-02-18
ci: use local bindings in integration tests

muscariello:
ci: CI/tooling updates (setup actions, release/test setup workflows, golangci).

### DIFF
--- a/.github/workflows/control-plane.yaml
+++ b/.github/workflows/control-plane.yaml
@@ -26,8 +26,6 @@ jobs:
     with:
       go-version: 1.25.5
       working-directory: ./control-plane
-      use-local-bindings: true
-      cgo-enabled: '1'
 
   control-plane-docker-build:
     name: Build control plane docker image
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -37,14 +37,6 @@ jobs:
       - name: Generate data plane coverage report (Rust + Python)
         run: task data-plane:coverage-full
 
-      - name: Setup SLIM Go bindings
-        working-directory: ${{ github.workspace }}
-        run: |
-          CGO_ENABLED=1 task -d control-plane control-plane:go-bindings:setup \
-            TARGETOS=$(go env GOOS) \
-            TARGETARCH=$(go env GOARCH) \
-            USE_LOCAL_BINDINGS=true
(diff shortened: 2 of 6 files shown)


.github/actions/setup-go/action.yaml
```diff
@@ -36,7 +36,7 @@ runs:
     - name: Setup Taskfile
       uses: ./.github/actions/setup-task
       with:
-        version: latest
```
msardara, line 39: The latest version of task should be 3.48.0 IIRC

.github/actions/setup-rust/action.yaml
```diff
@@ -26,12 +26,12 @@ runs:
         cache-workspaces: ${{ inputs.workspace }}
         cache-on-failure: false
         cache-shared-key: ${{runner.os}}-rust-${{hashFiles('**/Cargo.lock')}}
-        cache-all-crates: trus
+        cache-all-crates: true
```
msardara, line 29: This is what happens when types are not enforced 🥲

.github/actions/setup-python/action.yaml
```diff
@@ -40,7 +40,7 @@ runs:
     - name: Setup Taskfile
       uses: ./.github/actions/setup-task
       with:
-        version: latest
+        version: v3.39.2
```
msardara: We should set a default in the setup-task action instead of repeating the version around.

data-plane/bindings/go/slim_bindings/nocgo_stub.go
msardara, line 1: We can remove this file, otherwise it will be published as part of the go module.

.github/actions/setup-go/action.yaml
muscariello: which version do you want.

muscariello: No latest. One number.

msardara: It's in the comment

> The latest version of task should be 3.48.0

muscariello: should or is. do I have to verify the number or are you sure?

msardara: I am sure, v3.48.0

PR #1248, opened 2026-02-18
fix(slimctl): improve startup error reporting

muscariello:
## Summary
- improve slimctl startup error reporting and formatting
- add coverage for formatted start errors
- adjust bindings config used by slimctl

## Testing
- task -d control-plane control-plane:test


data-plane/bindings/go/slim_bindings/nocgo_stub.go
msardara, line 1: This file should be removed not to be published with the rest of the golang module

PR #1255, opened 2026-02-23
feat(data-plane): port slimctl to Rust

muscariello:
## Summary
- add a new Rust slimctl crate under `data-plane/slimctl`
- implement node/controller/config/version/slim-start command paths with Go CLI parity
- generate and wire control-plane/controller protobuf clients for Rust slimctl
- integrate Rust slimctl into workspace/taskfiles and stage binary to `.dist/bin/slimctl`
- update integration suite wiring to allow/consume Rust slimctl path

## Validation
- `task -d data-plane data-plane:fmt`
- `task -d data-plane data-plane:sort`
- `task -d data-plane data-plane:lint`
- `task -d tests tests:integration`

### DIFF
--- a/data-plane/bindings/python/examples/common.py
+++ b/data-plane/bindings/python/examples/common.py
@@ -58,7 +58,6 @@ def format_message_print(message1: str, message2: str = "") -> None:
     print(format_message(message1, message2))
 
 
-
 def jwt_identity(
     jwt_path: str,
     spire_bundle_path: str,
--- a/data-plane/deny.toml
+++ b/data-plane/deny.toml
@@ -31,7 +31,9 @@ allow = [
 # Ignore local workspace license values for unpublished crates.
 private = { ignore = true }
 confidence-threshold = 0.8
-exceptions = []
+exceptions = [
+    { name = "duration-string", version = "0.5.3", allow = ["Apache-2.0", "MIT"] },
+]
 
 [bans]
 # TODO: Remove duplicate dependencies.


muscariello:
Added coverage-focused unit tests for the new Rust slimctl crate.

muscariello:
Added coverage-focused unit tests for the new Rust slimctl crate.

### What was added
- `config.rs` tests for precedence logic (defaults, file→CLI override, env→CLI/file override)
- `node_route.rs` tests for route/endpoint parsing (valid + invalid cases)

### Validation run
- `cargo test -p agntcy-slimctl` → 7/7 tests passing
- `cargo llvm-cov --package agntcy-slimctl --all-targets` (with explicit LLVM tool paths)

### Current slimctl line coverage (from lcov aggregation)
- `data-plane/slimctl/src/config.rs`: 189/248
- `data-plane/slimctl/src/node_route.rs`: 60/281
- total `data-plane/slimctl/src/*`: **249/1139 (21.86%)**

Not sure why codecov has a different number...

data-plane/bindings/python/examples/common.py
```diff
@@ -218,7 +226,7 @@ async def create_local_app(config: BaseConfig) -> tuple[slim_bindings.App, int]:
     service = setup_service()
 
     # Convert local identifier to a strongly typed Name.
-    local_name = slim_bindings.Name.from_string(config.local)
+    local_name = parse_name(config.local)
```
micpapal: here I think is better to keep using the binding implementation instead to create a new function. And same for all the the other calls to parse_name

data-plane/Taskfile.yaml
```diff
@@ -60,6 +60,23 @@ tasks:
           BIN: "slim"
           BIN_ARGS: "--config ./config/telemetry/server-config.yaml"
 
+  data-plane:run:slimctl:
+    desc: "Run the Rust slimctl binary"
+    cmds:
+      - task: data-plane:run
+        vars:
+          BIN: "slimctl"
+          BIN_ARGS: "{{.BIN_ARGS}}"
```
micpapal, line 70: this is not working to me. I need to specify the package name adding 
```
   ARGS: "--package agntcy-slimctl"
```
in vars

data-plane/Taskfile.yaml
```diff
@@ -60,6 +60,23 @@ tasks:
           BIN: "slim"
           BIN_ARGS: "--config ./config/telemetry/server-config.yaml"
 
+  data-plane:run:slimctl:
```
micpapal: here it is better to use data-plane:slimctl:run instead of data-plane:run:slimctl so that slimctl commands will remain close to each other in the in the task list

data-plane/slimctl/src/cli.rs
```diff
@@ -0,0 +1,403 @@
...
+#[derive(Debug, Args)]
+pub struct SlimCommand {
+    #[command(subcommand)]
+    pub command: SlimSubcommand,
+}
+
+#[derive(Debug, Subcommand)]
+pub enum SlimSubcommand {
+    Start {
+        #[arg(short = 'c', long = "config", default_value = "")]
```
micpapal: here the default value is set to "" but actually in slim_start.rs is set to 127.0.0.1:8080. I think we should move to port 46357 which is the one we use everywhere.

data-plane/slimctl/src/cli.rs
```diff
@@ -0,0 +1,403 @@
...
+        config: String,
+        #[arg(long = "endpoint", default_value = "")]
```
micpapal: here the default value can be remote as the default config is generated in slim_start.rs

data-plane/slimctl/src/main.rs
```diff
@@ -0,0 +1,408 @@
...
+                }
+            },
+            NodeSubcommand::Connection(connection) => match connection.command {
+                NodeConnectionSubcommand::List => {
+                    node_connection::connection_list(&effective_opts.server)?
+                }
+            },
+        },
+        Commands::Slim(slim_cmd) => match slim_cmd.command {
+            SlimSubcommand::Start { config, endpoint } => {
```
micpapal, line 185: if I run
```
slimctl slim start --help
```

as output I get
```
slimctl slim start --help
Usage: slimctl slim start [OPTIONS]

Options:
  -b, --basic_auth_creds <BASIC_AUTH_CREDS>  
  -c, --config <CONFIG>                      [default: ]
      --endpoint <ENDPOINT>                  [default: ]
  -s, --server <SERVER>                      
      --timeout <TIMEOUT>                    
      --tls.insecure <TLS_INSECURE>          [possible values: true, false]
      --tls.ca_file <TLS_CA_FILE>            
      --tls.cert_file <TLS_CERT_FILE>        
      --tls.key_file <TLS_KEY_FILE>          
  -h, --help                                 Print help
```
but the only usable options are --config and --endpoint. I think we should not dispaly the options that are not available for the command. This needs to be checked also for the others sub-commands.

data-plane/slimctl/src/config.rs
```diff
@@ -0,0 +1,258 @@
...
+        effective.tls_ca_file = value.clone();
+    }
+    if let Some(value) = &cli.tls_cert_file {
+        effective.tls_cert_file = value.clone();
+    }
+    if let Some(value) = &cli.tls_key_file {
+        effective.tls_key_file = value.clone();
+    }
+
+    if let Ok(value) = env::var("SLIMCTL_BASIC_AUTH_CREDS") {
```
micpapal, line 148: here the env:var overwrite the settings of the command flags. Usually is the opposite.

data-plane/slimctl/src/config.rs
```diff
@@ -0,0 +1,258 @@
...
+            tls_insecure: true,
+            tls_ca_file: String::new(),
+            tls_cert_file: String::new(),
+            tls_key_file: String::new(),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Default)]
+pub struct CliCommonOverrides {
```
micpapal: is there a reason here to do not reuse CommonOpts instead of declaring CliCommonOverrides? The structs are the same, the only difference is that CommonOpts derives Serialize, Deserialize

data-plane/slimctl/src/controller_cmds.rs
```diff
@@ -0,0 +1,785 @@
...
+        let mut client = connect_client(server).await?;
+        let request = build_delete_route_request(node_id, route, via_keyword, endpoint_or_node)?;
+
+        let response = client
+            .delete_route(request)
+            .await
+            .context("failed to delete route")?
+            .into_inner();
+
+        println!("ACK received success={}", response.success);
```
micpapal: here the print should be consistent with 
```
println!("Route created successfully with ID: {}", response.route_id);
```

data-plane/slimctl/src/controller_cmds.rs
```diff
@@ -0,0 +1,785 @@
...
+
+    with_runtime(async move {
+        let mut client = connect_client(server).await?;
+        let response = client
+            .create_channel(request)
+            .await
+            .context("failed to create channel")?
+            .into_inner();
+
+        println!("Received response: {}", response.channel_name);
```
micpapal: Also here should be something like "Channel created successfully with Name: {}"

data-plane/Taskfile.yaml
muscariello: we need to remove the old slimctl which has that name taken.

data-plane/slimctl/src/main.rs
muscariello: that was a leftover. Removed.

data-plane/Taskfile.yaml
muscariello: fixed

data-plane/slimctl/src/cli.rs
muscariello: I created defaults.rs to get a single point of truth. By using defaults clap will print default values as examples in the help.

data-plane/slimctl/src/cli.rs
muscariello: this is solved with the new defaults.rs

Issue #1311, opened 2026-03-02
Configuration model and data-plane config changes


msardara:
## Summary

Extend the data-plane configuration layer to support explicit transport selection (`grpc` or `websocket`) for both server listeners and client routes.

## Background

Currently the config layer only supports gRPC. WebSocket transport requires new fields for endpoint parsing, TLS reuse, auth at handshake time, and an optional query-parameter auth mode for browser clients.

## Tasks

- [ ] Add `transport` enum/schema with values `grpc` (default) and `websocket` to client and server config
- [ ] Add WebSocket endpoint parsing (ws:// and wss://) to client config
- [ ] Add WebSocket listener binding and accept-path to server config
- [ ] Add optional `websocket_auth_query_param` field to client config (JWT sent as query param instead of Authorization header)
- [ ] Reuse existing TLS schema for WebSocket (no new TLS fields required)
- [ ] Validate that existing gRPC configs continue to work unchanged when `transport` is omitted

### Server config with WebSocket

```yaml
services:
  slim/0:
    dataplane:
      servers:
        - endpoint: "0.0.0.0:46357"
          transport: websocket
          tls:
            insecure: true
```

### Client route config (for `slimctl node route add`)

```json
{
  "endpoint": "ws://127.0.0.1:46357",
  "transport": "websocket",
  "websocket_auth_query_param": "token",
  "auth": {
    "type": "static_jwt",
    "file": "/tmp/slim-browser.jwt"
  },
  "tls": {
    "insecure": true
  }
}
```

## Acceptance Criteria

- `transport: websocket` is accepted in server and client config without errors
- `transport: grpc` (or omitted) continues to behave as today
- `websocket_auth_query_param` field is parsed and available for handshake construction
- TLS fields are shared/reused — no duplication of schema

PR #1333, opened 2026-03-05
feat(websocket): update the config module while making sure backward compatibility

hackeramitkumar:
## Summary
This PR adds **websocket-aware transport configuration** across core config, schemas, and Rust bindings, while keeping default behavior backward compatible.

## Motivation
We needed a clear config model to choose transport (`grpc` vs `websocket`) and support websocket-specific client options without breaking existing deployments.

## What Changed
- Added shared transport enum in core config:
  - `slim_config::transport::TransportProtocol` (`grpc` default, `websocket` optional).
- Added `transport` to both client and server config structs:
  - Defaults to `grpc` when omitted.
- Added websocket client option:
  - `websocket_auth_query_param` for auth token/query-key style websocket endpoints.
- Updated JSON schemas:
  - `client-config.schema.json` and `server-config.schema.json` now include `transport` with default `grpc`.
- Added/updated validation and explicit errors:
  - Websocket endpoint scheme validation (`ws://` or `wss://`) for websocket mode.
  - Explicit unsupported errors when trying to use websocket transport through current gRPC builders.
- Rust UniFFI bindings cleanup:
  - Unified duplicated client/server transport enums into one shared bindings enum.
  - Preserved compatibility aliases for re-exports.

## Backward Compatibility
- Existing configs that do not set `transport` continue to work unchanged (default = `grpc`).
- `endpoint` remains the required field.
- No behavior change for existing gRPC-only users.

## Example
```yaml
# client
endpoint: "wss://example.com/ws"
transport: websocket
websocket_auth_query_param: token
```

```yaml
# server
endpoint: "0.0.0.0:50051"
transport: grpc
```

## Testing
- Rust bindings compile/tests pass after enum unification.
- SlimRPC doctests updated and passing.
- Schema generation/usage updated for new transport fields.

## Notes
This PR introduces config + validation groundwork for websocket transport selection.  
Where websocket runtime path is not yet supported in current gRPC builders, the system now fails with explicit, actionable errors instead of ambiguous failures.

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

Fixes: #1311


data-plane/bindings/rust/src/slimrpc.rs
```diff
@@ -25,21 +25,21 @@
 //! ## Client Example
 //!
 //! ```no_run
-//! # use slim_bindings::{Channel, Context, Status, Encoder, Decoder};
+//! # use slim_bindings::{Channel, Context, RpcError, Encoder, Decoder};
```
msardara, line 28: Nice, thanks for fixing these docs! 🫶

data-plane/config/websocket/client-config.json
msardara, line 1: Do we need the json example? Here we only have examples for the config file, which is yaml only 🤔

These json files can be used by the control plane to send a connection config, but I'd not put them here, as this folder only contains data plane startup configurations.

data-plane/core/config/src/transport.rs
msardara, line 1: We are not using the `mod.rs` notation anymore. Let's just create a file called `transport.rs`

msardara (changes requested):
LGTM, with minor nits ✅ Thanks a lot Amit!

I think next step is to return a trait or an enum from the to_channel() so that we can reuse it to get either a websocket "channel" or a grpc "channel".

data-plane/core/config/src/transport.rs
hackeramitkumar: done

data-plane/config/websocket/client-config.json
hackeramitkumar: Understood. I will remove these from here and add them to the control plane later when needed.

Issue #1325, opened 2026-03-03
Refactor server/client dispatching to use service and method metadata


msardara:
## Description
The current implementation relies on the SLIM subscription itself to differentiate RPC methods. This is a blocker for group sessions where we want to invite an entity to a group rather than its individual methods. We need to move method resolution into the RPC payload metadata.

## Key Changes:
 - Metadata Update: Ensure the client sends the service_name and method_name as part of the RPC metadata/header.
 - Server-Side Dispatcher: Refactor the server to demultiplex incoming requests by looking at the service_name and method_name in the payload rather than the subscription type.
 - Entity-Centric Invites: Update the invitation logic to support adding an entity to a group session without requiring specific method-level subscriptions.

## Acceptance Criteria:
[ ] Server can resolve and route calls to the correct handler based solely on payload metadata.
[ ] Successful RPC execution within a group session where members are invited as entities.
[ ] Documentation updated to reflect the new dispatching requirements for service implementers.

PR #1334, opened 2026-03-06
feat(slimrpc): session reuse and method demultiplexing

### LINKED ISSUE #1324: Implement Request Multiplexing and Session Reuse
## Description
Currently, slimrpc creates a new session for every RPC call. This prevents efficient scaling for group transport and adds unnecessary overhead. This task involves refactoring the client-side transport to allow a single session to persist across multiple calls.

## Key Changes:
 - Introduce RPC IDs: Add a unique **rpc_id** to the RPC metadata
 - Persistent Transport: Modify the client to hook into an existing session instead of initiating a teardown after a single response.
 - Response Mapping: Implement a registry (e.g., a Map or Dictionary) on the client to track pending request_ids and route incoming responses to the correct asynchronous handler/callback.

## Acceptance Criteria:
[ ] Multiple concurrent RPC calls can be initiated over a single transport session.
[ ] Each response is correctly routed to its specific caller via the request_id.
[ ] No regression in existing p2p single-session logic.

msardara:
# Description

Fixes #1324 and #1325

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/.github/actions/setup-rust/action.yaml
+++ b/.github/actions/setup-rust/action.yaml
@@ -2,7 +2,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
 ---
-
 name: Setup Environment
 description: setup environment to build/test/lint rust applications
 inputs:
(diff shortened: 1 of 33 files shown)


data-plane/bindings/rust/src/slimrpc/channel.rs
```diff
@@ -114,76 +191,273 @@ impl Channel {
...
+                .publish(request_bytes, Some("msg".to_string()), Some(msg_meta))
+                .await?;
+            handles.push(handle);
+        }
+
+        // EOS marker — always carries status OK + rpc-id.
+        // When the stream was empty, also include service/method/deadline so the
+        // server can identify the target handler from this sole message.
+        let routing = if first {
+            Some((ctx, service_name, method_name))
```
micpapal, line 402: This happens if the stream request sends an empty stream or in other words no message is sent except for the EOS. Is this a valid scenario? should we trigger an error here instead or say to the server side how to handle the request?

micpapal: This actually makes sense if you open for example a bidirectional stream but you don't send messages.

PR #1335, opened 2026-03-06
feat: multicast RPC support

### LINKED ISSUE #1332: Group RPC design proposal
## Introduction

Standard SlimRPC is a point-to-point protocol: a client opens a connection to a single server and exchanges messages with it. Group RPC is the natural extension of this model, where a client opens a single connection to a group of servers simultaneously and interacts with all of them at once. From the client's perspective the API is almost identical — the same four interaction patterns (unary-unary, unary-stream, stream-unary, stream-stream) are available — but every request is broadcast to all group members and every member sends back its own independent response. The client receives a stream of `MulticastItem` values, each tagged with the source member's identity, allowing it to correlate responses with their origin. Under the hood this is built on top of the SLIM GROUP (Multicast) session type.

---

## Architecture overview

```
                        ┌──────────────────────────────────────┐
                        │              CLIENT                  │
                        │                                      │
                        │  Channel (is_group=true)             │
                        │  ┌────────────────────────────────┐  │
                        │  │ session (Multicast)            │  │
                        │  │ ResponseDispatcher             │  │
                        │  │  rpc_id → mpsc::Rx             │  │
                        │  └────────────────────────────────┘  │
                        └──────────────┬───────────────────────┘
                                       │ GROUP session
                          ┌────────────┼────────────┐
                          ▼            ▼            ▼
               ┌──────────────┐  ┌──────────────┐  ...
               │   Member 0   │  │   Member 1   │
               │   Server     │  │   Server     │
               │  run_session │  │  run_session │
               │  _demux      │  │  _demux      │
               └──────────────┘  └──────────────┘
```

A `Channel` with `is_group=true` holds a single SLIM Multicast session. The GROUP session name is derived from the client app name (same org/namespace, random UUID as the third component) so it lives in the client's namespace.

On the first multicast call the channel auto-invites all `pending_members`. Every subsequent call reuses the same live session; the session is transparently recreated if it dies.

---

## Session naming

The GROUP session name is derived at `Channel` construction time:

```
client name:   org / namespace / my-client
group name:    org / namespace / <uuid-v4>
```

This ensures the group address is scoped to the same org/namespace as the client and does not collide with any service address.

---

## Message metadata keys

| Key             | Set by    | Meaning                                         |
|-----------------|-----------|-------------------------------------------------|
| `service`       | client    | Service name of the call; absent on responses   |
| `method`        | client    | Method name of the call; absent on responses    |
| `rpc-id`        | client    | UUID identifying the call; echoed on responses  |
| `slimrpc-code`  | server    | Integer `RpcCode`; present on every response    |
| `slimrpc-timeout` | client  | Deadline as Unix timestamp (`f64` seconds since epoch) |

A message is an **EOS marker** when `slimrpc-code = 0` (Ok) **and** the payload is empty. A message is a **request** when `service` is present. A message is a **response** when `slimrpc-code` is present and `service` is absent.

---

### Session and dispatcher

```mermaid
flowchart TD
    A[multicast_unary / multicast_stream_stream / ...] --> B[get_or_create_session]
    B -->|session alive| C[reuse ChannelSession]
    B -->|session dead| D[create_raw_session_typed Multicast]
    D --> E[invite pending_members once]
    E --> F[spawn response_dispatcher_task]
    F --> C
    C --> G[generate rpc_id]
    G --> H[ResponseDispatcher.register rpc_id → mpsc::Rx]
    H --> I[send request / request stream]
    I --> J[read from mpsc::Rx loop]
```

### ResponseDispatcher

The `ResponseDispatcher` is a shared map `rpc_id → mpsc::UnboundedSender`. A single background task (`response_dispatcher_task`) reads every incoming message from the GROUP session and routes it by `rpc-id` to the registered sender.

```mermaid
sequenceDiagram
    participant BG as response_dispatcher_task
    participant D  as ResponseDispatcher
    participant R0 as RPC call 0 (rpc_id=A)
    participant R1 as RPC call 1 (rpc_id=B)

    R0->>D: register("A") → rx_A
    R1->>D: register("B") → rx_B

    BG->>BG: session_rx.get_message()
    BG->>D: dispatch(msg{rpc_id="A"})
    D-->>R0: rx_A.send(msg)

    BG->>BG: session_rx.get_message()
    BG->>D: dispatch(msg{rpc_id="B"})
    D-->>R1: rx_B.send(msg)

    Note over R0: DispatcherGuard drops → unregister("A")
```

Multiple concurrent calls over the same GROUP session are fully multiplexed via their unique `rpc_id`.

---

### Per-session demultiplexer

Each incoming GROUP session on the server runs `run_session_demux` in a dedicated task.

```mermaid
flowchart TD
    A[session_rx.get_message] --> B{rpc_id in pending_streams?}
    B -->|yes stream continuation| C[forward msg to handler via mpsc]
    C --> D{msg is EOS?}
    D -->|yes| E[remove from pending_streams]
    D -->|no| A
    B -->|no| F{service key present?}
    F -->|no — response from other member| A
    F -->|yes| G[lookup handler in registry]
    G -->|not found| H[send_error_for_rpc]
    H --> A
    G -->|found| I[spawn_handler_task]
    I -->|stream-input handler| J[insert rpc_id into pending_streams]
    I --> A
```

**Key filtering rule**: a message with no `service` key is not a new inbound request. Because responses are sent unicast directly to the client, other members never receive them. This filter therefore handles only other unexpected non-request messages that may arrive on the GROUP session. A future extension will introduce a mechanism to explicitly deliver peer responses to members for context propagation, allowing members to observe each other's responses within the same GROUP session.

**Stream-input handlers** (`StreamUnary`, `StreamStream`) receive their first message directly when spawned. Subsequent client messages arrive through `pending_streams`: the demux forwards them into the handler's mpsc channel until the client sends an EOS.

### Handler response — unicast reply

Handlers reply **unicast** to `first_message.source` (the client's app name) via `SessionTx::publish(target, ...)`. This means each member's response travels directly to the client without being re-broadcast to other members.

```mermaid
sequenceDiagram
    participant Client
    participant SLIM
    participant M0 as Member 0 server
    participant M1 as Member 1 server

    Client->>SLIM: GROUP broadcast  request{service, method, rpc_id=X}
    SLIM-->>M0: deliver request
    SLIM-->>M1: deliver request
    M0->>Client: unicast  response{rpc_id=X, slimrpc-code=0, payload}
    M0->>Client: unicast  EOS{rpc_id=X, slimrpc-code=0, empty}
    M1->>Client: unicast  response{rpc_id=X, slimrpc-code=0, payload}
    M1->>Client: unicast  EOS{rpc_id=X, slimrpc-code=0, empty}
```

Because responses are unicast directly to the client, other members never receive them, keeping each member's session free of cross-member traffic. A future extension will introduce an explicit mechanism to deliver peer responses to members for context propagation, allowing members to observe each other's responses within the same GROUP session.

---

### Unary-Unary multicast

```mermaid
sequenceDiagram
    participant C  as Client
    participant M0 as Member 0
    participant M1 as Member 1

    par broadcast
        C->>M0: request{service, method, rpc_id=X, payload}
    and
        C->>M1: request{service, method, rpc_id=X, payload}
    end

    M0->>C: response{rpc_id=X, payload}
    M0->>C: EOS{rpc_id=X, empty}

    M1->>C: response{rpc_id=X, payload}
    M1->>C: EOS{rpc_id=X, empty}

    Note over C: eos_count reaches 2 → stream ends
    Note over C: yield MulticastItem{source=M0, ...}
    Note over C: yield MulticastItem{source=M1, ...}
```

The client counts EOS markers. Only when `eos_count == num_members` does the response stream terminate. Each data message is yielded as a `MulticastItem` carrying the source member's `Name`.

### Unary-Stream multicast

Each member streams multiple response items followed by one EOS.

```mermaid
sequenceDiagram
    participant C  as Client
    participant M0 as Member 0
    participant M1 as Member 1

    par broadcast
        C->>M0: request{service, method, rpc_id=X, payload}
    and
        C->>M1: request{service, method, rpc_id=X, payload}
    end

    M0->>C: item1{rpc_id=X}
    M0->>C: item2{rpc_id=X}
    M0->>C: EOS{rpc_id=X}

    M1->>C: item1{rpc_id=X}
    M1->>C: EOS{rpc_id=X}

    Note over C: eos_count=1 (M0 done)
    Note over C: eos_count=2 (M1 done) → stream ends
```

### Stream-Unary multicast

The client sends a stream of requests; each member accumulates them and replies with one response + EOS. Sending and receiving happen concurrently.

```mermaid
sequenceDiagram
    participant C  as Client
    participant M0 as Member 0
    participant M1 as Member 1

    par broadcast
        C->>M0: msg1{rpc_id=X, service, method}
    and
        C->>M1: msg1{rpc_id=X, service, method}
    end
    par broadcast
        C->>M0: msg2{rpc_id=X}
    and
        C->>M1: msg2{rpc_id=X}
    end
    par broadcast
        C->>M0: EOS{rpc_id=X}
    and
        C->>M1: EOS{rpc_id=X}
    end

    M0->>C: result{rpc_id=X, payload}
    M0->>C: EOS{rpc_id=X}

    M1->>C: result{rpc_id=X, payload}
    M1->>C: EOS{rpc_id=X}
```

On the server the first message (`service` + `method` present) spawns the handler task. Subsequent messages and the client EOS are routed to the handler via `pending_streams`.

### Stream-Stream multicast

Both sides stream. The client sends concurrently with reading responses.

```mermaid
sequenceDiagram
    participant C  as Client
    participant M0 as Member 0
    participant M1 as Member 1

    par broadcast
        C->>M0: msg1{rpc_id=X, service, method}
    and
        C->>M1: msg1{rpc_id=X, service, method}
    end

    M0->>C: echo1{rpc_id=X}   (unicast)
    M1->>C: echo1{rpc_id=X}   (unicast)

    par broadcast
        C->>M0: msg2{rpc_id=X}
    and
        C->>M1: msg2{rpc_id=X}
    end

    M0->>C: echo2{rpc_id=X}
    M1->>C: echo2{rpc_id=X}

    par broadcast
        C->>M0: EOS{rpc_id=X}
    and
        C->>M1: EOS{rpc_id=X}
    end

    M0->>C: EOS{rpc_id=X}
    M1->>C: EOS{rpc_id=X}

    Note over C: 4 items total, eos_count=2 → stream ends
```

---

## EOS protocol

Every server handler sends an explicit EOS after its final response, regardless of the interaction pattern:

| Pattern       | Server sends                           |
|---------------|----------------------------------------|
| Unary-Unary   | `response` → `EOS`                    |
| Unary-Stream  | `item…` → `EOS`                       |
| Stream-Unary  | `response` → `EOS`                    |
| Stream-Stream | `item…` → `EOS`                       |

The EOS is an empty-payload message with `slimrpc-code=0` (Ok). P2P callers drop the stream after reading the single item they expect, so the EOS arrives on a dropped channel and is discarded harmlessly. GROUP callers depend on it to know when all members have finished.

---

## Error handling

Member errors are yielded as `Err` stream items and count as that member's EOS, preventing the stream from hanging when one member fails.

```mermaid
flowchart LR
    A[receive message] --> B{slimrpc-code}
    B -->|0 + empty| C[EOS: eos_count++]
    B -->|0 + payload| D[yield MulticastItem]
    B -->|non-zero| E[yield Err: eos_count++]
    C --> F{eos_count == N?}
    E --> F
    F -->|yes| G[stream ends]
    F -->|no| A
```

---

## UniFFI bindings

Since UniFFI does not support Rust async streams, pull-based wrappers are provided:

| Type                      | Role                                        |
|---------------------------|---------------------------------------------|
| `MulticastResponseReader` | Reader for unary-input multicast results    |
| `MulticastBidiStreamHandler` | Concurrent send/receive for stream-input multicast |
| `MulticastStreamMessage`  | `Data(RpcMulticastItem) \| Error \| End`    |
| `RpcMulticastItem`        | Payload + `RpcMessageContext{source: Name}` |

`MulticastResponseReader` wraps an `mpsc::UnboundedReceiver` populated by a background task that runs the underlying async stream. `next()` (blocking) and `next_async()` (async) both return a `MulticastStreamMessage`.

`MulticastBidiStreamHandler` wraps `multicast_stream_stream`: a background task bridges the two mpsc channels (one for requests, one for responses) to the underlying async stream.

```mermaid
flowchart LR
    subgraph UniFFI caller
        A[send payload] --> B[MulticastBidiStreamHandler.send]
        C[MulticastBidiStreamHandler.recv] --> D[MulticastStreamMessage]
    end
    subgraph Rust async
        B --> E[request mpsc tx]
        E --> F[multicast_stream_stream]
        F --> G[response mpsc rx]
        G --> C
    end
```

msardara:
# Description

Fixes #1332

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/.github/workflows/reusable-bindings-build-and-test.yaml
+++ b/.github/workflows/reusable-bindings-build-and-test.yaml
@@ -226,7 +226,12 @@ jobs:
 
       - name: Generate Go bindings
         run: |
-          task generate PROFILE=release
+          # Extract version from tag name if present (e.g., slim-bindings-v1.3.0 -> v1.3.0)
+          VERSION="${{ inputs.tag-name }}"
+          VERSION="${VERSION#slim-bindings-}"
+          VERSION="${VERSION:-devel}"
+
+          task generate PROFILE=release VERSION="$VERSION"
 
           echo "📁 slim-bindings-go files:"
           find ./slim_bindings -type f
--- a/data-plane/Cargo.lock
+++ b/data-plane/Cargo.lock
@@ -4,7 +4,7 @@ version = 4
 
 [[package]]
 name = "agntcy-protoc-slimrpc-plugin"
-version = "1.2.0"
+version = "1.3.0-rc.1"
 dependencies = [
  "anyhow",
  "heck",
@@ -14,7 +14,7 @@ dependencies = [
 
 [[package]]
 name = "agntcy-slim"
-version = "1.3.0-rc.0"
+version = "1.3.0-rc.1"
 dependencies = [
  "agntcy-slim-config",
  "agntcy-slim-service",
@@ -80,7 +80,7 @@ dependencies = [
 
 [[package]]
 name = "agntcy-slim-bindings"
-version = "1.3.0-rc.0"
+version = "1.3.0-rc.1"
 dependencies = [
  "agntcy-slim",
  "agntcy-slim-auth",
@@ -357,11 +357,11 @@ dependencies = [
 
 [[package]]
 name = "agntcy-slim-version"
-version = "1.3.0-rc.0"
+version = "1.3.0-rc.1"
 
 [[package]]
 name = "agntcy-slimctl"
-version = "1.3.0-rc.0"
+version = "1.3.0-rc.1"
 dependencies = [
  "agntcy-slim",
  "agntcy-slim-config",
--- a/data-plane/Cargo.toml
+++ b/data-plane/Cargo.toml
@@ -38,9 +38,9 @@ resolver = "2"
 
 [workspace.dependencies]
 # Local dependencies
-agntcy-slim = { path = "core/slim", version = "1.3.0-rc.0" }
+agntcy-slim = { path = "core/slim", version = "1.3.0-rc.1" }
 agntcy-slim-auth = { path = "core/auth", version = "0.5.2" }
-agntcy-slim-bindings = { path = "bindings/rust", version = "1.3.0-rc.0" }
+agntcy-slim-bindings = { path = "bindings/rust", version = "1.3.0-rc.1" }
 agntcy-slim-config = { path = "core/config", version = "0.7.0" }
 agntcy-slim-controller = { path = "core/controller", version = "0.4.9" }
 agntcy-slim-datapath = { path = "core/datapath", version = "0.11.5" }
@@ -50,8 +50,8 @@ agntcy-slim-session = { path = "core/session", version = "0.1.8" }
 agntcy-slim-signal = { path = "core/signal", version = "0.1.5" }
 agntcy-slim-testing = { path = "testing" }
 agntcy-slim-tracing = { path = "core/tracing", version = "0.3.5" }
-agntcy-slim-version = { path = "core/version", version = "1.3.0-rc.0" }
-agntcy-slimctl = { path = "slimctl", version = "1.3.0-rc.0" }
+agntcy-slim-version = { path = "core/version", version = "1.3.0-rc.1" }
+agntcy-slimctl = { path = "slimctl", version = "1.3.0-rc.1" }
 
 anyhow = "1.0.98"
 
@@ -147,7 +147,7 @@ uuid = { version = "1.15.1", features = ["v4"] }
 wiremock = "0.6"
 
 [workspace.package]
-version = "1.3.0-rc.0"
+version = "1.3.0-rc.1"
 license = "Apache-2.0"
 edition = "2024"
 repository = "https://github.com/agntcy/slim"
--- a/data-plane/bindings/go/Taskfile.yaml
+++ b/data-plane/bindings/go/Taskfile.yaml
@@ -41,6 +41,8 @@ tasks:
   patch-cgo-directives:
     desc: Patch generated Go bindings with proper cgo directives
     internal: true
+    vars:
+      VERSION: '{{.VERSION | default "devel"}}'
     cmds:
       - echo "🔧 Patching cgo directives..."
       - |
@@ -52,11 +54,11 @@ tasks:
 
         /*
         #cgo CFLAGS: -I${SRCDIR}
-        #cgo linux,amd64 LDFLAGS: -L${SRCDIR} -L${SRCDIR}/../../../../../.cgo-cache/slim-bindings -lslim_bindings_x86_64_linux_gnu -lm
-        #cgo linux,arm64 LDFLAGS: -L${SRCDIR} -L${SRCDIR}/../../../../../.cgo-cache/slim-bindings -lslim_bindings_aarch64_linux_gnu -lm
-        #cgo darwin,amd64 LDFLAGS: -L${SRCDIR} -L${SRCDIR}/../../../../../.cgo-cache/slim-bindings -lslim_bindings_x86_64_darwin -Wl,-undefined,dynamic_lookup
-        #cgo darwin,arm64 LDFLAGS: -L${SRCDIR} -L${SRCDIR}/../../../../../.cgo-cache/slim-bindings -lslim_bindings_aarch64_darwin -Wl,-undefined,dynamic_lookup
-        #cgo windows,amd64 LDFLAGS: -L${SRCDIR} -L${SRCDIR}/../../../../../.cgo-cache/slim-bindings -lslim_bindings_x86_64_windows_gnu -lws2_32 -lbcrypt -ladvapi32 -luserenv -lntdll -lgcc_eh -lgcc -lkernel32 -lole32
+        #cgo linux,amd64 LDFLAGS: -L${SRCDIR} -L${SRCDIR}/../../../../../.cgo-cache/slim-bindings/{{.VERSION}} -lslim_bindings_x86_64_linux_gnu -lm
+        #cgo linux,arm64 LDFLAGS: -L${SRCDIR} -L${SRCDIR}/../../../../../.cgo-cache/slim-bindings/{{.VERSION}} -lslim_bindings_aarch64_linux_gnu -lm
+        #cgo darwin,amd64 LDFLAGS: -L${SRCDIR} -L${SRCDIR}/../../../../../.cgo-cache/slim-bindings/{{.VERSION}} -lslim_bindings_x86_64_darwin -Wl,-undefined,dynamic_lookup
+        #cgo darwin,arm64 LDFLAGS: -L${SRCDIR} -L${SRCDIR}/../../../../../.cgo-cache/slim-bindings/{{.VERSION}} -lslim_bindings_aarch64_darwin -Wl,-undefined,dynamic_lookup
+        #cgo windows,amd64 LDFLAGS: -L${SRCDIR} -L${SRCDIR}/../../../../../.cgo-cache/slim-bindings/{{.VERSION}} -lslim_bindings_x86_64_windows_gnu -lws2_32 -lbcrypt -ladvapi32 -luserenv -lntdll -lgcc_eh -lgcc -lkernel32 -lole32
         #include <slim_bindings.h>
         */
         import "C"
@@ -121,6 +123,8 @@ tasks:
       # exists and sources have not changed (sources/generates fingerprinting).
       - task: adapter:bindings:build:all
       - task: install-uniffi-bindgen-go
+    vars:
+      VERSION: '{{.VERSION | default "devel"}}'
     cmds:
       - echo "🔧 Generating Go bindings from compiled library..."
       - |
@@ -130,6 +134,8 @@ tasks:
       - task: copy-library
       - task: patch-header
       - task: patch-cgo-directives
+        vars:
+          VERSION: '{{.VERSION}}'
       - echo "✅ Bindings generated in slim_bindings/"
     silent: true
 
@@ -137,6 +143,8 @@ tasks:
     desc: Generate Go bindings using a cached native library
     deps:
       - task: install-uniffi-bindgen-go
+    vars:
+      VERSION: '{{.VERSION | default "devel"}}'
     cmds:
       - echo "🔧 Generating Go bindings from cached library..."
       - |
@@ -190,7 +198,7 @@ tasks:
         esac
 
         GOPATH_DIR=$(go env GOPATH)
-        CACHE_DIR="$GOPATH_DIR/.cgo-cache/slim-bindings"
+        CACHE_DIR="$GOPATH_DIR/.cgo-cache/slim-bindings/{{.VERSION}}"
         LIB_PATH="$CACHE_DIR/$LIB_NAME"
 
         if [ ! -f "$LIB_PATH" ]; then
@@ -204,6 +212,8 @@ tasks:
 
       - task: patch-header
       - task: patch-cgo-directives
+        vars:
+          VERSION: '{{.VERSION}}'
       - echo "✅ Bindings generated in slim_bindings/"
     silent: true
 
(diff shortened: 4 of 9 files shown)


data-plane/bindings/rust/src/slimrpc/channel.rs
```diff
@@ -376,88 +490,177 @@ impl Channel {
...
+                            )));
+                        }
+                        break;
+                    }
+                    Ok(Some(m)) => m,
+                };
+
+                if received.is_eos() {
+                    eos_count += 1;
+                    if eos_count >= num_members { break; }
```
micpapal: should generate an error if eos_count > num_members?

data-plane/bindings/rust/src/slimrpc/channel.rs
```diff
@@ -376,88 +490,177 @@ impl Channel {
...
+                    continue;
+                }
+
+                let code = RpcCode::from_metadata_str(
+                    received.metadata.get(STATUS_CODE_KEY).map(String::as_str)
+                );
+                if code != RpcCode::Ok {
+                    let msg_text = String::from_utf8_lossy(&received.payload).to_string();
+                    eos_count += 1;
```
micpapal: is it possible to receive multiple errors from the same remote? if yes the eos_counter is not enough and you need to track also the source of the packet.

data-plane/bindings/rust/src/slimrpc/channel.rs
```diff
@@ -127,124 +219,216 @@ fn first_msg_metadata(ctx: &Context, service: &str, method: &str, rpc_id: &str)
...
+    }
+
+    /// Invite a participant into the GROUP session.
+    ///
+    /// Creates the GROUP session if not yet established. Checks for duplicates
+    /// before any session work, then invites and inserts
+    ///
+    /// Returns an error if called on a P2P channel or if `destination` is
+    /// already a member.
+    pub async fn invite_participant(&self, destination: Name) -> Result<(), RpcError> {
```
micpapal: This should be removed as the group must remain static during the RPC

data-plane/bindings/rust/src/slimrpc/channel.rs
msardara: +1, I will add the source as well

data-plane/bindings/rust/src/slimrpc/channel.rs
```diff
@@ -127,124 +219,216 @@ fn first_msg_metadata(ctx: &Context, service: &str, method: &str, rpc_id: &str)
...
+/// when dead.
 ///
-/// A single SLIM session is maintained per remote and reused across concurrent
-/// RPC calls (see module-level documentation for the demultiplexing design).
+/// ## Constructor
+///
+/// - `new_with_members_internal(app, members)` — Smart constructor:
+///   - **1 member**: creates a P2P channel.
+///   - **Many members**: creates a GROUP channel with a generated session name
+///     and auto-invites all members on the first multicast call.
```
Tehsmash, line 238: Is there an advantage to supporting both P2P sessions and Group sessions? It seems like P2P sessions are essentially a sub-set of a group session i.e. a group with only 2 members (sender + receiver). I feel like we could simplify by only supporting one type of session.

No suggestion here just curious if there was overhead to a group session which a P2P session doesn't have.

data-plane/bindings/rust/src/slimrpc/channel.rs
```diff
@@ -127,124 +219,216 @@ fn first_msg_metadata(ctx: &Context, service: &str, method: &str, rpc_id: &str)
...
-    ) -> Result<(SessionTx, Arc<ResponseDispatcher>), RpcError> {
-        // Fast path: cheap read lock, no session creation needed.
+    ) -> Result<(SessionTx, Arc<ResponseDispatcher>, usize), RpcError> {
+        let mut guard = self.session.lock().await;
+        self.ensure_session(&mut guard).await?;
+        let cs = guard
+            .as_ref()
+            .ok_or_else(|| RpcError::internal("session missing after creation"))?;
+        Ok((cs.tx.clone(), cs.dispatcher.clone(), cs.members.len()))
+    }
```
Tehsmash, line 325: For some reason I'd envisioned us creating the Channel with an already existing session i.e.

```
session = slim.NewSession(...)
session.invite(...)
session.invite(...)

channel = slim.NewChannel(session)
```

Then internally in channel doing:

```
number_of_participants = session.members.len()
```

This more easily allows for a client which has been invited to a group session to initialise an RPC channel over that session. For example moderator App creates channel Foo and invites Agent A, Agent B and Agent C. A needs to talk to B and C so creates a channel from the session, and B wants to talk to A and C so creates a channel from the same session.

In this case A, B and C keep track of the sessions they are members of similar to Webex/Discord channels, and then pick which session they want to use based on who they are interested in talking too.

This also removes the need for all the session state management in the Channel pushing it up a level so that the Channel only needs to worry about tracking and demultiplexing the RPC messages

data-plane/bindings/rust/src/slimrpc/channel.rs
```diff
@@ -376,88 +490,177 @@ impl Channel {
...
     }
 
-    /// Make a unary RPC call
+    // ── RPC methods ───────────────────────────────────────────────────────────
+
+    /// Unary RPC: single request → single response.
     ///
-    /// Sends a single request and receives a single response.
+    /// Equivalent to `unary_stream(...).next()`.
     pub async fn unary<Req, Res>(
```
Tehsmash, line 633: Do you think we need multicast_unary and unary exposed? vs just making unary return a list<Res> where len of the list is the number of participants? It could even be a Dict<string, Res> where key is the participant local name removing the need for the MulticastItem wrapper.

Similarly for unary_stream the response would be list<Stream<Item>> or Dict<string, Stream<item>>.

Personal preference is the Dict version.

unary -> Dict<string, Res>
unary_stream -> Dict<string, Stream<item>>
stream_unary -> Dict<string, Res>
stream_stream -> Dict<string, Stream<item>>

(Please excuse the messy half python/half rust syntax, but you get the idea)

This then allows the application on top of the Channel like a2a to enumerate the responses separately and process the responses from each participants in parallel etc.

data-plane/bindings/rust/src/slimrpc/channel.rs
msardara: I'd say there is no real advantage - the p2p is just like a group session of 2 participants. However I'd keep the 2 for the moment as the P2P session uses direct app names instead of setting up routes for a group.

data-plane/bindings/rust/src/slimrpc/channel.rs
msardara: I like the idea 💡 Let's address it in a different PR as I think it might introduce some braking changes, but in principle removing the burden of managing the session in the channel would be great 🔝

data-plane/bindings/rust/src/slimrpc/channel.rs
msardara: At a first I wanted to do that, but that would introduce a breaking change as the unary API is public :/

The way it is done now allows the channel to receive responses from potentially many server as soon as they arrive, e.g. a multicast_unary to 10 servers would return the 10 reponses in a stream, so that we are not blocked by the slowest server.

We could add an API where we wait for all responses before returning, that would make sense if a client does not need the reponses one by one in a stream but can wait for the slowest responder. I'll do it in another PR ✅

data-plane/bindings/rust/src/slimrpc/channel.rs
msardara: It is currently not possible to receive more than one error from the same remote, but I will add a map anyway to track who sent the error

PR #1353, opened 2026-03-18
feat: add link negotiation protocol between SLIM nodes

### LINKED ISSUE #1351: Add link negotiation protocol between SLIM nodes
### Summary

When two SLIM nodes establish a connection, they should perform a link-local handshake to agree on a shared link identifier and exchange version information. This would enable future features to gate behavior based on the capabilities of the remote peer.

### Motivation

Currently, SLIM nodes have no way to:
  - Identify a specific link between two peers beyond the connection index
  - Know the SLIM version running on the remote side
  - Conditionally enable features based on what the remote peer supports

### Proposed Design

A new top-level Link message type should be added to the protobuf definition. Unlike Publish, Subscribe, and Unsubscribe, Link messages must be link-local: never routed, never forwarded, and never sent on local (application-facing) connections.

### Protocol

  1. The connecting side (client) generates a link_id (random UUID v4, or a user-configured value from ClientConfig) and sends a LinkNegotiationPayload with is_reply: false.
  2. The accepting side (server) stores the link_id and the sender's version, then echoes back a reply with the same link_id, its own version, and is_reply: true.
  3. The is_reply flag prevents echo loops.

### Backwards compatibility

Old SLIM nodes do not know about the new Link proto field. They should silently ignore the unknown field and the connection should continue to work normally. A None remote version should be treated as "peer is old or negotiation has not completed".

### Requirements

  - proto: add a Link message and LinkNegotiationPayload (link_id, slim_version, is_reply) as a top-level message type, distinct from session messages
  - Connection: expose link_id and remote_slim_version as interior-mutable fields (connection is stored behind Arc)
  - ClientConfig: add an optional link_id field; if absent, a random UUID v4 should be generated at connection time
  - ProtoMessageBuilder: add a build_link_negotiation() method consistent with the existing builder pattern
  - MessageProcessor: intercept Link messages before any SLIM-header-dependent processing; skip header clearing and telemetry for Link messages in send_msg
  - versioning: remote version must be stored as semver::Version (not a raw string) to allow comparisons like v >= semver::Version::new(1, 2, 1)
  - Link messages must never be sent or processed on local connections

msardara:
# Description

Fixes #1351 

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/bindings/python/examples/common.py
+++ b/data-plane/bindings/python/examples/common.py
@@ -17,6 +17,7 @@ import argparse
 import base64  # Used to decode base64-encoded JWKS content (when provided).
 import datetime  # Used for timedelta in JWT configs
 import json  # Used for parsing JWKS JSON and dynamic option values.
+from pathlib import Path
 from typing import Any
 
 import slim_bindings  # The Python bindings package we are demonstrating.
@@ -220,7 +221,17 @@ async def create_local_app(config: BaseConfig) -> tuple[slim_bindings.App, int]:
     # Convert local identifier to a strongly typed Name.
     local_name = slim_bindings.Name.from_string(config.local)
 
-    client_config = slim_bindings.new_insecure_client_config(config.slim)
+    if config.slim_client_config:
+        json_text = Path(config.slim_client_config).read_text(encoding="utf-8")
+        try:
+            client_config = slim_bindings.new_config_from_json(json_text)
+        except slim_bindings.SlimError as e:
+            raise RuntimeError(
+                f"Invalid slim client config JSON ({config.slim_client_config}): {e}"
+            ) from e
+    else:
+        client_config = slim_bindings.new_insecure_client_config(config.slim)
+
     conn_id = await service.connect_async(client_config)
 
     # Determine authentication mode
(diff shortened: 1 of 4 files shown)


data-plane/core/controller/src/service.rs
```diff
@@ -403,6 +403,9 @@ impl ControlPlane {
                                                     debug!("Ignoring publish message with session type: {:?}", msg.get_session_message_type());
                                                 }
                                             }
+                                            LinkType(_) => {
+                                                debug!("Ignoring link message received by controller");
```
micpapal: These are not the messages coming from the controller but from the local data plane. I don't see any problem but the debug message is wrong.

data-plane/core/datapath/src/message_processing.rs
```diff
@@ -112,6 +119,17 @@ fn create_span(function: &str, out_conn: u64, msg: &Message) -> Span {
     span
 }
 
+fn local_version() -> &'static str {
+    slim_version::version()
+}
+
+fn is_valid_uuid_v4(s: &str) -> bool {
```
micpapal: This same check is also con config/src/grcp/client.rs (line 414). Maybe this function can go in messages/utils.rs

data-plane/core/datapath/src/message_processing.rs
```diff
@@ -449,6 +487,119 @@ impl MessageProcessor {
...
+        };
+
+        if payload.is_reply {
+            // Client path: verify the server echoed back the link_id we sent.
+            if conn.link_id().as_deref() != Some(link_id.as_str()) {
+                debug!(%in_connection, %link_id, "ignoring link negotiation reply with mismatched link_id");
+                return Ok(());
+            }
+            // Atomically check-and-set; returns false if already negotiated (replay protection).
+            if !conn.complete_negotiation(None, version) {
```
micpapal: Here instead to set the link_id to None in the complete_negotiation we can do the check
```
conn.link_id().as_deref() != Some(link_id.as_str()
```
inside the complete_negotiation and move all the checks needed inside that functions
Some comment for the else branch

Issue #1363, opened 2026-03-20
Implement reliable subscription delivery with ACK and retry for remote subscriptions


msardara:
## Problem

When a SLIM node forwards a subscription to an upstream relay, the operation is fire-and-forget. There is no confirmation that the upstream relay had successfully registered the subscription. If the subscription message was lost or the upstream relay encountered an error, the subscriber would silently never receive matching publications — with no indication that anything was wrong.

## Solution

Introduce a remote subscription ACK protocol between SLIM nodes (v ≥ 1.2.0). When forwarding a subscription to an upstream relay that supports ACKs, the forwarding node:

  1. Attaches a unique ack_id to the subscribe message.
  2. Waits for an explicit SubscriptionAck response from the upstream relay.
  3. Retries up to 3 times with a 5-second timeout per attempt if no ACK is received.
  4. Sends a negative ACK (with error) back to the original subscriber if all retries are exhausted.

The upstream relay sends a SubscriptionAck (success or failure) after processing the subscription.

PR #1364, opened 2026-03-20
feat: ack for remote subscriptions

msardara:
# Description

Fixes: #1363 

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/core/datapath/src/message_processing.rs
+++ b/data-plane/core/datapath/src/message_processing.rs
@@ -545,7 +545,10 @@ impl MessageProcessor {
         );
 
         for info in subs {
-            let subscription_id = rand::random::<u64>();
+            let subscription_id = match info.subscription_id() {
+                0 => rand::random::<u64>(),
+                id => id,
+            };
             let msg = Message::builder()
                 .source(info.source().clone())
                 .destination(info.name().clone())
@@ -749,8 +752,14 @@ impl MessageProcessor {
                 let identity = msg.get_identity();
 
                 self.send_msg(msg, out_conn).await.map(|_| {
-                    self.forwarder()
-                        .on_forwarded_subscription(source, dst, identity, out_conn, add);
+                    self.forwarder().on_forwarded_subscription(
+                        source,
+                        dst,
+                        identity,
+                        out_conn,
+                        add,
+                        subscription_id,
+                    );
                 })
             }
         }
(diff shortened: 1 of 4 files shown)


data-plane/core/datapath/proto/v1/data_plane.proto
```diff
@@ -255,6 +260,18 @@ message GroupNackPayload {}
 // Ping
 message PingPayload {}
 
+// SubscriptionAck is delivered directly to the requesting connection in response
+// to a Subscribe or Unsubscribe that carried a non-empty ack_id field.
+// It is never routed through the subscription table.
+message SubscriptionAck {
+  // Echoes the ack_id from the originating Subscribe or Unsubscribe request.
+  string ack_id = 1;
```
micpapal: this I think can be an integer instead of a string

data-plane/core/datapath/src/message_processing.rs
```diff
@@ -504,15 +513,56 @@ impl MessageProcessor {
         }
     }
 
+    /// Re-send every subscription already forwarded on `out_conn` using the remote
+    /// ack path.  Called after link negotiation completes so that subscriptions
+    /// that were initially forwarded via the default path (before the remote peer's
+    /// version was known) are now confirmed with a full ACK round-trip.
+    fn upgrade_forwarded_subscriptions(&self, out_conn: u64) {
```
micpapal, line 531: And what happens here if there is no link negotiation?
Why do you need to send them twice? isn't it better to buffer the subscriptions and send them after the link negotiation?

msardara: Yes good point, I fixed it in this way: subscription ref coutner are kept only for local connections, they are ignored for remote connections. So when the node re-send the subscribe message, it will not increment the ref counter.

Buffering unfortunately is not an option as an old slim node will not reply with an ack, so we may need to delay the forward of the subscription without the guarantee to receive a reply.

data-plane/core/datapath/src/message_processing.rs
```diff
@@ -504,15 +524,56 @@ impl MessageProcessor {
...
+        }
+
+        debug!(
+            %out_conn,
+            count = subs.len(),
+            "link negotiation complete: upgrading forwarded subscriptions to remote ack path",
+        );
+
+        for info in subs {
+            let subscription_id = rand::random::<u64>();
```
msardara: Here we shohuld reuse the existing ID

PR #1394, opened 2026-03-23
feat: MLS identity key integration and security dependency upgrades

msardara:
# Description

Upgrade to MLS 0.54 to address security advisories in mls dependencies.

This needed some work as latest version of MLS does not support work with async calls,
so changed a bit the logic. Check https://github.com/awslabs/mls-rs/issues/325

Now the identity providers generates the keys for MLS, allowing to remove the async
dependency get_token_with_claims, which was passing the keys generated by MLS to the
identity provider.

In details:
  - Embed MLS Ed25519 signature public keys transparently in identity tokens issued by all identity providers
  - Move IdentityClaims from slim-mls to slim-auth as the canonical identity abstraction
  - Remove get_token_with_claims from the TokenProvider trait and all implementations

Also:
 - update most of the dependency to compaible versions
 - Upgrade aws-lc-sys and rustls-webpki to resolve 5 security advisories:
     - https://rustsec.org/advisories/RUSTSEC-2026-0048 — High (7.4): CRL Distribution Point scope check logic error in AWS-LC (aws-lc-sys ≥0.39.0)
    - https://rustsec.org/advisories/RUSTSEC-2026-0047 — High (7.5): PKCS7_verify signature validation bypass in AWS-LC (aws-lc-sys ≥0.38.0)
    - https://rustsec.org/advisories/RUSTSEC-2026-0046 — High (7.5): PKCS7_verify certificate chain validation bypass in AWS-LC (aws-lc-sys ≥0.38.0)
    - https://rustsec.org/advisories/RUSTSEC-2026-0045 — Medium (5.9): Timing side-channel in AES-CCM tag verification in AWS-LC (aws-lc-sys ≥0.38.0)
    - https://rustsec.org/advisories/RUSTSEC-2026-0049 — CRLs not considered authoritative by distribution point due to faulty matching logic (rustls-webpki ≥0.103.10)

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/.github/workflows/benchmark.yaml
+++ b/.github/workflows/benchmark.yaml
@@ -1,0 +1,73 @@
+# Copyright AGNTCY Contributors (https://github.com/agntcy)
+# SPDX-License-Identifier: Apache-2.0
+
+---
+name: benchmark-data-plane
+
+on:
+  pull_request:
+    paths:
+      - '.github/workflows/benchmark.yaml'
+      - 'data-plane/**'
+      - 'scripts/run-benchmark.sh'
+      - 'scripts/flamegraph-profile.sh'
+  workflow_dispatch:
+    inputs:
+      senders:
+        description: 'Comma-separated sender counts'
+        required: false
+        default: '1,2,4,8'
+        type: string
+      messages:
+        description: 'Messages per sender'
+        required: false
+        default: '1000000'
+        type: string
+      payload-sizes:
+        description: 'Comma-separated payload sizes in bytes'
+        required: false
+        default: '8,16,64,256,1024,4096'
+        type: string
+      runs:
+        description: 'Repetitions per configuration'
+        required: false
+        default: '3'
+        type: string
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+jobs:
+  benchmark:
+    name: Data plane benchmark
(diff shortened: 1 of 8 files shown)


data-plane/core/auth/src/jwt.rs
```diff
@@ -196,8 +203,13 @@ impl<T> Jwt<T> {
     ///     .build()?;
     /// ```
     #[allow(clippy::too_many_arguments)]
```
micpapal: is this still needed?

data-plane/core/session/src/mls_state.rs
```diff
@@ -47,7 +47,7 @@ where
     V: Verifier + Send + Sync + Clone + 'static,
 {
     pub(crate) async fn new(mut mls: Mls<P, V>) -> Result<Self, SessionError> {
```
micpapal: looks like there is no need anymore to use an async fn. And the same for the other functions here.

data-plane/core/auth/src/jwt.rs
msardara: Nope, let's remove it

data-plane/core/session/src/mls_state.rs
msardara: Done

PR #1397, opened 2026-03-24
perf(testing): avoid per-message allocs in benchmark hot loop

muscariello:
## Summary

Eliminate redundant heap allocations in the `run_benchmark` sender hot loop.

## Problem

Each iteration of the message-send loop was rebuilding the full `ProtoMessage` from scratch:

- `sender_name.clone()` — `Box<[String; 3]>` + 3 heap-allocated strings
- `topic.clone()` — same
- `content_clone.clone()` — `String` + `Vec<u8>` copy
- `build_publish()` — `SlimHeader::new()` re-encodes both `Name`s into proto wire format, allocates a new `SessionHeader`, `HashMap`, etc.

None of these values change between messages, so all of this was pure overhead on every one of the potentially millions of loop iterations.

## Fix

Build one `template: ProtoMessage` per sender task before the loop, moving all owned values directly into the builder (zero clones at construction time). The hot loop body is then reduced to a single `template.clone()` of the already-encoded struct — the minimum unavoidable work since the channel takes ownership of each message.

```rust
// Before (per-iteration):
let msg = Message::builder()
    .source(sender_name.clone())       // alloc Box + 3 Strings
    .destination(topic.clone())        // alloc Box + 3 Strings
    .payload(content_clone.clone())    // alloc String + Vec<u8>
    .build_publish().unwrap();         // alloc SlimHeader, SessionHeader, HashMap

// After (once per sender task, then):
let template = Message::builder()
    .source(sender_name)               // move, no alloc
    .destination(topic)                // move, no alloc
    .payload(content_clone)            // move, no alloc
    .build_publish().unwrap();

for _ in 0..msg_count {
    tx.send(Ok(template.clone()))...   // one struct clone only
}
```

## Results

Quick benchmark (macOS M-series, 500K msgs/sender):

| Senders | Before recv MPS | After recv MPS | Delta |
|---------|----------------|----------------|-------|
| 1       | ~900K          | ~911–977K      | +2–8% |
| 2       | ~1.4M          | ~1.53–1.56M    | +~8%  |
| 4       | ~1.7M          | ~1.93–2.23M    | +14–32% |
| 8       | ~1.4M          | ~1.40–1.50M    | similar (lock-contention bound) |

100% delivery maintained on all 16 configurations.

## Testing

All 17 unit tests pass. No functional change.

### DIFF
--- a/data-plane/testing/src/stress.rs
+++ b/data-plane/testing/src/stress.rs
@@ -160,15 +160,19 @@ pub async fn run_benchmark(
         let content_clone = content.clone();
 
         let handle = tokio::spawn(async move {
-            for _ in 0..msg_count {
-                let msg = Message::builder()
-                    .source(sender_name.clone())
-                    .destination(topic.clone())
-                    .payload(content_clone.clone())
-                    .build_publish()
-                    .unwrap();
+            // Build the publish message template once, moving all owned values
+            // in (no clone). Per-iteration we only pay for one ProtoMessage
+            // clone instead of two Name clones + one Content clone + a full
+            // SlimHeader/SessionHeader construction.
+            let template = Message::builder()
+                .source(sender_name)
+                .destination(topic)
+                .payload(content_clone)
+                .build_publish()
+                .unwrap();
 
-                if tx.send(Ok(msg)).await.is_err() {
+            for _ in 0..msg_count {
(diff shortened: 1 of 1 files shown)


micpapal (commented):
lgtm

micpapal (approved):

PR #1410, opened 2026-03-26
perf(rnd): Random number generation per match

### LINKED ISSUE #1384: Random number generation per match
**Problem:**
```rust
let mut rng = rand::rng();
let index = rng.random_range(0..self.pool.max_set() + 1);
```
A new RNG is seeded on every call.  `rand::rng()` reads from the OS entropy
source or thread-local state.  The subsequent linear scan of the pool from a
random start is O(pool capacity), not O(active connections).

**Mitigation plan:**
1. Use a thread-local `SmallRng` (no syscall, no reseeding).
2. Scan only active entries (use the bitmap or index map) instead of
   iterating over empty slots.

lpuskas:
Fixes #1384 

## Summary

Replace random output-connection selection in the subscription datapath with **per-thread round-robin** via a shared `NEXT_INDEX` cursor and a **`pick_one`** helper used by **`Connections::get_one`** and **`NameState::get_one_connection`**.

## Why

Deterministic, cheaper selection without RNG; predictable spread across candidates.

## Key changes

- `next_index(n)` + `pick_one(n, except, get)` for ring-ordered scans.
- Pool picks: indices over `0..pool.max_set()+1`.
- Named picks: collect IDs, **sort** for stable order, `pick_one` skips `incoming_conn` via `except`.

## Testing

`cargo test -p agntcy-slim-datapath subscription_table` (and CI datapath tests).

## Note

**Behavior change:** selection is round-robin, not random; pool and name paths share the same thread-local cursor.

### DIFF
--- a/data-plane/core/datapath/src/tables/subscription_table.rs
+++ b/data-plane/core/datapath/src/tables/subscription_table.rs
@@ -193,38 +193,44 @@ impl Connections {
         Ok(())
     }
 
-    /// Next pool slot index in round-robin order among active slots whose connection id is not
-    /// `except_conn`. Updates thread-local [`LAST_USED_POOL_INDEX`].
+    /// Next pool slot index for a connection other than `except_conn`.
+    ///
+    /// Single-connection case is O(1) on [`Self::index`]. Otherwise walks pool slots in ring order
+    /// starting after the thread-local last index (see `next_slot` below), without allocating.
     fn next_round_robin_pool_index(&self, except_conn: u64) -> Option<usize> {
-        // Build eligible slot indices using `max_set`/`get` (occupied slots with conn != except).
-        let mut eligible = Vec::new();
-        for idx in 0..=self.pool.max_set() {
-            if self.pool.get(idx).is_some_and(|c| c.conn_id != except_conn) {
-                eligible.push(idx);
+        if self.index.len() == 1 {
+            if self.index.contains_key(&except_conn) {
+                debug!("the only available connection cannot be used");
+                return None;
+            }
+            let (&conn_id, &slot_idx) = self.index.iter().next().unwrap();
+            debug_assert!(self.pool.get(slot_idx).is_some_and(|c| c.conn_id == conn_id));
+            LAST_USED_POOL_INDEX.with(|c| c.set(slot_idx));
+            return Some(slot_idx);
+        }
+
+        let num_slots = self.pool.max_set() + 1;
+        let last_raw = LAST_USED_POOL_INDEX.with(|c| c.get());
+        // First slot to try: advance one past last pick, wrapped into [0, num_slots).
+        let next_slot = if last_raw == LAST_USED_POOL_INDEX_NONE {
+            0
+        } else {
+            (last_raw + 1) % num_slots
+        };
+
+        // Ring traversal without per-step `%`: indices next_slot..end, then 0..next_slot.
+        for i in (next_slot..num_slots).chain(0..next_slot) {
+            if let Some(c) = self.pool.get(i)
+                && c.conn_id != except_conn
+            {
+                LAST_USED_POOL_INDEX.with(|c| c.set(i));
+                return Some(i);
             }
         }
 
-        if eligible.is_empty() {
-            LAST_USED_POOL_INDEX.with(|c| c.set(LAST_USED_POOL_INDEX_NONE));
-            debug!("no output connection available");
-            return None;
-        }
-
-        let n = eligible.len();
-        let last_raw = LAST_USED_POOL_INDEX.with(|c| c.get());
-        let next_pos = if last_raw == LAST_USED_POOL_INDEX_NONE {
-            0
-        } else {
-            eligible
-                .iter()
-                .position(|&idx| idx == last_raw)
-                .map(|p| (p + 1) % n)
-                .unwrap_or(0)
-        };
-
-        let idx = eligible[next_pos];
-        LAST_USED_POOL_INDEX.with(|c| c.set(idx));
-        Some(idx)
+        LAST_USED_POOL_INDEX.with(|c| c.set(LAST_USED_POOL_INDEX_NONE));
+        debug!("no output connection available");
+        None
     }
 
     fn get_one(&self, except_conn: u64) -> Option<u64> {
--- a/data-plane/core/datapath/src/tables/subscription_table.rs
+++ b/data-plane/core/datapath/src/tables/subscription_table.rs
@@ -1,7 +1,7 @@
 // Copyright AGNTCY Contributors (https://github.com/agntcy)
 // SPDX-License-Identifier: Apache-2.0
 
-use std::cell::{Cell, RefCell};
+use std::cell::Cell;
 use std::collections::{HashMap, HashSet};
 use std::fmt::{Display, Formatter};
 use std::hash::{Hash, Hasher};
@@ -15,14 +15,17 @@ use crate::errors::DataPathError;
 use crate::messages::Name;
 
 thread_local! {
-    /// Per-thread round-robin state for [`Connections::get_one`]: last **used** pool slot
-    /// (initialized to `0` before the first pick).
-    static LAST_USED_POOL_INDEX: Cell<usize> = const { Cell::new(0_usize) };
-    /// Per-name round-robin (RR): last **used connection position** (index in the sorted candidate list) for
-    /// [`NameState::get_one_connection`] (key = `(name id, 0 = local pool / 1 = remote pool)`).
-    /// Missing keys use `0_usize` like [`LAST_USED_POOL_INDEX`] before the first pick for that key.
-    static LAST_USED_CONN_POS: RefCell<HashMap<(u64, u8), usize>> =
-        RefCell::new(HashMap::new());
+    /// Per-thread cursor for [`next_index`]: used by [`Connections::get_one`] and [`NameState::get_one_connection`].
+    static NEXT_INDEX: Cell<usize> = const { Cell::new(0_usize) };
+}
+
+/// Returns the current [`NEXT_INDEX`] modulo `n`, then sets it to `(that + 1) % n`.
+fn next_index(n: usize) -> usize {
+    NEXT_INDEX.with(|cell| {
+        let i = cell.get() % n;
+        cell.set(i + 1);
+        i
+    })
 }
 
 #[derive(Debug, Clone)]
@@ -190,11 +193,9 @@ impl Connections {
         Ok(())
     }
 
-    /// Pick a connection id other than `except_conn` using round-robin over pool slots.
-    ///
-    /// Single-connection case is O(1) on [`Self::index`]. Otherwise walks pool slots in ring order
-    /// starting from one slot after the thread-local last-used index, without allocating.
+    /// Pick a connection id other than `except_conn` using pseudo-random selection
     fn get_one(&self, except_conn: u64) -> Option<u64> {
+        let num_slots = self.pool.max_set() + 1;
         if self.index.len() == 1 {
             let (&conn_id, &slot_idx) = self
                 .index
@@ -210,23 +211,15 @@ impl Connections {
                     .get(slot_idx)
                     .is_some_and(|c| c.conn_id == conn_id)
             );
-            LAST_USED_POOL_INDEX.with(|c| c.set(slot_idx));
+            let _ = next_index(num_slots);
             return Some(conn_id);
         }
 
-        let num_slots = self.pool.max_set() + 1;
-        // First slot to try: one past last used, wrapped into [0, num_slots).
-        // Read last-used slot once; compute where to start this scan (no write here).
-        let last_used = LAST_USED_POOL_INDEX.with(|c| c.get());
-        let next_slot = (last_used + 1) % num_slots;
-
-        // Ring traversal without per-step `%`: indices next_slot..end, then 0..next_slot.
-        for i in (next_slot..num_slots).chain(0..next_slot) {
+        let start = next_index(num_slots);
+        for i in (start..num_slots).chain(0..start) {
             if let Some(c) = self.pool.get(i)
                 && c.conn_id != except_conn
             {
-                // Store the last **used** slot; next call will start at `(i + 1) % num_slots`.
-                LAST_USED_POOL_INDEX.with(|c| c.set(i));
                 return Some(c.conn_id);
             }
         }
@@ -388,7 +381,6 @@ impl NameState {
             return None;
         }
 
-        let rr_key = (id, side as u8);
         let mut non_incoming_conn_ids: Vec<u64> = Vec::with_capacity(refs.len());
         non_incoming_conn_ids.extend(refs.keys().copied().filter(|&c| c != incoming_conn));
 
@@ -404,14 +396,8 @@ impl NameState {
         non_incoming_conn_ids.sort_unstable();
         let n = non_incoming_conn_ids.len();
 
-        let conn_id = LAST_USED_CONN_POS.with(|m| {
-            let mut map = m.borrow_mut();
-            let last_pos = *map.get(&rr_key).unwrap_or(&0_usize);
-            let next_pos = (last_pos + 1) % n;
-            let picked = non_incoming_conn_ids[next_pos];
-            map.insert(rr_key, next_pos);
-            picked
-        });
+        let idx = next_index(n);
+        let conn_id = non_incoming_conn_ids[idx];
 
         Some(conn_id)
     }
--- a/data-plane/core/datapath/src/tables/subscription_table.rs
+++ b/data-plane/core/datapath/src/tables/subscription_table.rs
@@ -208,7 +208,11 @@ impl Connections {
                 debug!("the only available connection cannot be used");
                 return None;
             }
-            debug_assert!(self.pool.get(slot_idx).is_some_and(|c| c.conn_id == conn_id));
+            debug_assert!(
+                self.pool
+                    .get(slot_idx)
+                    .is_some_and(|c| c.conn_id == conn_id)
+            );
             LAST_USED_POOL_INDEX.with(|c| c.set(slot_idx));
             return Some(slot_idx);
         }
--- a/data-plane/core/datapath/src/tables/subscription_table.rs
+++ b/data-plane/core/datapath/src/tables/subscription_table.rs
@@ -18,10 +18,10 @@ thread_local! {
     /// Per-thread round-robin state for [`Connections::get_one`]: last **used** pool slot
     /// (initialized to `0` before the first pick).
     static LAST_USED_POOL_INDEX: Cell<usize> = const { Cell::new(0_usize) };
-    /// Last picked index in the sorted non-incoming connection id list for [`NameState::get_one_connection`]
-    /// (key = `(name id, 0 = local pool / 1 = remote pool)`). Missing keys use `0_usize` like
-    /// [`LAST_USED_POOL_INDEX`] before the first pick for that key.
-    static LAST_USED_NON_INCOMING_CONN_RR_POS: RefCell<HashMap<(u64, u8), usize>> =
+    /// Per-name round-robin (RR): last **used connection position** (index in the sorted candidate list) for
+    /// [`NameState::get_one_connection`] (key = `(name id, 0 = local pool / 1 = remote pool)`).
+    /// Missing keys use `0_usize` like [`LAST_USED_POOL_INDEX`] before the first pick for that key.
+    static LAST_USED_CONN_POS: RefCell<HashMap<(u64, u8), usize>> =
         RefCell::new(HashMap::new());
 }
 
@@ -404,7 +404,7 @@ impl NameState {
         non_incoming_conn_ids.sort_unstable();
         let n = non_incoming_conn_ids.len();
 
-        let conn_id = LAST_USED_NON_INCOMING_CONN_RR_POS.with(|m| {
+        let conn_id = LAST_USED_CONN_POS.with(|m| {
             let mut map = m.borrow_mut();
             let last_pos = *map.get(&rr_key).unwrap_or(&0_usize);
             let next_pos = (last_pos + 1) % n;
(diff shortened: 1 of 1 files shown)


data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -1,19 +1,28 @@
...
 use super::pool::Pool;
 use crate::errors::DataPathError;
 use crate::messages::Name;
 
+thread_local! {
+    static THREAD_RNG: RefCell<SmallRng> = RefCell::new(SmallRng::from_os_rng());
+}
+
+fn random_index(upper: usize) -> usize {
+    THREAD_RNG.with(|rng| rng.borrow_mut().random_range(0..upper))
```
micpapal: I think that the main performance issue was the generation of the random value. These modifications should already improve the performance by removing the per call rand::rng(). 

However I think we can remove the random generation completely by implementing a round robin on the available connections. You just need to keep track of the index of the last used connection. When you need to send a packet you send it on last_used + 1 % self.pool.max_set(). You can reuse a loop similar to the one that was already there to keep increase the index as long as you don't find a usable connection.

lpuskas: In my understanding setting up the randomizer is the expensive step (not the generation itself) and that was executed on every call; as per the current change an instance of the randomizer is attached to every thread and reused by subsequent calls eventually (this is the optimization in this case - along with the scan over the active indexes only)

If we removed the randomizer and kept track of the last used connection - what if the connection is removed between two calls (we can track that as well and figure out another approach for the first used conn, but i thin that can be more expensive than the current approach); in my understanding

data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -126,13 +140,16 @@ struct Connections {
     index: HashMap<u64, usize>,
     // pool of all connections ids that can to be used in the match
     pool: Pool<ConnId>,
+    /// Last pool slot index chosen by round-robin ([`LAST_USED_POOL_INDEX_NONE`] until first use).
+    last_used_pool_index: AtomicUsize,
```
micpapal: I notice that the subscription table takes a read lock on match, so I am not sure this index can be modified here. One way to remove the problem is to make the last_used_pool_index thread local as you did it a line 20. So each thread will have its own index that can be used as base for the selection of the next. This will not be a real round-robin but in the end we just want something that gives a random selection and this will work without the need of extra locks. As usual the increment of the value % self.pool.max_set() + 1 keeps as safe for overflow as max_set()  is specific for each pool instance.

msardara: Yes, and even if AtomicUsize removes the necessity of keeping a write lock, it is still more expensive than a normal integer as updating it triggers a cache sync mechanism between cores not to keep a stale value.

data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -179,36 +196,44 @@ impl Connections {
...
-            }
-            i = (i + 1) % (self.pool.max_set() + 1);
-            if i == index {
-                stop = true;
-            }
+        if eligible.is_empty() {
+            self.last_used_pool_index
+                .store(LAST_USED_POOL_INDEX_NONE, Ordering::Relaxed);
+            debug!("no output connection available");
+            return None;
```
micpapal: this check can be quite costly if the pool is large. It should not be the case most of the time but you don't need to compute the eligible vec. You can handle size = 1 as special case and the rest with the increment on the last used index. So here you can do something like what was in main

```suggestion
        if self.index.len() == 1 {
            let out = self.index.iter().filter(|(k, _)| *k != &except_conn).next();
            if let Some(val) = out {
                return Some(*val.0);
            } else {
                return None; 
            }
        }
```

data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -179,36 +196,44 @@ impl Connections {
...
         }
-        debug!("no output connection available");
-        None
+
+        let n = eligible.len();
+        let last_raw = self.last_used_pool_index.load(Ordering::Relaxed);
+        let next_pos = if last_raw == LAST_USED_POOL_INDEX_NONE {
```
micpapal: here eligible is not need again.  You can 
```suggestion
        let stop = (LAST_USED + 1) % (self.pool.max_set() + 1)
        let i = stop
        while true {
           // check is i is usable an return i if possible and set the new index in LAST_USED
           // or continue
            i = (i + 1) % (self.pool.max_set() + 1);
            if i == stop {
                 break
            }
        }
```

data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -384,23 +409,21 @@ impl NameState {
...
                 }
-                debug!("no output connection available");
-                None
+
+                let selected = random_index(eligible_count);
+                refs[index]
+                    .keys()
+                    .filter(|&&conn_id| conn_id != incoming_conn)
+                    .nth(selected)
+                    .copied()
```
micpapal: here I would use the same approach as before to remove the rnd number also in this case

data-plane/core/datapath/src/tables/subscription_table.rs
lpuskas: I've addressed the observations, thanks!

data-plane/core/datapath/src/tables/subscription_table.rs
lpuskas: this has been implicitly addressed

lpuskas:
## Data plane benchmark: `main` vs `perf/rnd`

**Inputs:** `benchmark-results-main.csv` vs `benchmark-results-rnd.csv`  
**Workload:** 500,000 messages per sender; payload sizes 8, 16, 64, 1024 bytes; sender counts 1, 2, 4, 8, 10.  
**Metric:** mean of `send_mps` and `recv_mps` over **10 runs** per configuration (identical `total_messages` per row).

### Mean throughput (M messages / s)

| senders | payload B | main send | rnd send | Δ send % | main recv | rnd recv | Δ recv % |
|--------:|----------:|----------:|---------:|---------:|----------:|---------:|---------:|
| 1 | 8 | 1.020 | 0.964 | −5.5 | 1.015 | 0.960 | −5.5 |
| 1 | 16 | 1.009 | 0.952 | −5.6 | 1.004 | 0.948 | −5.6 |
| 1 | 64 | 0.980 | 1.003 | +2.4 | 0.976 | 0.999 | +2.4 |
| 1 | 1024 | 0.998 | 0.979 | −1.9 | 0.993 | 0.974 | −1.9 |
| 2 | 8 | 1.375 | 1.311 | −4.7 | 1.370 | 1.307 | −4.6 |
| 2 | 16 | 1.416 | 1.390 | −1.8 | 1.411 | 1.385 | −1.8 |
| 2 | 64 | 1.438 | 1.304 | −9.3 | 1.433 | 1.301 | −9.3 |
| 2 | 1024 | 1.398 | 1.269 | −9.2 | 1.394 | 1.265 | −9.2 |
| 4 | 8 | 1.998 | 1.840 | −7.9 | 1.994 | 1.836 | −7.9 |
| 4 | 16 | 1.957 | 1.888 | −3.5 | 1.953 | 1.884 | −3.5 |
| 4 | 64 | 1.949 | 1.892 | −2.9 | 1.945 | 1.887 | −2.9 |
| 4 | 1024 | 2.044 | 1.890 | −7.5 | 2.040 | 1.886 | −7.5 |
| 8 | 8 | 1.188 | 1.713 | **+44.2** | 1.187 | 1.711 | **+44.2** |
| 8 | 16 | 1.184 | 1.783 | **+50.6** | 1.183 | 1.782 | **+50.6** |
| 8 | 64 | 1.200 | 1.661 | **+38.4** | 1.199 | 1.659 | **+38.4** |
| 8 | 1024 | 1.210 | 1.664 | **+37.5** | 1.209 | 1.663 | **+37.5** |
| 10 | 8 | 0.831 | 1.081 | **+30.1** | 0.830 | 1.080 | **+30.1** |
| 10 | 16 | 0.843 | 1.086 | **+28.9** | 0.843 | 1.086 | **+28.9** |
| 10 | 64 | 0.859 | 1.026 | **+19.4** | 0.859 | 1.025 | **+19.4** |
| 10 | 1024 | 0.840 | 1.101 | **+31.1** | 0.840 | 1.101 | **+31.1** |

**Unweighted mean across the 20 configurations:** send **+8.0%**, recv **+8.0%** (rnd vs main).

### Short summary

- **Low parallelism (1–4 senders):** rnd is **mixed**—often a few percent **slower** on send/recv (especially 2–4 senders / several payloads), with small **gains** in a few cells (e.g. 1 sender, 64 B).
- **High parallelism (8–10 senders):** rnd is **much higher** on both send and recv in these CSVs (**~+19% to +51%** depending on payload/senders), which dominates the **~+8%** unweighted average.
- **Caveat:** End-to-end MPS is noisy (OS scheduling, thermal, other load). Treat large swings—especially the **8-sender** block—as **needs confirmation** on a quiet, repeated run if you are deciding from this alone.

data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -179,38 +191,59 @@ impl Connections {
...
-                let val = self.index.iter().next().unwrap();
-                return Some(*val.0);
             }
+            debug_assert!(
+                self.pool
+                    .get(slot_idx)
+                    .is_some_and(|c| c.conn_id == conn_id)
+            );
+            LAST_USED_POOL_INDEX.with(|c| c.set(slot_idx));
+            return Some(slot_idx);
```
micpapal: why do not return the connection id directly here? this will avoid to do another get in the caller function

data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -179,38 +191,59 @@ impl Connections {
...
-        let mut stop = false;
-        let mut i = index;
-        while !stop {
-            let opt = self.pool.get(i);
-            if let Some(opt) = opt
-                && opt.conn_id != except_conn
+        let num_slots = self.pool.max_set() + 1;
+        let last_raw = LAST_USED_POOL_INDEX.with(|c| c.get());
+        // First slot to try: advance one past last pick, wrapped into [0, num_slots).
+        let next_slot = if last_raw == LAST_USED_POOL_INDEX_NONE {
```
micpapal: you can avoid this if by initialize LAST_USED_POOL_INDEX to 0. the interface with index 0 will be simply used as the last option.

data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -179,38 +191,59 @@ impl Connections {
...
+            if let Some(c) = self.pool.get(i)
+                && c.conn_id != except_conn
             {
-                return Some(opt.conn_id);
-            }
-            i = (i + 1) % (self.pool.max_set() + 1);
-            if i == index {
-                stop = true;
+                LAST_USED_POOL_INDEX.with(|c| c.set(i));
+                return Some(i);
```
micpapal: also here you have the connection id so you can return it immediately

data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -179,38 +191,59 @@ impl Connections {
...
             }
         }
+
+        LAST_USED_POOL_INDEX.with(|c| c.set(LAST_USED_POOL_INDEX_NONE));
```
micpapal: why do you need to reset the LAST_USED_POOL_INDEX here? I think there is no problem in do not update it and restart from there for the next match

data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -179,38 +191,59 @@ impl Connections {
...
         debug!("no output connection available");
         None
     }
 
+    fn get_one(&self, except_conn: u64) -> Option<u64> {
+        self.next_round_robin_pool_index(except_conn)
+            .and_then(|idx| self.pool.get(idx).map(|conn| conn.conn_id))
+    }
```
micpapal, line 228: if you return the connection directly from next_round_robin_pool_index this becomes simply
```
    fn get_one(&self, except_conn: u64) -> Option<u64> {
        self.next_round_robin_pool_index(except_conn)
    }
```

data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -344,65 +377,59 @@ impl NameState {
...
-            }
+        let rr_key = (id, side as u8);
+        let mut non_incoming_conn_ids: Vec<u64> = Vec::with_capacity(refs.len());
+        non_incoming_conn_ids.extend(refs.keys().copied().filter(|&c| c != incoming_conn));
+
+        if non_incoming_conn_ids.is_empty() {
+            LAST_USED_NON_INCOMING_CONN_RR_POS.with(|m| {
+                m.borrow_mut().insert(rr_key, LAST_USED_POOL_INDEX_NONE);
+            });
+            debug!("the only available connection cannot be used");
```
micpapal: If non_incoming_conn_ids there is no need to reset LAST_USED_NON_INCOMING_CONN_RR_POS. And the debug print can be

```
debug!("no output connection available");
```

data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -344,65 +377,59 @@ impl NameState {
...
+            return Some(non_incoming_conn_ids[0]);
+        }
+
+        non_incoming_conn_ids.sort_unstable();
+        let n = non_incoming_conn_ids.len();
+
+        let conn_id = LAST_USED_NON_INCOMING_CONN_RR_POS.with(|m| {
+            let mut map = m.borrow_mut();
+            let last_raw = *map.get(&rr_key).unwrap_or(&LAST_USED_POOL_INDEX_NONE);
+            let next_pos = if last_raw == LAST_USED_POOL_INDEX_NONE {
```
micpapal: same comment of line 221

data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -1,19 +1,31 @@
...
-use rand::Rng;
 use tracing::{debug, error, warn};
 
 use super::SubscriptionTable;
 use super::pool::Pool;
 use crate::errors::DataPathError;
 use crate::messages::Name;
 
+/// Sentinel for round-robin cursors: none chosen yet. Real indices are far below `usize::MAX`.
+const LAST_USED_POOL_INDEX_NONE: usize = usize::MAX;
```
micpapal: I don't think this constant is really needed, see comments below

data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -411,49 +438,45 @@ impl NameState {
...
+            return pool.get_all(incoming_conn);
         }
 
-        let val = self.ids.get(&id);
-        match val {
-            None => {
-                debug!(%id, "cannot find out connection, id does not exists");
-                None
+        let Some(name_refs) = self.ids.get(&id) else {
+            if pool.index.len() == 1 {
```
micpapal: Here in main we always return None

data-plane/core/datapath/src/tables/subscription_table.rs
lpuskas: it turns out that this change induces a logical shift into this method (the first index to consider - 0 - would be skipped) so i'd recommend keeping this condition here :)

lpuskas: that'd imply a change in the implemented cursor model; i'd stick with this if instead

msardara: We could initialize LAST_USED_POOL_INDEX  to `~0`, so that the first ++ will give 0 🤔

msardara: In that case we need to use [wrapping_add](https://doc.rust-lang.org/std/primitive.usize.html#method.wrapping_add) to make rust happy 😊

data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -1,19 +1,30 @@
...
 use crate::messages::Name;
 
+thread_local! {
+    /// Per-thread round-robin state for [`Connections::get_one`]: last **used** pool slot
+    /// (initialized to `0` before the first pick).
+    static LAST_USED_POOL_INDEX: Cell<usize> = const { Cell::new(0_usize) };
+    /// Per-name round-robin (RR): last **used connection position** (index in the sorted candidate list) for
+    /// [`NameState::get_one_connection`] (key = `(name id, 0 = local pool / 1 = remote pool)`).
+    /// Missing keys use `0_usize` like [`LAST_USED_POOL_INDEX`] before the first pick for that key.
+    static LAST_USED_CONN_POS: RefCell<HashMap<(u64, u8), usize>> =
```
msardara: I think here we can easily get away with a single u64, which is shared across all the tasks running in the thread.

We won't get a real round robin "per subscription", but that is not really needed here. We just need an integer that gets updated over time by the tasks and that will basically pseudo-randomly select a connection.

Let's also call it "NEXT_INDEX" for clarity.

We could also introduce a helper for advancing the index, like e.g.

```rust
fn next_index() -> usize {
    NEXT_INDEX.with(|c| {
        let val = c.get();
        c.set(val.wrapping_add(1));
        val
    })
}
```

data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -179,34 +190,47 @@ impl Connections {
         Ok(())
     }
 
+    /// Pick a connection id other than `except_conn` using round-robin over pool slots.
```
```suggestion
    /// Pick a connection id other than `except_conn` using pseudo-random selection.
```

data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -179,34 +190,47 @@ impl Connections {
...
+    ///
+    /// Single-connection case is O(1) on [`Self::index`]. Otherwise walks pool slots in ring order
+    /// starting from one slot after the thread-local last-used index, without allocating.
```
msardara: We can remove this comment.

data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -179,34 +190,47 @@ impl Connections {
...
-        let mut i = index;
-        while !stop {
-            let opt = self.pool.get(i);
-            if let Some(opt) = opt
-                && opt.conn_id != except_conn
+        let num_slots = self.pool.max_set() + 1;
+        // First slot to try: one past last used, wrapped into [0, num_slots).
+        // Read last-used slot once; compute where to start this scan (no write here).
+        let last_used = LAST_USED_POOL_INDEX.with(|c| c.get());
+        let next_slot = (last_used + 1) % num_slots;
```
```suggestion
        let start = next_index() % num_slots;
```

data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -179,34 +190,47 @@ impl Connections {
...
             {
-                return Some(opt.conn_id);
-            }
-            i = (i + 1) % (self.pool.max_set() + 1);
-            if i == index {
-                stop = true;
+                // Store the last **used** slot; next call will start at `(i + 1) % num_slots`.
+                LAST_USED_POOL_INDEX.with(|c| c.set(i));
+                return Some(c.conn_id);
             }
```
```suggestion
        for i in (start..num_slots).chain(0..start) {
            if let Some(c) = self.pool.get(i)
                && c.conn_id != except_conn
            {
                return Some(c.conn_id);
            }
        }
```

data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -179,34 +190,47 @@ impl Connections {
...
-            } else {
-                let val = self.index.iter().next().unwrap();
-                return Some(*val.0);
             }
+            debug_assert!(
+                self.pool
+                    .get(slot_idx)
+                    .is_some_and(|c| c.conn_id == conn_id)
+            );
+            LAST_USED_POOL_INDEX.with(|c| c.set(slot_idx));
```
```suggestion
            // Advance the counter even when there is only one connection,
            // so that subsequent calls to other pools stay well-distributed.
            next_index();
```

data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -344,65 +368,52 @@ impl NameState {
...
+        let conn_id = LAST_USED_CONN_POS.with(|m| {
+            let mut map = m.borrow_mut();
+            let last_pos = *map.get(&rr_key).unwrap_or(&0_usize);
+            let next_pos = (last_pos + 1) % n;
+            let picked = non_incoming_conn_ids[next_pos];
+            map.insert(rr_key, next_pos);
+            picked
+        });
+
+        Some(conn_id)
```
msardara: I am not toally sure to get this part. We should just get one connection out of refs right?

Why don't we just create a helper, say pick_one, which we call both here and in the get_one function?

The pick_one function could be something like

```rust
fn pick_one(n: usize, except: u64, get: impl Fn(usize) -> Option<u64>) -> Option<u64> {
    if n == 0 {
        return None;
    }
    let start = next_index() % n;
    for i in (start..n).chain(0..start) {
        if let Some(val) = get(i) {
            if val != except {
                return Some(val);
            }
        }
    }
    None
}
```

And then for get_one you would do:

```rust
fn get_one(&self, except_conn: u64) -> Option<u64> {
        let num_slots = self.pool.max_set() + 1;
        let result = pick_one(num_slots, except_conn, |i| {
            self.pool.get(i).map(|c| c.conn_id)
        });
        if result.is_none() {
            debug!("no output connection available");
        }
        result
    }
```

while here you can modify as follows:

```rust
let result = pick_one(refs.len(), incoming_conn, |i| {
            refs.keys().nth(i).copied()
        });
```

data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -411,49 +422,42 @@ impl NameState {
...
+                .copied()
+                .find(|&conn_id| conn_id != incoming_conn)
+            else {
+                debug!("the only available connection cannot be used");
+                return None;
+            };
+            return Some(vec![conn_id]);
+        }
+
+        let mut out = Vec::new();
```
msardara: Here we can already preallocate the vector as we know its size

msardara (changes requested):

data-plane/core/datapath/src/tables/subscription_table.rs
micpapal: agree, this can be simplified

Tehsmash:
Merged into: https://github.com/agntcy/slim/pull/1496

PR #1426, opened 2026-03-31
fix(bindings/go): use ChannelInterface and ServerInterface in generated stubs

### LINKED ISSUE #1419: protoc-gen-slimrpc-go: use ChannelInterface instead of *Channel in generated client stubs
**Version**
`protoc-gen-slimrpc-go` (current version, as used by `slim-a2a-go`)

**Platform**
Go (all platforms)

**Description**

We generated Go client stubs for an A2A service using `protoc-gen-slimrpc-go`.

I tried this code:

```go
// Generated by protoc-gen-slimrpc-go
func NewA2AServiceClient(channel *slim_bindings.Channel) A2AServiceClient {
    return &a2AServiceClient{channel: channel}
}
```

I expected to see this happen: the constructor accepts `slim_bindings.ChannelInterface` so that any implementation can be injected.

Instead, this happened: the constructor is bound to the concrete `*slim_bindings.Channel` type, making the interface impossible to use for dependency injection.

**Impact**

In [`slim-a2a-go`](https://github.com/agntcy/slim-a2a-go), a `Transport` type wraps the generated `A2AServiceClient`. Because `NewA2AServiceClient` takes `*Channel`, it is impossible to inject a mock channel in unit tests — the only option is to stand up a real in-process SLIM server for every client-level test. Using `slim_bindings.ChannelInterface` instead would allow lightweight mocking and make Transport tests fully isolated from the SLIM server.

Tehsmash:
# Description

`protoc-gen-slimrpc-go` previously generated client stubs that accepted the concrete `*slim_bindings.Channel` type in constructors, and server registration functions that accepted `*slim_bindings.Server`. This made it impossible to inject test doubles without standing up a real SLIM server.

The UniFFI-generated `slim_bindings.go` already defines `ChannelInterface` and `ServerInterface` — this change updates the code generator to use those interfaces everywhere the concrete pointer types were used in generated client/server registration code.

**Changes:**
- `New{{Service}}Client(channel *Channel)` → `New{{Service}}Client(channel ChannelInterface)` — allows mock channel injection in client tests
- `Register{{Service}}Server(server *Server, ...)` → `Register{{Service}}Server(server ServerInterface, ...)` — allows mock server injection for registration tests

Because `*Channel` and `*Server` already implement the respective interfaces, all existing callers continue to compile and work without any modifications.

Fixes #1419

## Type of Change

- [x] Bugfix

## Checklist

- [x] I have read the contributing guidelines

### DIFF
--- a/data-plane/slimrpc-compiler/src/golang.rs
+++ b/data-plane/slimrpc-compiler/src/golang.rs
@@ -40,11 +40,11 @@ type {{SERVICE_NAME}}Client interface {
 }
 
 type {{SERVICE_NAME}}ClientImpl struct {
-	channel *slim_bindings.Channel
+	channel slim_bindings.ChannelInterface
 }
 
 // New{{SERVICE_NAME}}Client creates a new {{SERVICE_NAME}} client.
-func New{{SERVICE_NAME}}Client(channel *slim_bindings.Channel) {{SERVICE_NAME}}Client {
+func New{{SERVICE_NAME}}Client(channel slim_bindings.ChannelInterface) {{SERVICE_NAME}}Client {
 	return &{{SERVICE_NAME}}ClientImpl{
 		channel: channel,
 	}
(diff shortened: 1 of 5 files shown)


msardara:
@Tehsmash should we release this as part of v1.3.0?

data-plane/slimrpc-compiler/src/golang.rs
msardara, line 1: We should also update the `GroupClient` part I think.

For the SLIM 2.0.0 we should probably unify these 2 interfaces into a single Client.

PR #1457, opened 2026-04-02
feat(multicluster-demo): add a2acli skill with SLIM RPC transport

Tehsmash:
# Description

Adds `a2acli`, a CLI tool and Claude Code skill for interacting with A2A protocol agents. It supports both standard HTTP transports (JSON-RPC, REST) and SLIM RPC for agents deployed behind a SLIM messaging node — the pattern used by the k8s troubleshooting agent in this demo.

**Commands:**
- `list` — list all agents in the local store
- `get-card` — print an agent's full AgentCard JSON
- `send-message` — send a message, poll until done by default; `--return-immediately` asks the agent to return a Task straight away
- `get-task` — retrieve a task by ID

**SLIM RPC transport:**
- Registers both A2A v0.3 and v1.0 SLIM RPC transports via `slim-a2a-go`
- Transport selected automatically from the agent's `protocolVersion`
- A demo AgentCard for `k8s_troubleshooting_agent` is included in `.a2aagents/`

**Configuration:**
- Credentials and SLIM settings load from `.a2acli.yaml` (CWD or `~/.a2acli.yaml`); flags always take precedence
- An annotated example config is provided at `.a2acli.yaml.example`

**Packaging:**
- `Taskfile.yaml` with `setup`, `build`, `build-skill`, `package-skill`, and `install` targets
- `setup` downloads the `slim-bindings-go` native library (CGO); runs automatically as a dep of build targets
- `package-skill` produces `bin/a2acli.zip` for skill distribution

## Type of Change

- [x] New Feature

## Checklist


multicluster-demo/a2acli-skill/cmd/get_card.go
```diff
@@ -0,0 +1,56 @@
+// Copyright 2025 Sam Betts
```
micpapal: you left your own copyright. is it done on purpose?
it is in all the go files

multicluster-demo/a2acli-skill/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
micpapal: also the license file should be removed

micpapal (approved):

PR #1462, opened 2026-04-03
feat(mulitcluster-demo): add MCP clien for k8s troubleshooting agent

micpapal:
# Description
Integrates the MCP client to enable the Kubernetes troubleshooting agent to query cluster resources via an MCP server proxy over SLIM.

### Changes

**New MCP Client (`mcp_client.py`)**
- `K8sMCPClient` class that connects to an MCP proxy via SLIM
- `list_available_tools()`: discovers available MCP tools from the server
- `call_tool()`: executes MCP tool calls and returns results

**Configuration**
- `.env.template`: Added `MCP_PROXY_NAMESPACE`, `MCP_PROXY_GROUP`, `MCP_PROXY_NAME` settings

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/multicluster-demo/k8s_troubleshooting_agent/k8s_troubleshooting_agent/tools.py
+++ b/multicluster-demo/k8s_troubleshooting_agent/k8s_troubleshooting_agent/tools.py
@@ -1,73 +1,156 @@
-"""Tools for the Kubernetes troubleshooting agent."""
-
-import json
 import logging
-from typing import Annotated
+from typing import Any, Optional
 
-from pydantic import Field
+from google.adk.tools.base_tool import BaseTool
+from google.adk.tools.base_toolset import BaseToolset
+from google.adk.tools.tool_context import ToolContext
+from google.adk.agents.readonly_context import ReadonlyContext
+from google.genai.types import FunctionDeclaration
+from mcp.types import Tool as McpBaseTool
+from typing_extensions import override
 
 from k8s_troubleshooting_agent.mcp_client import K8sMCPClient
 
 logger = logging.getLogger(__name__)
 
-# Global MCP client instance (will be initialized in main.py)
-mcp_client: K8sMCPClient | None = None
+class SLIMMcpTool(BaseTool):
+    """Wraps an MCP tool to work with ADK framework using SLIM MCP client."""
 
+    def __init__(
+        self,
+        mcp_tool: McpBaseTool,
+        mcp_client: K8sMCPClient,
+    ):
+        """Initialize a SLIM MCP tool.
 
-def set_mcp_client(client: K8sMCPClient) -> None:
-    """Set the global MCP client instance."""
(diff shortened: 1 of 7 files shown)


multicluster-demo/k8s_troubleshooting_agent/k8s_troubleshooting_agent/tools.py
```diff
@@ -0,0 +1,73 @@
...
+        
+    logger.debug(f"Calling MCP tool: {tool_name} with args: {arguments}")
+
+    try:
+        result = await mcp_client.call_tool(tool_name, arguments)
+        logger.debug(f"MCP tool {tool_name} returned successfully")
+        return result
+    except Exception as e:
+        logger.error(f"Error calling MCP tool {tool_name}: {e}", exc_info=True)
+        return f"Error calling MCP tool {tool_name}: {e}"
```
Tehsmash: I think we should follow the ADK design for MCP tools which uses an "McpToolSet" which extends "BaseToolSet", this will handle all the tool loading etc for the agent preventing us needing to add the instructions to check for tools as part of the prompt.

https://github.com/google/adk-python/blob/bbad9ec64ce1617bc45148de97e6246752845b98/src/google/adk/tools/mcp_tool/mcp_toolset.py#L66

Annoyingly we can't override the MCPClient for the McpToolSet and its not even that clean to do it through overriding... the Go Interface is much cleaner. Instead I think we should create our own McpToolSet like:

```
class SLIMMcpTool(BaseTool):
	
	def __init__(self, mcp_tool, mcpClient):
		...

	@override
  	def _get_declaration(self) -> FunctionDeclaration:
    """Gets the function declaration for the tool.

    Returns:
        FunctionDeclaration: The Gemini function declaration for the tool.
    """
    input_schema = self._mcp_tool.inputSchema
    output_schema = self._mcp_tool.outputSchema
    function_decl = FunctionDeclaration(
          name=self.name,
          description=self.description,
          parameters_json_schema=input_schema,
          response_json_schema=output_schema,
      )
    return function_decl

	async def run_async(...):
		# clone code from: https://github.com/google/adk-python/blob/main/src/google/adk/tools/mcp_tool/mcp_tool.py#L276-L397

class SLIMMcpToolSet(BaseToolSet):

	def __init__(self, mcpClient):
		
	def get_tools(self):
		# clone code from: https://github.com/google/adk-python/blob/main/src/google/adk/tools/mcp_tool/mcp_toolset.py#L305-L350
```

We can create a pass the mcpClient in and use that instead of the "session" manager thing they have in the original.

Tehsmash (changes requested):

Tehsmash (approved):

PR #1496, opened 2026-04-10
refactor(datapath): replace Pool bitmap/vec with three-vec design

### LINKED ISSUE #1382: RwLock on subscription table per message
**File:** `core/datapath/src/tables/subscription_table.rs` — `match_one()` / `match_all()`

**Problem:**
Every `process_publish` acquires a `parking_lot::RwLock` read lock on
`self.table` (the subscription HashMap).  While `parking_lot` read locks are
low-cost under no contention, they become a serialization point when:
- Subscription churn (add/remove) takes a write lock on the same table.
- High message rates cause many concurrent readers to contend with any writer.

The `match_all` path additionally builds a `Vec<u64>` per call.

**Mitigation plan:**
1. **Replace `RwLock<HashMap>` with a lock-free concurrent map**
   (e.g., `dashmap` or `flurry`).  Reads become wait-free; writes are
   sharded.
2. If lock-free maps are too invasive, **use `parking_lot::RwLock` with
   `read_recursive()`** and ensure writers are batched (coalesce rapid
   subscribe/unsubscribe bursts).
3. For `match_all`, pre-allocate the output `Vec` with a small inline array
   (`SmallVec<[u64; 4]>`) to avoid heap allocation for common low-fanout
   cases.

Tehsmash:
# Description

\`Pool<T>\` previously stored items in a \`Vec<Option<T>>\` with a \`BitVec\`
tracking occupied slots.  This created an inconsistency: \`iter()\` skipped
empty slots (logical indexing) while \`get(n)\` indexed raw storage positions
(physical indexing).  A stale ID could also silently resolve to a *different*
element after a remove+insert cycle because physical slots were reused.

This replaces the implementation with a purpose-built three-Vec design:

\`\`\`
pool:           Vec<Option<PoolEntry<T>>>  // sparse storage; Vec index == ID
active_indexes: Vec<usize>                // dense list of live indices
free_slots:     Vec<usize>                // recycled slot stack (free-list)
\`\`\`

- **O(1) \`get()\`** — direct Vec index, no hash computation.
- **Dense \`iter()\`** — \`active_indexes\` walk, no empty slots visited; \`iter()\` and \`get(id)\` are now semantically consistent.
- **O(1) amortised \`insert()\`** — free-list stack avoids the O(n²) bitmap scan.
- **O(1) \`remove()\`** — each entry carries an \`active_pos\` back-pointer into \`active_indexes\`, so removal jumps directly to the right position instead of scanning; \`swap_remove\` then fixes up the moved entry in a single indexed write.
- \`insert_at(element, id)\` still supported for state-sync scenarios.
- Adds \`iter_with_ids()\` for callers that need \`(id, &T)\` pairs.
- Removes the \`bit-vec\` crate dependency from the workspace entirely.
- All callers updated to use \`u64\` IDs directly, removing the \`as usize\` / \`as u64\` casts that papered over the semantic mismatch.

**Note:** IDs are reused after \`remove\` + \`insert\`.  Callers hold IDs only
for the lifetime of the element and call \`remove\` before any reuse, so this
is safe.

## Benchmarks

Criterion benchmarks vs \`main\` baseline (1024-element pool, Apple M-series):

| Benchmark | \`main\` | this PR | Change |
|---|---|---|---|
| pool lookup (1024 × \`get\` by ID) | 431 ns | 453 ns | within noise |
| pool insert (1024 inserts, pre-allocated) | 240 µs | 2.25 µs | **~107× faster** |
| pool grow (1024 inserts, start cap 16) | 240 µs | 3.36 µs | **~71× faster** |
| pool remove (1024 removes) | 243 µs | 4.91 µs | **~50× faster** |
| pool remove+insert cycle (1024 pairs) | 479 µs | 6.68 µs | **~72× faster** |
| pool next\_id (round-robin, 8-element pool) | — | 2.20 ns | new |
| pool next\_val (round-robin, 8-element pool) | — | 2.13 ns | new |

All significant results have p = 0.00.  The \`remove\` improvement breaks down across two commits: the three-Vec redesign (linear scan, ~2.9× vs main), then the \`active_pos\` back-pointer (O(1) scan-free, ~18× vs the three-Vec baseline — ~50× total vs main).

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/core/datapath/benches/pool_benchmark.rs
+++ b/data-plane/core/datapath/benches/pool_benchmark.rs
@@ -96,9 +96,7 @@ fn bench_next_id(c: &mut Criterion) {
         pool.insert(i);
     }
 
-    c.bench_function("pool next_id", |b| {
-        b.iter(|| black_box(pool.next_id()))
-    });
+    c.bench_function("pool next_id", |b| b.iter(|| black_box(pool.next_id())));
 }
 
 fn bench_next_val(c: &mut Criterion) {
@@ -107,9 +105,7 @@ fn bench_next_val(c: &mut Criterion) {
         pool.insert(i);
     }
 
-    c.bench_function("pool next_val", |b| {
-        b.iter(|| black_box(pool.next_val()))
-    });
+    c.bench_function("pool next_val", |b| b.iter(|| black_box(pool.next_val())));
 }
 
 criterion_group!(
--- a/data-plane/core/datapath/benches/pool_benchmark.rs
+++ b/data-plane/core/datapath/benches/pool_benchmark.rs
@@ -67,7 +67,7 @@ fn bench_remove(c: &mut Criterion) {
     c.bench_function("pool remove", |b| {
         b.iter(|| {
             let mut pool = Pool::with_capacity(1024);
-            let ids: Vec<usize> = (0..1024).map(|i| pool.insert(i)).collect();
+            let ids: Vec<u64> = (0..1024).map(|i| pool.insert(i)).collect();
             for id in &ids {
                 black_box(pool.remove(*id));
             }
@@ -80,7 +80,7 @@ fn bench_remove_insert_cycle(c: &mut Criterion) {
     c.bench_function("pool remove+insert cycle", |b| {
         b.iter(|| {
             let mut pool = Pool::with_capacity(1024);
-            let ids: Vec<usize> = (0..1024).map(|i| pool.insert(i)).collect();
+            let ids: Vec<u64> = (0..1024).map(|i| pool.insert(i)).collect();
             for (i, &id) in ids.iter().enumerate() {
                 pool.remove(id);
                 black_box(pool.insert(i as i32));
--- a/data-plane/core/datapath/benches/pool_benchmark.rs
+++ b/data-plane/core/datapath/benches/pool_benchmark.rs
@@ -1,31 +1,15 @@
-use bit_vec::BitVec;
 use criterion::{Criterion, black_box, criterion_group, criterion_main};
 use slim_datapath::tables::pool::Pool;
 use std::mem::MaybeUninit;
 
 fn bench_lookup(c: &mut Criterion) {
     let mut pool = Pool::with_capacity(1024);
-    for i in 0..1024 {
-        pool.insert(i);
-    }
+    let ids: Vec<u64> = (0..1024i32).map(|i| pool.insert(i)).collect();
 
     c.bench_function("pool lookup", |b| {
         b.iter(|| {
-            for i in 0..1024 {
-                black_box(pool.get(i));
-            }
-        })
-    });
-}
-
-fn bench_bitvec_lookup(c: &mut Criterion) {
-    let size = 1024;
-    let bitvec = BitVec::from_elem(size, true);
-
-    c.bench_function("bitvec lookup", |b| {
-        b.iter(|| {
-            for i in 0..size {
-                black_box(bitvec.get(i));
+            for &id in &ids {
+                black_box(pool.get(id));
             }
         })
     });
@@ -108,7 +92,6 @@ fn bench_remove_insert_cycle(c: &mut Criterion) {
 criterion_group!(
     benches,
     bench_lookup,
-    bench_bitvec_lookup,
     bench_assume_init_ref,
     bench_insert,
     bench_grow,
--- a/data-plane/core/datapath/benches/pool_benchmark.rs
+++ b/data-plane/core/datapath/benches/pool_benchmark.rs
@@ -89,6 +89,29 @@ fn bench_remove_insert_cycle(c: &mut Criterion) {
     });
 }
 
+fn bench_next_id(c: &mut Criterion) {
+    // Representative pool size: a handful of outbound connections per name.
+    let mut pool = Pool::with_capacity(8);
+    for i in 0..8i32 {
+        pool.insert(i);
+    }
+
+    c.bench_function("pool next_id", |b| {
+        b.iter(|| black_box(pool.next_id()))
+    });
+}
+
+fn bench_next_val(c: &mut Criterion) {
+    let mut pool = Pool::with_capacity(8);
+    for i in 0..8i32 {
+        pool.insert(i);
+    }
+
+    c.bench_function("pool next_val", |b| {
+        b.iter(|| black_box(pool.next_val()))
+    });
+}
+
 criterion_group!(
     benches,
     bench_lookup,
@@ -98,6 +121,8 @@ criterion_group!(
     bench_capacity,
     bench_remove,
     bench_remove_insert_cycle,
+    bench_next_id,
+    bench_next_val,
 );
 
 criterion_main!(benches);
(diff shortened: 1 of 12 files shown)


lpuskas (approved):
lgtm 👍

data-plane/core/datapath/src/tables/pool.rs
```diff
@@ -1,374 +1,246 @@
...
-
-        // Store the new element in the pool
-        self.pool[index] = Some(element);
+        self.pool[idx] = None;
 
-        // If the index is greater than the max_set, update max_set
-        if index > self.max_set {
-            self.max_set = index;
+        // Remove from active_indexes (swap_remove is O(1) after the scan).
+        if let Some(pos) = self.active_indexes.iter().position(|&i| i == idx) {
```
msardara: Not a big concerns as remove operations are not in the datapath, but here removing an element from the pool has linear cost.

Maybe we could keep a reverse map (as vec) mapping pool index -> position in self.active_indexes? Like `Vec<Option<usize>>`

data-plane/core/datapath/src/tables/pool.rs
```diff
@@ -1,374 +1,246 @@
...
-            debug_assert!(self.bitmap.capacity() >= self.capacity);
+    /// Insert `element` at a specific `id`.
+    ///
+    /// Grows `pool` if `id` is beyond the current length.  If the slot was
+    /// empty the index is added to `active_indexes`; if it was already
+    /// occupied the element is replaced (length unchanged).
+    ///
+    /// Returns `true` always; the bool return mirrors the old API for
+    /// compatibility with call-sites that check the return value.
+    pub fn insert_at(&mut self, element: T, id: u64) -> bool {
```
msardara: I think we can break the API if we change the behavior to extending the pool if id is > self.pool.len() . Returning bool is a bit confusing as this operation now can never fail (OOM errors excluded)

data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -190,21 +189,19 @@ impl Connections {
             }
         }
 
-        // we need to iterate and find a value starting from a random point in the pool
+        // Pick a random starting point in the (dense) pool and walk forward,
+        // looking for a connection that isn't except_conn.
+        let conn_ids: Vec<u64> = self.pool.iter().map(|c| c.conn_id).collect();
```
msardara: This allocates a new Vec<u64> at each call. I think we should not change this as this function is called in the data path.

data-plane/core/datapath/src/tables/pool.rs
```diff
@@ -1,374 +1,246 @@
...
+        let was_empty = self.pool[idx].is_none();
+        self.pool[idx] = Some(element);
 
-            index
-        } else {
-            // This should never happen
-            panic!("pool is full");
+        if was_empty {
+            self.active_indexes.push(idx);
         }
```
msardara: We should also handle the case where idx is in the free slots. Currently we could have an entry for the same index in free slots and in active_indexes I think.

data-plane/core/datapath/benches/pool_benchmark.rs
msardara, line 1: We should also add a benchmark for remove, and iter(), so to cover all the operations. This might be part of another PR though.

The old remove was probably faster as it was just updating the bitmap, but I think the operations that matter are iter() and get()

msardara (changes requested):

data-plane/core/datapath/src/tables/pool.rs
Tehsmash: free_slots is handled lazily in this case, "insert(..)" will pop a free slot, then check if pool[slot] is not None, if it is populated it will skip and pop another free slot. insert_at appears to only be used as part of reconnect so its not a common operation.

msardara: Perfect, not an issue then 🔝

data-plane/core/datapath/src/tables/pool.rs
```diff
@@ -1,376 +1,276 @@
...
+    /// write lock.  If elements are removed between calls the modulo
+    /// wraps the cursor back into range automatically.
+    pub fn next_id(&self) -> Option<u64> {
+        let n = self.active_indexes.len();
+        if n == 0 {
+            return None;
+        }
+        let pos = self.cursor.fetch_add(1, Ordering::Relaxed) % n;
+        Some(self.active_indexes[pos] as u64)
+    }
```
micpapal, line 207: does it make sense here to add next_val() as well? instead of return the next id of the pool returns immediately the value. This cloud be used at line 200 in the subscription table.

data-plane/core/datapath/src/tables/pool.rs
micpapal: yes this would make also the remove O(1). just need to keep track of the position in this vector

data-plane/core/datapath/src/tables/pool.rs
```diff
@@ -1,376 +1,276 @@
...
+    /// Stack of recycled indices available for reuse by the next `insert`.
+    free_slots: Vec<usize>,
 
-    /// index of the bit sit with max pos
-    max_set: usize,
+    /// Cursor for the persistent round-robin iterator (`next_id`).
+    /// Stored as an `AtomicUsize` so that `next_id` can advance it through
+    /// a shared `&self` reference — callers that hold only a read lock on
+    /// an outer container are not forced to upgrade to a write lock.
+    cursor: AtomicUsize,
```
msardara, line 44: I guess atomic here allows to round robin over the pool without getting a write lock for the pool. I like the solution but I am concerned by the cache coherence mechanism that triggers when updating the atomic. The thread local variable on the other side allowed to use a not-atomic cursor.

However, it is possible this is not an issue and they perform in the same way. cargo bench will tell us the truth :D

Tehsmash: +1, I'm considering using @lpuskas's thread local cursor to avoid the need for an Atomic, the downside is that the roundrobin logic would no longer be global it would be thread local

Tehsmash: If I can get consistent benchmarks then I can hopefully confirm

data-plane/core/datapath/src/tables/pool.rs
Tehsmash: Handled this with a back reference stored in self.pool along with the data

data-plane/core/datapath/src/tables/pool.rs
Tehsmash: I did a bit of analysis of using a thread local cursor, and there was a couple of interesting outcomes but the main insight was that the Atomic is about as performant as a Cell, and having the cursor per pool ensures that we have consistent round robin even with multiple pool instances, but having a global thread local cursor means that multiple pool instances will influence each other indirectly.

msardara (approved):
LGTM ✅

PR #1577, opened 2026-04-27
perf(datapath): eliminate allocations from publish routing hot path

### LINKED ISSUE #1386: `Name` allocation overhead
**File:** `core/datapath/src/messages/encoder.rs` — `Name::from_strings()`

**Problem:**
Each `Name` creation allocates 3 `String`s + a `Box<[String; 3]>` + 3 hash
computations.  Names are created when parsing messages and during subscription
matching.

**Mitigation plan:**
1. Cache `Name` objects for hot destinations (LRU or interning table).
2. Accept `&str` and hash on the fly without allocating owned `String`s when
   only matching is needed.

Tehsmash:
# Description

The publish routing hot path allocates unnecessarily on every inbound publish message. `SlimHeader::get_dst()` converts the proto `Name` into an internal `Name`, allocating `Arc<[String; 3]>` and cloning 3 strings from the wire message. The routing path (subscription table matching) never uses these strings — it only needs the four `u64` hash components already present as `EncodedName` in the proto.

This PR threads a zero-allocation `EncodedName` (a `Copy` struct of 4 `u64`s) through the entire publish routing path, eliminating all heap allocation from the hot path.

**Benchmark results (Apple M-series, release profile):**

| Benchmark | Time | Change |
|---|---|---|
| `routing_via_name` (before) | ~108 ns | baseline |
| `routing_via_encoded_name` (after) | ~38 ns | **~65% faster** |

Closes #1386

## Changes

**Phase 1 — Benchmark baseline (separate commit)**
- Add criterion benchmark suite covering `Name` construction and the full publish routing loop (`routing_via_name` baseline)

**Phase 2 — Code changes**
- Add `SlimHeader::get_encoded_dst()` / `get_encoded_source()` returning `EncodedName` (Copy, zero allocation)
- Add `From<&Name> for EncodedName` conversion
- Replace `SubscriptionTable::match_one` / `match_all` to accept `&EncodedName` instead of `&Name`
- Replace `InternalName([u64; 4])` with `SubscriptionName([u64; 3])`, eliminating unsafe `transmute` tricks and the redundant 4th component (the id is never part of the subscription group key)
- Replace `representative_name: Name` in `NameState` with `representative_strings: [String; 3]` — the human-readable name this `NameState` represents in the subscription table — reconstructing a `Name` only in the cold `for_each` path
- Simplify `process_publish` and `match_and_forward_msg`; retain `get_dst()` call in the debug log for human-readable output
- Add `DataPathError::NoMatchEncoded` for the allocation-free error path

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


data-plane/core/datapath/src/messages/utils.rs
```diff
@@ -264,13 +264,21 @@ impl SlimHeader {
         }
     }
 
+    pub fn get_encoded_source(&self) -> EncodedName {
+        self.source.as_ref().unwrap().name.unwrap()
```
msardara, line 272: I think in the validate function we make sure `source` and `destination` are not None, but we do not check the EncodedName part, so a malformed message might make the data plane panic. We should add the validation in [validate_routed_header](https://github.com/agntcy/slim/blob/main/data-plane/core/datapath/src/messages/utils.rs#L548)

data-plane/core/datapath/src/messages/utils.rs
```diff
@@ -264,13 +264,21 @@ impl SlimHeader {
...
+
     pub fn get_dst(&self) -> Name {
         match &self.destination {
             Some(destination) => Name::from(destination),
             None => panic!("destination not found"),
         }
     }
 
+    pub fn get_encoded_dst(&self) -> EncodedName {
+        self.destination.as_ref().unwrap().name.unwrap()
```
msardara, line 283: Same comment as get_encoded_source

msardara (changes requested):
L almost GTM, just one small nit to fix ✅

msardara (approved):

PR #1581, opened 2026-04-28
refactor: control plane in rust

msardara:
# Description

Move control plane implementation to rust.

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


data-plane/control-plane/src/db/model.rs
```diff
@@ -0,0 +1,612 @@
...
+/// `client_config` is a raw JSON value that mirrors the data-plane's
+/// `ClientConfig` structure so it can be sent verbatim as connection
+/// config data.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ConnectionDetails {
+    pub endpoint: String,
+    pub external_endpoint: Option<String>,
+    pub trust_domain: Option<String>,
+    pub mtls_required: bool,
+    pub client_config: serde_json::Value,
```
micpapal: here client client_config contains already info about mtls_required and the spire trust_domain. do we still need them?

data-plane/control-plane/src/db/model.rs
```diff
@@ -0,0 +1,612 @@
...
+            2 => RouteStatus::Failed,
+            _ => RouteStatus::Pending,
+        })
+    }
+}
+
+#[derive(Debug, Clone, Queryable, Selectable, Identifiable, Insertable)]
+#[diesel(table_name = routes)]
+pub struct Route {
+    pub id: i64,
```
micpapal: is there a reason to use i64 as id? it is required by the DB? otherwise we can simple associate an uuid and store it as string, this will remove the need of pub fn unique_id

data-plane/control-plane/src/db/model.rs
```diff
@@ -0,0 +1,612 @@
...
+            2 => LinkStatus::Failed,
+            _ => LinkStatus::Pending,
+        })
+    }
+}
+
+#[derive(Debug, Clone, Queryable, Selectable, Insertable)]
+#[diesel(table_name = links, primary_key(link_id, source_node_id, dest_node_id, dest_endpoint))]
+pub struct Link {
+    pub link_id: String,
```
micpapal, line 377: here the id is a string indeed

data-plane/control-plane/src/db/model.rs
```diff
@@ -0,0 +1,612 @@
...
+            self.source_node_id, self.dest_node_id, self.dest_endpoint, self.link_id, self.status,
+        )
+    }
+}
+
+// ─── Channel ──────────────────────────────────────────────────────────────────
+
+#[derive(Debug, Clone, Queryable, Selectable, Identifiable, Insertable)]
+#[diesel(table_name = channels)]
+pub struct Channel {
```
micpapal: if we use the channel manager app  this part can be removed. however we need to implement something similar in the channel manager app to keep persistent state of the groups. But this should be part of another PR

data-plane/control-plane/src/db/inmemory.rs
```diff
@@ -0,0 +1,1268 @@
...
+    }
+}
+
+// ── InMemoryDb ─────────────────────────────────────────────────────────────────
+
+pub struct InMemoryDb {
+    nodes: RwLock<HashMap<String, Node>>,
+    routes: RwLock<RouteStore>,
+    links: RwLock<LinkStore>,
+    channels: RwLock<HashMap<String, Channel>>,
```
micpapal: Same as the comment for the channel I added before. this may go away

data-plane/control-plane/src/db/inmemory.rs
```diff
@@ -0,0 +1,1268 @@
...
+        Ok(())
+    }
+
+    async fn mark_route_deleted(&self, route_id: i64) -> Result<()> {
+        let mut store = self.routes.write();
+        let route = store
+            .primary
+            .get_mut(&route_id)
+            .ok_or(Error::RouteNotFound { id: route_id })?;
+        route.deleted = true;
```
micpapal: instead to use a bool this can be another state of the route. and you can have only one fn instead of tree where the state is passed as parameter

data-plane/control-plane/src/db/model.rs
```diff
@@ -0,0 +1,612 @@
...
+    pub conn_config_data: String,
+    pub status: LinkStatus,
+    pub status_msg: String,
+    pub deleted: bool,
+    #[diesel(deserialize_as = DbTimestamp, serialize_as = DbTimestamp)]
+    pub last_updated: SystemTime,
+}
+
+impl Link {
+    pub fn storage_key(&self) -> String {
```
micpapal, line 392: is the storage key needed? or link_id is already enough?

data-plane/control-plane/src/db/inmemory.rs
```diff
@@ -0,0 +1,1268 @@
...
+                    Some(prev) if link.last_updated > prev.last_updated => latest = Some(link),
+                    _ => {}
+                }
+            }
+        }
+        latest.cloned()
+    }
+
+    async fn find_link_between_nodes(
+        &self,
```
micpapal, line 627: I thinks this function can be a wrap of get_link_for_source_and_endpoint. Should be something like
```
{
  let if Some(link) = get_link_for_source_and_endpoint(node1, node2) {
    return link
  }  
  get_link_for_source_and_endpoint(node2, node1)
}
```

data-plane/control-plane/src/services/routes/link_reconciler.rs
```diff
@@ -0,0 +1,505 @@
...
+    let mut entries = Vec::new();
+    for chunk in chunks {
+        if let Some(Payload::ConnectionListResponse(r)) = chunk.payload {
+            entries.extend(r.entries);
+        }
+    }
+    Ok(entries)
+}
+
+async fn handle_request(
```
micpapal: This fn can be part of LinkReconciler so that you don't need to pass all the self.* arguments

data-plane/control-plane/src/services/routes/link_reconciler.rs
```diff
@@ -0,0 +1,505 @@
...
+
+            self.queue.done(&node_id);
+        }
+
+        tracing::info!("link reconciler: shutting down");
+    }
+}
+
+/// Query the node's active connections and return all entries.
+async fn query_node_connections(
```
micpapal: this fn can be part of the LinkReconciler

data-plane/control-plane/src/services/routes/link_reconciler.rs
```diff
@@ -0,0 +1,505 @@
...
+    node_id: &str,
+    route_queue: &WorkQueue<String>,
+    enable_orphan_detection: bool,
+) -> Result<()> {
+    if cmd_handler.get_connection_status(node_id).await != NodeStatus::Connected {
+        tracing::info!("link reconciler: node {node_id} not connected, skipping");
+        return Ok(());
+    }
+
+    let links = db.get_links_for_node(node_id).await;
```
micpapal: every time we scan the links list we always check if if link.source_node_id == node_id. maybe we can add a function in the db like get_links_for_src_node(node_id) that filter out all the links where node_id is a destination

data-plane/control-plane/src/services/routes/link_reconciler.rs
```diff
@@ -0,0 +1,505 @@
...
+    Ok(entries)
+}
+
+async fn handle_request(
+    db: &SharedDb,
+    cmd_handler: &DefaultNodeCommandHandler,
+    node_id: &str,
+    route_queue: &WorkQueue<String>,
+    enable_orphan_detection: bool,
+) -> Result<()> {
```
micpapal: overall the function seems correct but is way too big. It can be split in multiple smaller function and it may take advantage from other helper function on the db search to filter out unwanted entries as the same filtering seems to be done multiple times.

data-plane/control-plane/src/services/routes/route_reconciler.rs
```diff
@@ -0,0 +1,637 @@
...
+            self.queue.done(&node_id);
+        }
+
+        tracing::info!("route reconciler: shutting down");
+    }
+}
+
+/// Query the node's active connection table and return the set of link IDs
+/// currently registered on it. Aggregates all chunks until `done=true`.
+async fn query_node_connections(
```
micpapal: same comment as for the link reconciler. all this functions can be part of the route reconciler itself

data-plane/control-plane/src/services/routes/route_reconciler.rs
```diff
@@ -0,0 +1,637 @@
...
+                }
+            }
+        }
+    }
+    Ok(result)
+}
+
+/// `NULL_COMPONENT` is the sentinel used by the data-plane for subscriptions
+/// without an explicit component ID (see `slim_datapath::messages::Name::NULL_COMPONENT`).
+const NULL_COMPONENT_ID: u64 = u64::MAX;
```
micpapal: this should be taken form the data-plane itself or defined in a common place, so that if we will change on one side everything will be in sync

data-plane/control-plane/src/services/northbound.rs
```diff
@@ -0,0 +1,392 @@
...
+                        let mut fields = std::collections::BTreeMap::new();
+                        fields.insert(
+                            "external_endpoint".to_string(),
+                            Value {
+                                kind: Some(Kind::StringValue(ee.clone())),
+                            },
+                        );
+                        metadata = Some(Struct { fields });
+                    }
+                    ConnectionDetails {
```
micpapal: now that we have a full connection setup this may be way bigger. I think we can simply put the json of the connection setting

data-plane/control-plane/src/db/model.rs
msardara: I think these are required for automatic link creation from the control plane.

My problem here is the client_config itself, which we should not set as part of the node configuration as most likely clients connecting from different places will have a different client_config.

I plan to remove this and add the possibility to create the link directly in the client as static configuration parameter, that we will reconcile with the control plane: in this way each client will have its own ClientConnection with custom parameter, no need to set it in the server.

data-plane/control-plane/src/db/model.rs
msardara: No need, I will switch to uuid

data-plane/control-plane/src/services/southbound.rs
```diff
@@ -0,0 +1,330 @@
...
+    // bound for crash-disconnected nodes that never re-register.
+    route_service.remove_node_lock(node_id).await;
+}
+
+/// Parse proto `ConnectionDetails` into the DB model.
+fn parse_conn_details(
+    peer_host: &str,
+    detail: &crate::api::proto::controller::proto::v1::ConnectionDetails,
+) -> ConnectionDetails {
+    let meta = detail.metadata.as_ref();
```
sancyx: In go version we saved matadata.client_config into ConnectionDetails, I see we dont do this here, so how do you pass details coming in metadata at node registration time?

sancyx (commented):
some comment around node registration, otherwise lgtm

data-plane/control-plane/src/services/southbound.rs
msardara: I slightly changed that part to use the client configured connection: basically when a SLIM node starts with a statically configured link in the conf file, and it is connected to the control plane, the controller on SLIM side will send that link as part of the registration, so that the control plane knows it when computing the routes

micpapal (approved):

Issue #1579, opened 2026-04-28
Use slimctl to send commands to the channel manager app


micpapal:
The channel manager application can get commands using the gRCP interface. Use slimctl to send commands to the channel manager application. 
Test that is all working as expected using the integration tests.

PR #1586, opened 2026-04-28
feat(slimctl): add commands for channel manager

micpapal:
# Description

Add commands to slimctl to manage groups using the channel manager application.

Also: use slim_config::grpc::client::ClientConfig to configure the gRCP connections between slimctl and the different endpoints

ref: #1579 

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


data-plane/slimctl/src/cli.rs
```diff
@@ -61,14 +65,6 @@ struct GlobalOpts {
     #[arg(long, env = "SLIMCTL_COMMON_OPTS_TIMEOUT", global = true)]
     timeout: Option<String>,
 
-    /// Disable TLS (plain HTTP/2)
-    #[arg(
-        long = "tls.insecure",
-        env = "SLIMCTL_COMMON_OPTS_TLS_INSECURE",
-        global = true
-    )]
-    tls_insecure: bool,
```
msardara, line 70: Did we remove this parameter on purpose?

data-plane/slimctl/src/commands/slim_cmd.rs
```diff
@@ -4,7 +4,7 @@
 use anyhow::{Context, Result};
 use clap::{ArgGroup, Args, Subcommand};
 
-use crate::defaults::DEFAULT_ENDPOINT;
+use crate::config::DEFAULT_NODE_ENDPOINT;
```
msardara: Now that DEFAULT_NODE_ENDPOINT is 46358, also the dslim node will start with that port bu default. Maybe we should create a new endpoint for the slim_cmd command.

msardara (commented):
LGTM, modulo there are some breaking changes. Are we targeting this for v2 right?

data-plane/slimctl/src/cli.rs
micpapal: yes, it is detected automatically by slimctl so you don't need to type --tls.insecure in the command

data-plane/slimctl/src/commands/slim_cmd.rs
micpapal: I added a new default with the right port

data-plane/slimctl/src/config.rs
```diff
@@ -4,128 +4,119 @@
...
+
+    // ── endpoint scheme ─────────────────────────────────────────────
+    if !config.endpoint.starts_with("http://") && !config.endpoint.starts_with("https://") {
+        let scheme = if config.tls_setting.insecure {
+            "http"
+        } else {
+            "https"
+        };
+        config.endpoint = format!("{}://{}", scheme, config.endpoint);
+    }
```
Tehsmash, line 112: Should we add an:
```
if config.endpoint.starts_with("http://") {
	<switch to insecure = true> 
}
```

msardara (approved):
LGTM ✅

PR #1596, opened 2026-05-01
refactor(data-plane): replace encoder::Name with ProtoName throughout

Tehsmash:
## Description

Removes the `encoder::Name` struct entirely and replaces it with `ProtoName` (the prost-generated protobuf message type) across all crates in the data plane.

`encoder::Name` was a conversion shim between human-readable string components and the 4×u64 XxHash64-encoded wire representation (`EncodedName`). Every message already carried a `ProtoName` directly in its `SlimHeader`, making `Name` a pure overhead layer: code would convert `ProtoName → Name` to get ergonomic helpers (`from_strings`, `with_id`, etc.), then immediately convert `Name → ProtoName` before sending.

### What changed

**`ProtoName` gains constructor/utility methods** (added to `api.rs`):
- `from_strings([s0, s1, s2])` — hashes components via XxHash64 and stores both the encoded and string forms
- `with_id(u64)`, `id()`, `has_id()`, `set_id()`, `reset_id()` — manage the 4th component (instance ID)
- `match_prefix()` — compare first 3 components, ignoring ID (used for subscription prefix matching)
- `str_components() -> (&str, &str, &str)` — access human-readable components
- `NULL_COMPONENT` / `is_reserved_id()` — sentinel value for unset IDs

**Routing hot path** — `ProtoName → EncodedName` is now a direct field access (`name.name.unwrap()`), with no intermediate `Name` allocation. All routing table keys are `EncodedName` (a `Copy` + `Hash` + `Eq` 4×u64 struct).

**Collection types** — `HashMap<Name, _>` / `HashSet<Name>` across session, service, and controller crates are replaced with `HashMap<EncodedName, _>` / `HashSet<EncodedName>` for hot-path collections, and `HashMap<ProtoName, _>` where the full name is needed as a value.

**`SlimHeader` constructor** — the `new_from_protos(ProtoName, ProtoName, ...)` / `new(&ProtoName, &ProtoName, ...)` split is collapsed into a single `new(ProtoName, ProtoName, ...)` that owns its arguments directly.

**Bindings** — the `Name` UniFFI wrapper in `bindings/rust/src/name.rs` now holds a `ProtoName` directly rather than converting to/from `encoder::Name`.

**`encoder.rs`** — the `Name` struct, all its `From` impls, and the re-exports in `messages/mod.rs` are removed. The file retains only `calculate_hash` (used internally by `ProtoName::from_strings`).

### Scope

68 files changed across `datapath`, `session`, `service`, `controller`, `bindings/rust`, `slimctl`, `testing`, and `examples`.

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


data-plane/core/datapath/src/message_processing.rs
msardara: I'd print the no match error in debug mode here, as we have the complete string at this point:


```suggestion
            Err(e) => {
                debug!(name = %header.get_dst(), %fanout, error = %e, "no match for publish destination");
                Err(DataPathError::MessageProcessingError {
                    source: Box::new(e),
                    msg: Box::new(msg),
                })
            }
```

data-plane/core/datapath/src/api.rs
```diff
@@ -43,3 +45,157 @@ pub use proto::dataplane::v1::message::MessageType::Publish as ProtoPublishType;
...
+
+    /// Returns `true` if `id` is the reserved null/unset sentinel.
+    pub const fn is_reserved_id(id: u64) -> bool {
+        id == Self::NULL_COMPONENT
+    }
+
+    /// Construct from three string components (org, namespace, app).
+    /// Encoded components are computed via XxHash64, matching the legacy `Name::from_strings` behaviour.
+    pub fn from_strings(components: [impl Into<String>; 3]) -> Self {
+        let strings = components.map(Into::into);
```
msardara: This will be probably optimized, but we could write it in this way:

```rust
let [s0, s1, s2] = components.map(Into::into);
```

And then skip the clone 

```rust
str_name: Some(StringName {
    str_component_0: s0,
    str_component_1: s1,
    str_component_2: s2,
}),
```

msardara (approved):
LGTM ✅

PR #1602, opened 2026-05-06
feat(slimctl): add bench sub/pub and channel sub/pub commands

Tehsmash:
## Description

Adds `slimctl bench` — a benchmarking tool for SLIM messaging performance, inspired by `nats bench`. It measures throughput and latency against a running SLIM server.

### Commands added

**Point-to-point bench**
- `slimctl bench sub` — start N subscriber workers, each waiting for a session from a paired publisher
- `slimctl bench pub` — start N publisher workers, each creating a point-to-point session to its paired subscriber and publishing messages; supports request-reply latency mode (`--request`)

**Channel (group session) bench**
- `slimctl bench channel sub` — start N subscriber workers that wait for a group session invite from the channel publisher
- `slimctl bench channel pub` — create a group session, invite all subscribers, broadcast N messages and report throughput

All commands report aggregate throughput (msgs/sec, bytes/sec) with per-worker min/avg/max/stddev, and optionally append results to a CSV file.

### Notable design decisions & fixes included

- Single connection per invocation: one `connect_async()` call in the orchestrator with `conn_id` passed to workers, avoiding a routing loop that would occur when multiple connections share the global `Service`.
- Throughput timer starts on first received message (not at session join) so idle time waiting for the publisher to finish inviting participants is excluded from the measurement.
- `--start-index` on `bench channel sub` allows the subscriber population to be split across multiple processes or machines without identity collisions.
- `--size` accepts human-readable units: plain bytes, SI decimal (`kb`/`mb`/`gb`, ×1000), or IEC binary (`kib`/`mib`/`gib`, ×1024).

## Type of Change

- [x] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/Taskfile.yaml
+++ b/data-plane/Taskfile.yaml
@@ -92,9 +92,9 @@ tasks:
     cmds:
       - cargo build --package agntcy-slimctl --profile {{.PROFILE}} --target {{.TARGET}}
       - mkdir -p ../.dist/bin
-      - cp target/debug/slimctl ../.dist/bin/slimctl
+      - cp target/{{.PROFILE}}/slimctl ../.dist/bin/slimctl
     vars:
-      PROFILE: '{{.PROFILE | default "dev"}}'
+      PROFILE: '{{.PROFILE | default "debug"}}'
       TARGET: '{{.TARGET | default ""}}'
 
   data-plane:telemetry:start:
--- a/data-plane/Taskfile.yaml
+++ b/data-plane/Taskfile.yaml
@@ -90,9 +90,12 @@ tasks:
   data-plane:slimctl:build:
     desc: "Build Rust slimctl and stage it in .dist/bin"
     cmds:
-      - cargo build --package agntcy-slimctl
+      - cargo build --package agntcy-slimctl --profile {{.PROFILE}} --target {{.TARGET}}
       - mkdir -p ../.dist/bin
       - cp target/debug/slimctl ../.dist/bin/slimctl
+    vars:
+      PROFILE: '{{.PROFILE | default "dev"}}'
+      TARGET: '{{.TARGET | default ""}}'
 
   data-plane:telemetry:start:
     desc: "Start telemetry stack"
--- a/data-plane/slimctl/src/commands/bench.rs
+++ b/data-plane/slimctl/src/commands/bench.rs
@@ -81,8 +81,8 @@ pub struct BenchSubArgs {
     #[arg(short = 'n', long = "msgs", default_value_t = 100_000)]
     pub msgs: u64,
 
-    /// Expected payload size in bytes (used for throughput calculation)
-    #[arg(long = "size", default_value_t = 128)]
+    /// Payload size — plain bytes or a unit suffix. SI: kb/mb/gb (×1000). IEC: kib/mib/gib (×1024).
+    #[arg(long = "size", default_value = "128", value_parser = parse_byte_size)]
     pub size: usize,
 
     /// SLIM server URL
@@ -122,8 +122,8 @@ pub struct BenchPubArgs {
     #[arg(short = 'n', long = "msgs", default_value_t = 100_000)]
     pub msgs: u64,
 
-    /// Payload size in bytes
-    #[arg(long = "size", default_value_t = 128)]
+    /// Payload size — plain bytes or a unit suffix. SI: kb/mb/gb (×1000). IEC: kib/mib/gib (×1024).
+    #[arg(long = "size", default_value = "128", value_parser = parse_byte_size)]
     pub size: usize,
 
     /// SLIM server URL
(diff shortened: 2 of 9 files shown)


data-plane/slimctl/src/commands/bench.rs
```diff
@@ -0,0 +1,1484 @@
...
+            .await
+        {
+            Ok(s) => break s,
+            Err(e) => {
+                // Log non-timeout errors so the user can see unexpected issues,
+                // but always retry — the session INIT may still be in transit.
+                let msg = e.to_string().to_lowercase();
+                if !msg.contains("timed out") {
+                    eprintln!("[sub-{i}] listen_for_session: {e} (retrying)");
+                }
```
msardara, line 714: Let's open a subissue to return a TimeoutError here, as we expose one in the bindings: https://github.com/agntcy/slim/blob/main/data-plane/bindings/rust/src/errors.rs#L28

This will prevent the string match, which will break if we change the error message in the future.

data-plane/Taskfile.yaml
```diff
@@ -90,9 +90,12 @@ tasks:
   data-plane:slimctl:build:
     desc: "Build Rust slimctl and stage it in .dist/bin"
     cmds:
-      - cargo build --package agntcy-slimctl
+      - cargo build --package agntcy-slimctl --profile {{.PROFILE}} --target {{.TARGET}}
       - mkdir -p ../.dist/bin
       - cp target/debug/slimctl ../.dist/bin/slimctl
```
msardara: Shouldn't this be target/{{.PROFILE}}/slimctl ?

data-plane/slimctl/src/commands/bench.rs
```diff
@@ -0,0 +1,1484 @@
...
+                "unknown size unit: '{other}' (use b/kb/mb/gb or kib/mib/gib)"
+            ));
+        }
+    };
+    usize::try_from(n * multiplier).map_err(|_| format!("size too large: '{s}'"))
+}
+
+fn human_bytes(bytes: f64) -> String {
+    const BASE: f64 = 1024.0;
+    const UNITS: &[&str] = &["B", "KB", "MB", "GB", "TB"];
```
msardara: For consistency with parse_byte_size:

```suggestion
    const UNITS: &[&str] = &["B", "KiB", "MiB", "GiB", "TiB"];
```

data-plane/slimctl/src/commands/bench.rs
```diff
@@ -0,0 +1,1484 @@
...
+        job_msg_cnt: if job_msg_cnt == 0 {
+            msg_cnt
+        } else {
+            job_msg_cnt
+        },
+        msg_cnt,
+        msg_bytes,
+        start,
+        end,
+    })
```
msardara: Should we extract an helper for this? I see it repeated twice

data-plane/slimctl/src/commands/bench.rs
```diff
@@ -0,0 +1,1484 @@
...
+                    eprintln!("[sub-{i}] listen_for_session: {e} (retrying)");
+                }
+                continue;
+            }
+        }
+    };
+
+    println!("[sub-{i}] session established");
+
+    let recv_timeout = Duration::from_secs(30);
```
msardara: We shouhld create a constant for this timeout (30s), the session timeout (35s), the number of session retries (60) and the interval (500ms)

data-plane/slimctl/src/commands/bench.rs
```diff
@@ -0,0 +1,1484 @@
...
+    #[arg(long)]
+    pub reply: bool,
+
+    /// Append results to CSV file
+    #[arg(long)]
+    pub csv: Option<String>,
+
+    /// Suppress per-second progress output
+    #[arg(long)]
+    pub no_progress: bool,
```
msardara: It looks like this is not used anywhere

data-plane/slimctl/src/commands/bench.rs
```diff
@@ -0,0 +1,1484 @@
...
+
+    let payload = vec![0u8; args.size];
+
+    let conn_id = get_global_service()
+        .connect_async(new_insecure_client_config(args.server.clone()))
+        .await
+        .context("connect to server failed")?;
+
+    let (result_tx, mut result_rx) =
+        tokio::sync::mpsc::channel::<(Sample, Vec<Duration>)>(args.count);
```
msardara: I think this channel could be replaced by just returning the result as return value of the spawned task, where we do `h.await`, what do you think? At the end that await is likely implemented as oneshot channel underneath, so we already have one channel per task we can use.

I see the same pattern in other location as well.

msardara (changes requested):
Nice work! Almost good to go for me, there are just few nits to fix.

msardara (approved):

Issue #1556, opened 2026-04-22
Add integrity check for packet header


micpapal:
## Description

Currently, SLIM packet headers have no integrity protection. Any intermediary can intercept a packet, modify the header (e.g., change source/destination names, alter routing fields), and forward it without detection. This opens the door to spoofing, misrouting, and tampering attacks.

This issue tracks the investigation and implementation of an integrity check mechanism for SLIM packet headers.

## Problem

- A malicious or compromised node can **modify packet headers** in transit (e.g., change the destination name to redirect traffic, alter the source name to impersonate another service).
- There is currently **no way for a receiving node to verify** that the header has not been tampered with.
- This undermines trust in the routing layer, even if the payload is encrypted.

## Goals

1. **Investigate** viable approaches for header integrity protection in SLIM.
2. **Design** a solution that balances security with performance.
3. **Implement** the chosen approach.

### 1. Packet Signature (Sender-side)

The sender signs the packet header (or header + payload hash) using its private key. The signature is attached to the packet.

**Pros:**
- Strong integrity guarantee — any modification is detectable.
- Non-repudiation — proves who sent the packet.

**Cons:**
- Signature verification is computationally expensive (especially with asymmetric crypto).
- Verifying every packet on every hop may be too costly at high throughput.

**Mitigation — Probabilistic Verification:**
To reduce cost, SLIM nodes could verify signatures on only a **random subset of packets** (e.g., 1 in N). This provides statistical detection of tampering without paying the full per-packet verification cost. The sampling rate could be configurable.

### 2. MLS-based Integrity

Investigate whether the  can provide header integrity as part of its existing mechanisms.

### 3. HMAC with Per-Link Shared Keys (Hop-by-Hop)

Each pair of connected SLIM nodes computes an HMAC over the packet header using a **symmetric key negotiated during the link creation handshake** (which already exists). This provides hop-by-hop integrity — each node verifies the HMAC from its direct neighbor before forwarding.

**Pros:**
- Much faster than asymmetric signature verification — can be applied to every packet.
- No additional key distribution needed — leverages the existing link handshake to negotiate the HMAC key.
- Hop-by-hop verification catches tampering at each link, preventing corrupted packets from propagating.

**Cons:**
- Only protects against tampering **between hops** — a compromised node can still modify the header and recompute a valid HMAC for the next hop.
- No end-to-end non-repudiation (for that, combine with approach 1).

## Performance Considerations

[...]

PR #1609, opened 2026-05-08
feat(data-plane): add node-to-node header integrity check

markpmarton:
contributes to #1556

# Description

Currently SLIM does not verify if the header has been tampered. Node-to-node HMAC validation will ensure the integrity of the packet header in each hop.
In case the header validation fails, the packet will be taken as tampered and the message won't be processed.

## Actions

- Added HMAC for relevant header fields to the packet header.
- Extended link negotiation with ECDH key exchange
- Integration tests for node-to-node hop validation check.

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


data-plane/core/datapath/src/message_processing.rs
```diff
@@ -161,6 +168,49 @@ impl MessageProcessor {
...
+
+        // Publishes must carry a MAC once the inter-node session has derived a key.  Control
+        // messages (subscribe / unsubscribe) may still traverse the same gRPC stream without a
+        // tag on some federation paths; skipping verification only when the tag is absent keeps
+        // tamper detection for application traffic.
+        if (message.is_subscribe() || message.is_unsubscribe()) && !has_wire_mac {
+            return Ok(());
+        }
+
+        let Some(mac) = conn.header_hmac() else {
```
msardara, line 309: **Silent downgrade risk.** When `conn.header_hmac()` is `None` and the wire has no MAC, verification returns `Ok(())`. An on-path attacker who can MITM link negotiation (TLS-terminated proxy, insecure link) can strip `link_ecdh_public_key` from both directions and disable integrity entirely — there is no way for either side to detect the downgrade.

Consider adding a `require_header_mac` option on server/client config; in strict mode, links that complete negotiation without an installed HMAC session should be rejected (and inbound publishes without a tag should fail closed).

data-plane/core/datapath/src/message_processing.rs
```diff
@@ -1090,6 +1237,19 @@ impl MessageProcessor {
...
+                                            && !msg.is_subscription_ack()
+                                            && let Err(e) = self_clone
+                                                .verify_remote_header_mac(conn_index, &msg)
+                                        {
+                                            error!(
+                                                %conn_index,
+                                                error = %e.chain(),
+                                                "SLIM header integrity verification failed",
+                                            );
+                                            break;
```
msardara: **Single bad packet kills the link (DoS).** `break` tears down the entire connection on any MAC failure — one tampered/replayed/malformed packet is enough to drop a federation link. Here we should drop-and-continue

data-plane/core/datapath/src/message_processing.rs
```diff
@@ -563,19 +664,61 @@ impl MessageProcessor {
...
+                    sk,
+                    payload.link_ecdh_public_key.as_slice(),
+                    link_id,
+                ) {
+                    Ok(mac) => conn.install_header_hmac(mac),
+                    Err(e) => {
+                        error!(
+                            %in_connection,
+                            error = %e,
+                            "link ECDH key derivation failed (client path)",
```
msardara, line 924: **Crypto failure → silent insecure operation (client path).** If `derive_header_mac_from_ecdh` fails here, the error is logged but negotiation still completes and the link runs without HMAC.

data-plane/core/datapath/src/message_processing.rs
```diff
@@ -563,19 +664,61 @@ impl MessageProcessor {
...
+                        {
+                            Ok(mac) => {
+                                conn.install_header_hmac(mac);
+                                server_reply_ecdh = Some(server_pk);
+                            }
+                            Err(e) => {
+                                error!(
+                                    %in_connection,
+                                    error = %e,
+                                    "link ECDH key derivation failed (server path)",
```
msardara, line 965: Same concern as the client path: ECDH/HKDF failure here results in a log message and a normal negotiation reply (just without a server pubkey), so the link operates without HMAC. Failures of `generate_x25519_ephemeral` or `derive_header_mac_from_ecdh` should fail the negotiation instead of silently dropping integrity.

msardara (commented):
Thanks for the PR @markpmarton!

The PR looks correct, with some nits, mainly around policy/enforcement

data-plane/core/datapath/src/connection.rs
```diff
@@ -63,6 +74,24 @@ pub struct Connection {
 
     /// Link negotiation state (link_id + remote_slim_version) under one lock for atomic check-and-set.
     negotiation: Arc<RwLock<NegotiationState>>,
+
+    /// Per-link HMAC and transient ECDH material for inter-node links.
+    link_crypto: Arc<RwLock<LinkCryptoState>>,
```
msardara: I think we should merge negotiation and link_crypto under a single RwLock

data-plane/core/datapath/src/header_mac.rs
```diff
@@ -0,0 +1,342 @@
...
+    }
+
+    #[test]
+    fn tamper_fails() {
+        let lid = Uuid::new_v4().to_string();
+        let mac = HeaderMacSession::new(&test_key()).unwrap();
+        let mut hdr = sample_header();
+        mac.sign_slim_header(&mut hdr, &lid).unwrap();
+        hdr.fanout = 3;
+        assert!(mac.verify_slim_header(&hdr, &lid).is_err());
```
msardara, line 343: Nice tamper test. Consider adding symmetric coverage for: (1) modifying the `header_mac` field itself (a malicious peer that flips one byte of the tag — should fail with `VerificationFailed`, not `InvalidTagLength`); (2) replaying a header signed under a different `link_id` (cross-link replay should fail thanks to HKDF salt).

data-plane/core/datapath/src/message_processing.rs
markpmarton: Now the strict header validation is enforced by default. It can be opt-out with setting `require_header_mac` to `false` on both client and server side.

data-plane/core/datapath/src/message_processing.rs
markpmarton: MAC failure from now on only causes package drop, no connection drop.

data-plane/core/datapath/src/message_processing.rs
markpmarton: Crypto computing functions now have correct error handling.

data-plane/core/datapath/src/message_processing.rs
markpmarton: Crypto computing functions now have correct error handling.

data-plane/core/datapath/src/header_mac.rs
markpmarton: Mentioned cases now covered with `tampered_mac_fail` and `cross_link_replay_fail` tests.

data-plane/core/datapath/src/connection.rs
markpmarton: Makes sense, new fields are moved to `NegotiationState`.

data-plane/core/datapath/src/subscription_ack.rs
```diff
@@ -62,11 +62,13 @@ impl RemoteSubAckManager {
     }
 }
 
-/// Returns `true` if the remote peer supports subscription ACKs (version ≥ 1.2.0).
+/// Returns `true` when the remote peer supports subscription ACKs (version ≥ 1.2.0)
+/// **and** inter-node header HMAC is active (ECDH completed on this link).
 pub(crate) fn supports(conn: &Connection) -> bool {
```
micpapal: here this can be removed I think as we are introducing breaking changes. we need to be sure that we do not fall back to an old version. This link negotiation now is mandatory

data-plane/core/datapath/src/message_processing.rs
```diff
@@ -329,20 +416,81 @@ impl MessageProcessor {
...
+    async fn await_link_hmac_ready(
+        &self,
+        conn_index: u64,
+        require_header_mac: bool,
+    ) -> Result<(), DataPathError> {
+        if !require_header_mac {
+            return Ok(());
+        }
+
+        let timeout = std::time::Duration::from_secs(5);
```
micpapal: this is a minor thing, but we may want to put a config value for this timeout. And this can default to 5 sec is not specified.

micpapal (changes requested):

data-plane/core/datapath/src/subscription_ack.rs
markpmarton: Old version fallback option is now removed, also with deprecated tests.

data-plane/core/datapath/src/message_processing.rs
markpmarton: Changes made, now the timeout and the time between the status polls are configurable.

micpapal (approved):
LGTM

PR #1611, opened 2026-05-11
perf(datapath): refactor SubscriptionTableImpl for cache efficiency

### LINKED ISSUE #1382: RwLock on subscription table per message

Tehsmash:
# Description

Redesigns \`SubscriptionTableImpl\` in the data-plane datapath for lower hot-path latency, motivated by flame-graph profiling of message routing under load.

**Data layout** — replaces the previous per-prefix state object with a \`PrefixEntry\` using a struct-of-arrays layout. A flat \`ids: Vec<u64>\` stores component_3 values densely (8 per cache line). Per-slot connection lists are held in two parallel \`Vec<ConnList>\` fields (\`local\` and \`remote\`). \`ConnList\` bundles a \`Vec<u64>\` with its own \`AtomicUsize\` round-robin cursor and exposes a \`next(skip)\` method, replacing the four separate Vecs that previously tracked connections and cursors independently.

**Lock design** — the routing table is stored in an \`ArcSwap<HashMap<[u64;3], PrefixEntry>>\`, making the read path entirely lock-free: \`match_one\`/\`match_all\` call \`load()\` (a single atomic pointer load) and never block. Subscription writes hold a \`Mutex<SubscriptionState>\` for bookkeeping, then clone the current routing snapshot, apply changes, and atomically store the new \`Arc\`. \`SubscriptionState\` is a separate cold structure that is never accessed by the read path.

**Routing logic** — a single \`HashMap.get(&prefix)\` covers all component_3 values under a prefix. For specific IDs, \`get_one\`/\`get_all\` scan the dense \`ids\` Vec and delegate to \`ConnList::next\` for round-robin selection. For \`NULL_COMPONENT\`, \`pick_one_any\`/\`get_all_any\` iterate all \`local\` ConnLists across every slot before touching any \`remote\` ConnLists, providing a true global local-over-remote preference. Slot rotation for \`NULL_COMPONENT\` uses an \`AtomicUsize slot_cursor\` on \`PrefixEntry\`, replacing the previous \`rand::rng\` call on the hot path.

**Criterion benchmarks** — \`subscription_table_benchmark\` covers \`match_one\` and \`match_all\` across deep (many IDs or connections per prefix) and wide (many distinct prefixes) topologies, with scale parameters up to 4096.

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/core/datapath/src/tables/subscription_table.rs
+++ b/data-plane/core/datapath/src/tables/subscription_table.rs
@@ -3,643 +3,211 @@
 
 use std::collections::{HashMap, HashSet};
 use std::fmt::{Display, Formatter};
-use std::hash::{Hash, Hasher};
+use std::sync::atomic::{AtomicUsize, Ordering};
 
-use parking_lot::{RawRwLock, RwLock, lock_api::RwLockWriteGuard};
+use parking_lot::RwLock;
 use rand::Rng;
-use tracing::{debug, error, warn};
+use tracing::{debug, warn};
 
 use super::SubscriptionTable;
-use super::pool::Pool;
 use crate::api::{EncodedName, ProtoName};
 use crate::errors::DataPathError;
 
-#[derive(Debug, Clone)]
-struct SubscriptionName([u64; 3]);
+// ──────────────────────────────────────────────────────────────────────────────
+// IdEntry – one entry per unique component_3 value registered under a prefix.
+//
+// `local` and `remote` are insertion-ordered Vecs (deduped).  The AtomicUsize
+// cursors advance under the read lock for round-robin selection.
+// ──────────────────────────────────────────────────────────────────────────────
 
-impl Hash for SubscriptionName {
-    fn hash<H: Hasher>(&self, state: &mut H) {
-        self.0[0].hash(state);
-        self.0[1].hash(state);
-        self.0[2].hash(state);
-    }
+#[derive(Debug)]
+struct IdEntry {
+    id: u64,
+    local: Vec<u64>,
+    remote: Vec<u64>,
+    local_cursor: AtomicUsize,
+    remote_cursor: AtomicUsize,
 }
 
-impl PartialEq for SubscriptionName {
-    fn eq(&self, other: &Self) -> bool {
-        self.0 == other.0
-    }
-}
-
-impl Eq for SubscriptionName {}
-
-#[derive(Debug, Default, Clone)]
-struct ConnId {
-    conn_id: u64,   // connection id
-    counter: usize, // number of references
-}
-
-impl ConnId {
-    fn new(conn_id: u64) -> Self {
-        ConnId {
-            conn_id,
-            counter: 1,
-        }
-    }
-}
-
-#[derive(Debug, Default)]
-struct SubscriptionRefs {
-    // map from connection id to set of subscription_ids
-    refs: HashMap<u64, HashSet<u64>>,
-}
-
-impl SubscriptionRefs {
-    fn new(conn: u64, subscription_id: u64) -> Self {
-        let mut set = HashSet::new();
-        set.insert(subscription_id);
-        let refs = HashMap::from([(conn, set)]);
-        SubscriptionRefs { refs }
-    }
-
-    /// Inserts a subscription_id for the given connection.
-    /// Returns true if the subscription_id was actually new (not a duplicate).
-    fn insert(&mut self, conn: u64, subscription_id: u64) -> bool {
-        self.refs.entry(conn).or_default().insert(subscription_id)
-    }
-
-    /// Removes a subscription_id for the given connection.
-    /// Returns `Ok(true)` if the connection has no remaining subscription_ids.
-    /// Returns `Ok(false)` if the connection still has other subscription_ids.
-    /// Returns `Err(SubscriptionIdNotFound)` if the subscription_id was not present.
-    /// Returns `Err(ConnectionIdNotFound)` if the connection was not found.
-    fn remove(&mut self, conn: u64, subscription_id: u64) -> Result<bool, DataPathError> {
-        match self.refs.get_mut(&conn) {
-            None => {
-                debug!(%conn, "connection not found in refs");
-                Err(DataPathError::ConnectionIdNotFound(conn))
-            }
-            Some(set) => {
-                if !set.remove(&subscription_id) {
-                    return Err(DataPathError::SubscriptionIdNotFound(subscription_id));
-                }
-                if set.is_empty() {
-                    self.refs.remove(&conn);
-                    Ok(true)
-                } else {
-                    Ok(false)
-                }
-            }
+impl IdEntry {
+    fn new(id: u64) -> Self {
+        IdEntry {
+            id,
+            local: Vec::new(),
+            remote: Vec::new(),
+            local_cursor: AtomicUsize::new(0),
+            remote_cursor: AtomicUsize::new(0),
         }
     }
 
-    fn force_remove(&mut self, conn: u64) -> Result<usize, DataPathError> {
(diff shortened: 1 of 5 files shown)


data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -3,663 +3,424 @@
...
-    // subscriptions table
-    // name -> name state
-    // if a subscription comes for a specific id, it is added
-    // to that specific id, otherwise the connection is added
-    // to the ProtoName::NULL_COMPONENT id
-    table: RwLock<HashMap<SubscriptionName, NameState>>,
-    // connections tables
-    // conn_index -> set(name)
-    connections: RwLock<HashMap<u64, HashSet<ProtoName>>>,
+    routing: ArcSwap<HashMap<[u64; 3], PrefixEntry>>, // lock-free on the hot read path
```
msardara: I've been thinking, should we use something like `ArcSwap<HashMap<[u64;3], Arc<PrefixEntry>>>` here? in this way, when we clone the table during updates, we will just increase the refefence count of existing items that do not need to be modified.

data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -911,7 +823,7 @@ mod tests {
         assert!(out.contains(&1));
         assert!(out.contains(&2));
 
-        // run multiple times for randomenes
+        // run multiple times for randomnes
```
```suggestion
        // run multiple times for randomness
```

data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -3,663 +3,424 @@
...
-                }
-                Ok(())
-            }
+    /// Ensure a slot for `id` exists, then insert `conn` (deduped).
+    fn insert_conn_for_id(&mut self, id: u64, conn: u64, is_local: bool) {
+        if !self.ids.contains(&id) {
+            self.push_id(id);
         }
+        let idx = self.ids.iter().position(|&i| i == id).unwrap();
+        self.insert_conn(idx, conn, is_local);
```
```suggestion
        let idx = match self.ids.iter().position(|&i| i == id) {
            Some(i) => i,
            None => {
                self.push_id(id);
                self.ids.len() - 1
            }
        };
        self.insert_conn(idx, conn, is_local);
```

data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -3,663 +3,424 @@
...
-        self.refs
-            .remove(&conn)
-            .map(|set| set.len())
-            .ok_or(DataPathError::ConnectionIdNotFound(conn))
+    /// Remove `conn` (no-op if absent).
+    fn remove(&mut self, conn: u64) {
+        if let Some(pos) = self.conns.iter().position(|&c| c == conn) {
+            self.conns.swap_remove(pos);
+        }
     }
```
msardara, line 49: Here we should add a comment that this linear scans are ok for up to 256 items, and then a hashset becomes better.

msardara (changes requested):
Looks almost GTM, but I'll wait for @micpapal review as well.

data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -3,663 +3,424 @@
...
+        for remote in &self.remote {
+            for c in remote.iter() {
+                if c != skip && !out.contains(&c) {
+                    out.push(c);
                 }
-                debug!("no output connection available");
-                None
             }
         }
+        out
```
micpapal: This function has many iterations that are required to check duplicated conn_ids across all the vectors. Do you think it makes sense to keep track of the unique connection across all the name ids? If we add something like 
```
all_local_conns: Vec<u64>
all_remote_conns: Vec<u64>
```
to PrefixEntry this function becomes something like
```
 fn get_all_any(&self, skip: u64) -> Vec<u64> {
     self.all_local_conns.iter()
         .chain(self.all_remote_conns.iter())
         .copied()
         .filter(|&c| c != skip)
         .collect()
 }
```
pick_one_any could also use these vectors. Of course this needs to be maintained in the insert/remove operation. 

Do you think we can gain something or it will only add complexity to the data structure?

data-plane/core/datapath/src/tables/subscription_table.rs
micpapal: agree, this should improve the clone

data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -670,14 +431,41 @@ impl SubscriptionTable for SubscriptionTableImpl {
...
+            let (s0, s1, s2) = name.str_components();
+            PrefixEntry::new([s0.to_string(), s1.to_string(), s2.to_string()])
+        });
+
+        // 2 & 3. Ensure an ID slot exists and insert conn (deduped).
+        prefix_entry.insert_conn_for_id(id, conn, is_local);
+        self.routing.store(Arc::new(rs));
+
+        // 4. Record the subscription (idempotent: same sub_id overwrites itself).
+        ss.subscriptions.insert(
```
micpapal: this insert and the and the check at step 5 can be implemented in the SubscriptionState struct itself I think. Same for the remove steps in the remove_subscription function.

data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -3,663 +3,424 @@
...
+// SubscriptionState — COLD write path.
+//
+// Protected by a Mutex (exclusive access always required). Never held by
+// match_one / match_all, so subscription churn never blocks message routing.
+// ──────────────────────────────────────────────────────────────────────────────
+
 #[derive(Debug, Default)]
+struct SubscriptionState {
+    subscriptions: HashMap<u64, SubRecord>, // sub_id  → (encoded, conn_id)
+    conn_subs: HashMap<u64, Vec<u64>>,      // conn_id → sub_ids
```
micpapal: here the state is protected by a mutex so we don't care to much about performance but I think use an hash set as value in conn_subs here is a better fit here.

micpapal (changes requested):

micpapal:
> Looks almost GTM, but I'll wait for @micpapal review as well.

I think is a great improvement. we can probably improve a bit more with few changes.

data-plane/core/datapath/src/tables/subscription_table.rs
Tehsmash: Added these aggregates to speed up this path, it makes a good difference in the benchmark for the fan out case.

micpapal (approved):
lgtm

msardara (approved):
LGTM, thanks @Tehsmash!

PR #1624, opened 2026-05-12
perf(datapath): replace RwLock with ArcSwap on the connection table read path

### LINKED ISSUE #1383: RwLock + Arc::clone on connection lookup per message
**File:** `core/datapath/src/tables/connection_table.rs` — `get()`

**Problem:**
```rust
pub fn get(&self, index: usize) -> Option<Arc<T>> {
    let pool = self.pool.read();       // RwLock read
    pool.get(index).cloned()           // Arc refcount bump
}
```
Every `send_msg` call acquires the connection-table RwLock and bumps the Arc
refcount.  Arc refcounting is an atomic RMW instruction per clone + per drop.

**Mitigation plan:**
1. Store connections in a lock-free indexed structure (e.g., a `slab` with
   atomic slots or `sharded_slab`).  Index-based lookups become O(1) without
   a lock.
2. If keeping the current Pool, cache the `Arc<Connection>` in the
   `process_stream` loop for the duration of the stream rather than looking it
   up per message.

Tehsmash:
# Description

Follow-up to #1611. Flame-graph profiling showed that `ConnectionTable::get()` on the hot publish path was dominated by `RwLock::read()` acquisition, the same issue already fixed in `SubscriptionTableImpl`.

This PR makes the same change for `ConnectionTable<T>`:

- **Read path** (`get`, `for_each`, `len`, `is_empty`): completely lock-free via `ArcSwap::load()` — a single atomic pointer load, no CAS loop, no mutex.
- **Write path** (`insert`, `insert_at`, `remove`): serialised by a `Mutex<()>` (`write_lock`) to prevent lost-update races in the load → clone → modify → store sequence. No write lock is held during reads.
- **`Arc<T>` preserved**: connections are stored as `Arc<T>` so pool clones (performed on every write) only bump refcounts rather than deep-copying each `Connection`. `get()` returns `Option<Arc<T>>`; callers auto-deref into `T` unchanged.

`Pool<T>` and `PoolEntry<T>` gained manual `Clone` impls (`AtomicUsize` does not implement `Clone`) that snapshot cursor values with `Ordering::Relaxed`.

Fixes: #1383

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [x] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/core/datapath/src/tables/connection_table.rs
+++ b/data-plane/core/datapath/src/tables/connection_table.rs
@@ -3,17 +3,32 @@
 
 use std::sync::Arc;
 
-use parking_lot::RwLock;
+use arc_swap::ArcSwap;
+use parking_lot::Mutex;
 
 use super::pool::Pool;
 
+/// A connection table that provides lock-free reads via `ArcSwap`.
+///
+/// Concurrency design:
+///   - Hot read path (`get`, `for_each`, `len`, `is_empty`): lock-free via
+///     `ArcSwap::load()`.  A single atomic pointer load gives callers an
+///     immutable snapshot of the pool.
+///   - Write path (`insert`, `insert_at`, `remove`): serialised by
+///     `write_lock`.  Writers load the current snapshot, clone the pool,
+///     apply the change, then atomically store the new `Arc`.  No write
+///     lock is ever held during reads.
+///
+/// Connections are stored as `Arc<T>` so that pool clones (made on every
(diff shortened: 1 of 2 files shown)


micpapal (approved):
This is good to go after the modifications in https://github.com/agntcy/slim/pull/1611

PR #1628, opened 2026-05-13
fix(session): guard against late GroupAck after task completion

Tehsmash:
# Description

Fixes a panic in `SessionModerator` that occurs when a late or
retransmitted `GroupAck` arrives after the current task has already
been cleaned up.

The existing code called `.unwrap()` on `self.current_task` unconditionally
when processing a `GroupAck`. Under high-throughput or lossy conditions a
retransmitted ACK can arrive after the task entry has been removed, producing
an unexpected `None` and a panic.

The fix replaces the unconditional unwrap with a `let Some(task) = ... else`
guard. If no task is active the stale ACK is silently discarded with a debug
log entry, matching the existing `is_still_pending` defensive pattern in the
same function.

This panic was surfaced during `slimctl bench` channel benchmarking runs.

## Type of Change

- [x] Bugfix

## Checklist

### DIFF
--- a/data-plane/core/session/src/session_moderator.rs
+++ b/data-plane/core/session/src/session_moderator.rs
@@ -2134,4 +2134,41 @@ mod tests {
             "Participant2 should still be in group (task queued, not processed)"
         );
     }
+
+    /// A late or retransmitted GroupAck arriving when `current_task` is `None`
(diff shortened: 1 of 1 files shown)


msardara (approved):
LGTM ✅

PR #1638, opened 2026-05-15
feat(websocket): add WebSocket transport for data-plane

hackeramitkumar:
# Description

Adds a WebSocket transport alongside the existing gRPC transport in the data-plane config and datapath layers, and refactors the config crate so both transports share the same client/server entry points, auth pipeline, and proxy machinery.

## What changes

**New WebSocket transport (`core/config/src/websocket/`)**
- Client and server built on `hyper_util::client::legacy::Client` + `hyper::server::conn::http1`, with the WS upgrade completed via `fastwebsockets::WebSocket::after_handshake`.
- Auth is layered with `tower::ServiceBuilder` using the same `ClientAuthenticator::get_client_layer()` / `ServerAuthenticator::get_server_layer()` factories used by gRPC — no transport-specific
 auth code paths.
- `bearer_to_query_layer` (client) + `query_token_layer` (server) mirror a `Authorization: Bearer …` header into a configurable query parameter for environments where headers can't traverse the
upgrade.
- Datapath integration in `core/datapath/src/websocket/stream.rs` with biased control-frame handling (Pong/Close) over an mpsc channel.

**Config refactor (`core/config/src/`)**
- `ClientConfig` / `ServerConfig` split into transport-agnostic structs in `client.rs` / `server.rs` with a polymorphic `to_channel()` / `run_server()` entry point; gRPC- and WS-specific impls
live under `grpc/` and `websocket/`.
- `ConfigError` moved to the crate root so it can house WS, Unix-socket, and gRPC variants together (the old `GrpcConfigError` alias is renamed everywhere it was re-exported).
- HTTPS-connector and proxy-tunnel helpers lifted into `transport_common.rs` (`Alpn`, `build_https_connector`, `ProxyTunnel`, `ClientConfig::build_proxy_tunnel`) so gRPC and WS share one
implementation of TLS, HTTP/HTTPS proxy CONNECT, proxy auth, and custom proxy headers.
- Backward-compat re-exports under `grpc::client` / `grpc::server` keep existing import paths (used by the language bindings) working.

**Other**
- `wss://` example configs added under `data-plane/config/websocket/`.
- Integration tests for the WS transport (handshake, auth variants, query-token mirror, full datapath round-trip).

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


data-plane/core/datapath/src/websocket/stream.rs
```diff
@@ -0,0 +1,188 @@
...
+                    let _ = write_half.write_frame(Frame::close(1000, &[])).await;
+                    let _ = write_half.flush().await;
+                    break;
+                }
+                // Prioritise control frames so Pongs / Close echoes are not
+                // starved by a heavy outbound binary stream.
+                maybe_ctrl = rx_control.recv() => {
+                    let Some(ctrl) = maybe_ctrl else {
+                        continue;
+                    };
```
msardara, line 175: Here we should definitely break if we receive an error because the stream was closed. Calling continue will end up in a busy loop.

data-plane/core/datapath/src/websocket/stream.rs
```diff
@@ -0,0 +1,188 @@
...
+                                    .send(Err(Status::invalid_argument(format!(
+                                        "invalid protobuf payload in websocket frame: {err}",
+                                    ))))
+                                    .await;
+                            }
+                        },
+                        OpCode::Close => break,
+                        OpCode::Text => {
+                            warn!("ignoring text websocket frame, expected binary protobuf frame");
+                        }
```
msardara, line 138: Checking RFC 6455, this is a protocol violation for a binary subprotocol — we should close the connection with status 1003 (unsupported data) instead of logging-and-continuing.

data-plane/core/config/src/websocket/common.rs
```diff
@@ -0,0 +1,245 @@
...
+fn extract_query_param(query: Option<&str>, name: &str) -> Option<String> {
+    let query = query?;
+
+    for pair in query.split('&') {
+        let mut parts = pair.splitn(2, '=');
+        let key = parts.next().unwrap_or_default();
+        let value = parts.next().unwrap_or_default();
+
+        if key == name && !value.is_empty() {
+            return Some(value.to_string());
```
msardara: Here we need URL decoding. A token with `=` padding or `+` characters will fail verification, unless the client URL-encodes before sending.

We should percent-decode the value here.

data-plane/core/config/src/websocket/server.rs
```diff
@@ -0,0 +1,220 @@
...
+                                serve_connection(
+                                    stream,
+                                    auth,
+                                    expected_path,
+                                    on_accepted,
+                                    remote_addr,
+                                    local_addr,
+                                )
+                                .await;
+                            });
```
msardara: The two arms duplicate the entire `serve_connection` call. We should be able to use the `MaybeTlsStream` enum to avoid doing the same operation on both sides of the if.

data-plane/core/config/src/websocket/server.rs
```diff
@@ -0,0 +1,220 @@
...
+                        };
+
+                        let local_addr = stream.local_addr().ok();
+                        let auth = auth.clone();
+                        let expected_path = expected_path.clone();
+                        let on_accepted = on_accepted.clone();
+
+                        if let Some(acceptor) = tls_acceptor.clone() {
+                            tokio::spawn(async move {
+                                let stream = match acceptor.accept(stream).await {
```
msardara: We should set a timeout on `acceptor.accept(stream)` and on the subsequent HTTP/WS upgrade in `serve_connection`.

A malicious or silent client can pin a task indefinitely.

data-plane/core/config/src/client.rs
```diff
@@ -0,0 +1,683 @@
...
+use crate::errors::ConfigError;
+use crate::grpc::compression::CompressionType;
+use crate::grpc::proxy::ProxyConfig;
+use crate::tls::client::TlsClientConfig as TLSSetting;
+use crate::transport::TransportProtocol;
+use crate::websocket::client::WebSocketClientChannel;
+
+/// Result of [`ClientConfig::to_channel`]: either a gRPC channel (the generic
+/// `G` parameter) or a WebSocket channel.
+#[allow(clippy::large_enum_variant)]
```
msardara: We should probably Box the Websocket channel. I know the grpc channel is a wrap aroung an Arc, we should do the same for the Websocket one.

data-plane/core/config/src/websocket/client.rs
```diff
@@ -0,0 +1,157 @@
...
+
+        let websocket = if endpoint.secure {
+            let tls_config = self.tls_setting.load_rustls_config().await?;
+            let tls_config = tls_config.ok_or(ConfigError::WebSocketTlsConfiguration)?;
+            let connector = TlsConnector::from(Arc::new(tls_config));
+
+            let server_name = self
+                .server_name
+                .as_deref()
+                .unwrap_or(endpoint.host.as_str());
```
msardara: I'd change it to walk the chain explicitly:

```rust
let server_name = self
    .server_name
    .as_deref()
    .or_else(|| self.origin.as_deref().and_then(host_from_authority))
    .unwrap_or(endpoint.host.as_str());

Where host_from_authority strips any :port from the origin (since origin is a Host authority — example.com:8443 — but SNI takes only the hostname). Something like:

fn host_from_authority(authority: &str) -> Option<&str> {
    authority.rsplit_once(':').map(|(h, _)| h).or(Some(authority))
        .filter(|h| !h.is_empty())
}
```

  Why this order:
  - server_name is the explicit SNI override — highest precedence, already correct.
  - origin is the user saying "pretend the Host is X" — if they bothered to set it, they almost certainly want TLS verification against X too, otherwise cert validation fails on a name they didn't pick.
  - endpoint.host is the connect target — last-resort default

data-plane/core/datapath/tests/data_path_test.rs
```diff
@@ -206,6 +259,81 @@ mod tests {
...
+            .with_tls_setting(client_tls);
+        let conn_index =
+            run_transport_roundtrip(server_conf, client_conf, None, "websocket wss").await;
+
+        assert!(logs_contain(
+            "received message from connection conn_index=0"
+        ));
+        let expected = format!("received message from connection conn_index={conn_index}");
+        assert!(logs_contain(expected.as_str()));
+    }
```
msardara, line 356: We are missing some tests paths:
- Auth: no `test_websocket_*_basic_auth` / `_jwt_auth` / `_jwt_query_param` test — the `authorize()` logic in `common.rs` is currently unexercised.
- Path mismatch (the 404 branch in `serve_connection`).
- Handshake / connect timeout.
- Close-frame echo and Pong forwarding — the control-frame logic is non-trivial and otherwise untested.

Also: `sleep(100ms)` for readiness/ordering tends to flake in CI — prefer polling the listener or having `run_server` signal ready.

msardara (changes requested):

data-plane/core/datapath/tests/data_path_test.rs
amitami2: done

data-plane/core/config/src/websocket/client.rs
amitami2: done

data-plane/core/config/src/client.rs
hackeramitkumar: done

data-plane/core/config/src/websocket/server.rs
hackeramitkumar: done

data-plane/core/config/src/websocket/server.rs
hackeramitkumar: done

data-plane/core/config/src/websocket/common.rs
hackeramitkumar: done

data-plane/core/datapath/src/websocket/stream.rs
hackeramitkumar: done

data-plane/core/datapath/src/websocket/stream.rs
hackeramitkumar: done

data-plane/core/config/src/websocket/server.rs
```diff
@@ -0,0 +1,466 @@
...
+    }
+
+    #[tokio::test]
+    async fn test_websocket_server_401_on_failed_basic_auth() {
+        let port = available_port();
+        // pragma: allowlist secret
+        let cfg = ServerConfig::with_endpoint(&format!("ws://127.0.0.1:{port}"))
+            .with_transport(TransportProtocol::Websocket)
+            .with_tls_settings(TlsServerConfig::insecure())
+            .with_auth(ServerAuthConfig::Basic(BasicConfig::new("user", "pass")));
```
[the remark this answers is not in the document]
hackeramitkumar: Using this for the unit tests.

data-plane/core/datapath/src/websocket/stream.rs
```diff
@@ -0,0 +1,388 @@
...
+
+        server_token.cancel();
+    }
+
+    #[tokio::test]
+    async fn test_websocket_multiple_sequential_connections() {
+        let port = available_port();
+        let server_token = start_server_with_transport_tasks(port).await;
+
+        for i in 0..3 {
```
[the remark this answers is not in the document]
amitami2: done

data-plane/core/config/src/websocket/common.rs
```diff
@@ -0,0 +1,681 @@
...
+        ServerAuthConfig::Basic(basic) => ServerHandshakeAuth::Basic {
+            username: basic.username().to_string(),
+            password: basic.password().as_str().to_string(),
+        },
+        ServerAuthConfig::Jwt(jwt) => ServerHandshakeAuth::Jwt {
+            config: jwt.clone(),
+        },
+        #[cfg(not(target_family = "windows"))]
+        ServerAuthConfig::Spire(_) => ServerHandshakeAuth::None,
+    }
```
msardara: This will silently accept any client. Either implement SPIRE verification here or return an error (`unsupported`) so we can early fail the server rather than accept any client.

data-plane/core/config/src/websocket/client.rs
```diff
@@ -0,0 +1,268 @@
...
+}
+
+/// Strip any `:port` suffix from an HTTP authority, returning just the host.
+///
+/// Used to derive an SNI name from the `Origin` header value (e.g.
+/// `example.com:8443` -> `example.com`). Returns `None` for empty input so
+/// callers can fall back to the next name in the chain.
+fn host_from_authority(authority: &str) -> Option<&str> {
+    let host = authority
+        .rsplit_once(':')
```
msardara: `authority.rsplit_once(':')` breaks IPv6 authorities like `[::1]:8080` — it splits on the colon inside the address. We should use a proper parser (e.g. `http::uri::Authority::host()`) or detect a bracketed host first.

data-plane/core/config/src/websocket/client.rs
```diff
@@ -0,0 +1,268 @@
...
+                .server_name
+                .as_deref()
+                .or_else(|| self.origin.as_deref().and_then(host_from_authority))
+                .unwrap_or(endpoint.host.as_str());
+            let server_name =
+                tokio_rustls::rustls::pki_types::ServerName::try_from(server_name.to_string())
+                    .map_err(|_| ConfigError::WebSocketTlsConfiguration)?;
+
+            let tls_stream = connector
+                .connect(server_name, stream)
```
msardara: We should wrap `connector.connect(...).await`) in a timeout, otherwise we risk to block the client undefinitely if the server is not responsive.

data-plane/core/config/src/websocket/common.rs
```diff
@@ -0,0 +1,681 @@
...
+            secure,
+            host,
+            authority,
+            port,
+            path,
+        })
+    }
+
+    pub fn socket_address(&self) -> String {
+        format!("{}:{}", self.host, self.port)
```
msardara: `format!("{}:{}", self.host, self.port)` produces invalid socket addrs for IPv6 hosts. We should use something like `format!("[{}]:{}", host, port)` when `host` contains `:`

data-plane/core/config/src/websocket/common.rs
```diff
@@ -0,0 +1,681 @@
...
+        if let Some((key, value)) = query_param
+            && !key.is_empty()
+            && !value.is_empty()
+        {
+            if self.uri.query().is_some() {
+                uri.push('&');
+            } else {
+                uri.push('?');
+            }
+            uri.push_str(key);
```
msardara: We should pecent encode this before pushing it in the uri, otherwise we will get issue with special characters like `?`, `&` etc.

data-plane/core/datapath/src/websocket/stream.rs
```diff
@@ -0,0 +1,402 @@
...
+        // on the control channel if the read task exits first.
+        drop(control_tx);
+        read_cancel.cancel();
+    });
+
+    let write_cancel = cancellation_token.clone();
+    tokio::spawn(async move {
+        loop {
+            tokio::select! {
+                _ = write_cancel.cancelled() => {
```
msardara: The doc comment above states cancellation/control frames take priority, but this `tokio::select!` has no `biased;` — branch order is randomized. Add `biased;` as the first line of the `select!` to match the documented behavior:

```rust
tokio::select! {
    biased;
    _ = cancellation_token.cancelled() => { ... }
    ...
}
```

data-plane/core/config/src/websocket/server.rs
```diff
@@ -0,0 +1,573 @@
...
+                        };
+
+                        let local_addr = stream.local_addr().ok();
+                        let auth = auth.clone();
+                        let expected_path = expected_path.clone();
+                        let on_accepted = on_accepted.clone();
+                        let tls_acceptor = tls_acceptor.clone();
+
+                        tokio::spawn(async move {
+                            let stream = match tls_acceptor {
```
msardara, line 250: We create a task for each `tokio::spawn` per accepted connection — there's no max-connection cap or semaphore.

A malicious peer can exhaust memory/FDs.

We should reuse the `max_concurrent_streams` parameter and bound the max number of streams (tasks).

data-plane/core/config/src/websocket/server.rs
```diff
@@ -0,0 +1,573 @@
...
+pub type OnAcceptedWebSocket = Arc<
+    dyn Fn(AcceptedWebSocketConnection) -> Pin<Box<dyn Future<Output = ()> + Send>> + Send + Sync,
+>;
+
+impl ServerConfig {
+    pub async fn run_websocket_server(
+        &self,
+        drain_rx: drain::Watch,
+        on_accepted: OnAcceptedWebSocket,
+    ) -> Result<CancellationToken, ConfigError> {
```
msardara, line 175: `ConfigError::WebSocketTlsConfiguration` is reused across ~5 distinct failure modes here (cert load, key load, builder, acceptor wrap, etc.).

We should create more specific errors with clear messages for the callers.

msardara (changes requested):

data-plane/core/config/src/websocket/common.rs
amitami2: done

data-plane/core/config/src/websocket/client.rs
amitami2: done

data-plane/core/config/src/websocket/common.rs
amitami2: done

data-plane/core/config/src/websocket/common.rs
amitami2: done

data-plane/core/datapath/src/websocket/stream.rs
amitami2: done

data-plane/core/config/src/websocket/server.rs
amitami2: done

data-plane/core/config/src/websocket/server.rs
amitami2: done

data-plane/core/config/src/client.rs
```diff
@@ -0,0 +1,689 @@
...
+            .with_transport(TransportProtocol::Websocket);
+        let err = invalid
+            .validate()
+            .expect_err("expected invalid websocket scheme");
+        assert!(matches!(err, ConfigError::InvalidWebSocketEndpointScheme));
+    }
+
+    #[test]
+    fn test_client_config_with_proxy() {
+        let test_password = "test_password_123";
```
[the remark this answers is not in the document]
amitami2: done

data-plane/core/datapath/src/websocket/stream.rs
```diff
@@ -0,0 +1,403 @@
...
+                        (1000, Vec::new())
+                    };
+                    Some(ControlFrame::Close { code, reason })
+                }
+                _ => None,
+            };
+            let tx = control_tx.clone();
+            async move {
+                if let Some(ctrl) = owned {
+                    let _ = tx.send(ctrl).await;
```
msardara: `tx.send(ctrl).await` here is awaited inside the auto-Pong/Close callback, while the outer `select!` only races the *outer* `read_frame` against cancellation. A peer flooding Pings while the write task is blocked on a slow socket can fill `tx_control` (cap 8) and pin the read task indefinitely on this `send`.

Since Pong is best-effort, use `try_send` and drop on full (log at `debug!`).

data-plane/core/config/src/websocket/server.rs
```diff
@@ -0,0 +1,607 @@
...
+    let connection = http1::Builder::new()
+        .serve_connection(io, service)
+        .with_upgrades();
+
+    // Bound the HTTP/WS upgrade phase so a client that opens a TCP/TLS
+    // connection but never sends a valid upgrade request cannot pin this
+    // task. After the upgrade succeeds the underlying IO is hijacked by the
+    // spawned task above, so `serve_connection` should resolve quickly and
+    // before the timeout in the happy path.
+    match tokio::time::timeout(HTTP_UPGRADE_TIMEOUT, connection).await {
```
msardara: `tokio::time::timeout(HTTP_UPGRADE_TIMEOUT, connection)` bounds the *entire* `serve_connection` future, not just the upgrade phase. The `upgrade_done` flag only changes the log level on timeout; the connection is still dropped at 10s regardless.

In practice this is OK because `with_upgrades()` hijacks the IO into the spawned task above, so the outer future resolves quickly in the happy path. But it is fragile — a slow client header read after a successful upgrade could still race the timeout. Consider making the timeout configurable, or only timing out until `upgrade_done` flips.

data-plane/core/config/src/websocket/common.rs
msardara: We should implement it in a follow up PR. Let's open an issue for it

msardara (changes requested):

data-plane/core/config/src/websocket/common.rs
amitami2: sure

msardara (approved):

PR #1662, opened 2026-05-22
refactor: drop transport field from the config

hackeramitkumar:
# Description

Remove the transport: attribute from server and connection configuration and
use the protocol in the url (e.g. `https://`, `wss://`, `ws://`, `unix://` etc.)

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/bindings/rust/src/client_config.rs
+++ b/data-plane/bindings/rust/src/client_config.rs
@@ -433,8 +433,6 @@ mod tests {
     use super::*;
     use crate::common_config::{CaSource, TlsSource};
     use crate::errors::SlimError;
-    #[allow(unused_imports)]
-    use slim_config::component::configuration::Configuration as _;
     use slim_config::transport::TransportProtocol as CoreTransportProtocol;
     use std::collections::HashMap;
 
--- a/data-plane/bindings/rust/src/client_config.rs
+++ b/data-plane/bindings/rust/src/client_config.rs
@@ -17,7 +17,6 @@ use slim_config::grpc::client::{
 
 use crate::common_config::{ClientAuthenticationConfig, TlsClientConfig};
 use crate::errors::SlimError;
-use crate::transport_protocol::TransportProtocol;
 
 use slim_config::component::configuration::Configuration;
 
@@ -261,11 +260,11 @@ impl From<CoreBackoffConfig> for BackoffConfig {
 /// Client configuration for connecting to a SLIM server
 #[derive(uniffi::Record, Clone, Debug, PartialEq)]
 pub struct ClientConfig {
-    /// The target endpoint the client will connect to
+    /// The target endpoint the client will connect to.
+    ///
+    /// The transport protocol is inferred from the endpoint scheme:
+    /// `ws://`/`wss://` → WebSocket, otherwise gRPC.
     pub endpoint: String,
-
-    /// Transport protocol to use (defaults to gRPC in core config when omitted)
-    pub transport: Option<TransportProtocol>,
 
     /// TLS client configuration
     pub tls: TlsClientConfig,
@@ -315,10 +314,6 @@ impl From<ClientConfig> for CoreClientConfig {
         let core_defaults = CoreClientConfig::default();
         CoreClientConfig {
             endpoint: config.endpoint,
-            transport: config
-                .transport
-                .map(Into::into)
-                .unwrap_or(core_defaults.transport),
             origin: config.origin,
             server_name: config.server_name,
             compression: config.compression.map(Into::into),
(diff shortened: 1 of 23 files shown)


data-plane/core/config/src/server.rs
```diff
@@ -305,6 +300,12 @@ impl ServerConfig {
             }
         }
     }
+
+    /// Resolve the transport protocol for this configuration by inspecting
+    /// the endpoint URI scheme. See [`TransportProtocol::from_endpoint`].
+    pub fn resolved_transport(&self) -> TransportProtocol {
```
msardara, line 307: `ServerConfig::validate` does not validate the endpoint scheme, while `ClientConfig::validate` now calls `validate_endpoint_scheme()`. A server configured with e.g. `endpoint: "ftp://0.0.0.0:46357"` will silently resolve to gRPC via `TransportProtocol::from_endpoint`'s catch-all rather than erroring. 

I'd mirror `validate_endpoint_scheme` on `ServerConfig` (or factoring the check onto `TransportProtocol`) so both sides reject unknown schemes consistently.

data-plane/core/config/src/server.rs
```diff
@@ -82,12 +82,12 @@ pub enum AuthenticationConfig {
 #[derive(Debug, Serialize, Deserialize, PartialEq, Clone, JsonSchema)]
 pub struct ServerConfig {
     /// Endpoint is the address to listen on.
+    ///
+    /// The transport protocol is inferred from the endpoint scheme:
+    /// * `ws://`, `wss://`  → WebSocket (TLS when `wss`)
```
msardara, line 87: Doc is inconsistent with `ClientConfig`'s endpoint doc, which lists `unix://` as gRPC. Add `unix://` here too.

data-plane/core/config/src/client.rs
```diff
@@ -488,26 +482,33 @@ impl ClientConfig {
...
-    /// Validates that the endpoint scheme matches the transport when set to
-    /// WebSocket. For gRPC transports the scheme is checked at channel-build
-    /// time by `grpc::client`.
-    fn validate_websocket_endpoint(&self) -> Result<(), ConfigError> {
-        if self.transport != TransportProtocol::Websocket {
-            return Ok(());
-        }
+    /// Resolve the transport protocol for this configuration by inspecting
+    /// the endpoint URI scheme. See [`TransportProtocol::from_endpoint`].
+    pub fn resolved_transport(&self) -> TransportProtocol {
```
msardara, line 495: Minor: `resolved_transport()` is now called repeatedly from `Display`, `to_channel`, `validate`, and the grpc/websocket builders, each re-parsing the endpoint via `split_once`.

It's cheap, but if you'd prefer not to re-parse, you could cache the resolved value at construction (e.g. compute once in `with_endpoint`/`Default` and store alongside `endpoint`) or memoize via `OnceLock`.

data-plane/bindings/rust/src/client_config.rs
```diff
@@ -440,14 +433,15 @@ mod tests {
     use super::*;
     use crate::common_config::{CaSource, TlsSource};
     use crate::errors::SlimError;
+    #[allow(unused_imports)]
```
msardara: If this import is truly unused, delete it along with the `#[allow]`.

msardara (changes requested):
L almost GTM,  just a few minor fixes ✅

data-plane/bindings/rust/src/client_config.rs
hackeramitkumar: done

data-plane/core/config/src/server.rs
hackeramitkumar: done

data-plane/core/config/src/server.rs
hackeramitkumar: done

data-plane/core/config/src/client.rs
hackeramitkumar: Looked into caching, but it doesn't pay off here:

1. Display only formats the config at debug!(%config, …) in datapath/src/message_processing.rs:128 and grpc/server.rs:174 - both fire once per server bring-up. validate and the to_channel/builders are similarly one-shot at startup. No per-connection or per-message path calls resolved_transport().
2. Adding a OnceLock<TransportProtocol> field would mean #[serde(skip)] + #[schemars(skip)], dropping the PartialEq derive on both ClientConfig and ServerConfig to write it by hand (17 and 12 fields respectively), and updating every struct-literal construction site across core/config, control-plane, bindings, and slimctl to include the new field.
3. That's a lot of churn - and still leaves a staleness footgun on any cfg.endpoint = … write - for a split_once on a short string that runs 2–3 times per process at startup.

So I am thinking to keep it as it is.

msardara (approved):
LGTM ✅

PR #1667, opened 2026-05-26
perf(session): drop async_trait from MessageHandler and Transmitter

### LINKED ISSUE #1665: Drop `#[async_trait]` from session hot-path traits
### Problem

`MessageHandler` and `Transmitter` (in `core/session/src/traits.rs`) are currently declared with `#[async_trait]`. That macro rewrites every `async fn` into:

```rust
fn foo(&mut self, ...) -> Pin<Box<dyn Future<Output = T> + Send + '_>>
```

i.e. a heap allocation on every call.

Both traits sit on the per-message hot path of the session processing loop:

- MessageHandler::on_message is invoked once per session message in session_controller::process_messages (session_controller.rs:150-228).
- Transmitter::send_to_slim / send_to_app are invoked once per outbound message (and once per interceptor chain step).

At high message rates this is a measurable per-message allocation that buys us nothing — no dyn MessageHandler or dyn Transmitter exists anywhere in the codebase, so we don't actually need object safety.

### Proposal

Replace #[async_trait] with native Return-Position-Impl-Trait-In-Trait (RPITIT), available since Rust 1.75 and stable in Edition 2024 (already in use):

```rust
pub trait MessageHandler: Send + Sync {
    fn on_message(
        &mut self,
        message: SessionMessage,
    ) -> impl Future<Output = Result<(), SessionError>> + Send;
    // ...
}

Impls keep async fn syntax (clippy-clean, RPITIT-compatible under the hood):

impl MessageHandler for Foo {
    async fn on_message(&mut self, msg: SessionMessage) -> Result<(), SessionError> {
        // unchanged body
    }
}
```

Tradeoffs

- ✅ Zero allocation per call — the future is stack-allocated / inlined.
- ✅ async fn body syntax preserved, so impls read the same.
- ❌ Traits lose object safety. Not used today, but if dyn MessageHandler is ever needed, callers must opt-in with a BoxFuture wrapper.
- ⚠️  Recursive async (e.g. pop_task → process_control_message → pop_task in session_moderator.rs) must explicitly Box::pin at the recursion edge. async_trait was hiding this cost by
 boxing everything.

msardara:
# Description

Replace #[async_trait] with native RPITIT (`fn ... -> impl Future + Send`)
on both traits. async_trait rewrites every async method into
`fn(...) -> Pin<Box<dyn Future + Send>>`, allocating on every call.
These traits sit on the per-message hot path of the session processing
loop, so the allocation is real overhead.

No dyn MessageHandler / dyn Transmitter exists in the codebase, so
giving up object safety costs nothing. Recursive async in
session_moderator::pop_task is now bridged explicitly with Box::pin
at the recursion edge (previously hidden by async_trait's blanket
boxing).

Also fixes a clippy::let_unit_value in SessionTransmitter::send_to_app
that async_trait's desugaring had been masking.

Fixes: #1665 

## Type of Change

- [x] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/core/session/src/traits.rs
+++ b/data-plane/core/session/src/traits.rs
@@ -2,7 +2,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
 // Third-party crates
-use async_trait::async_trait;
 use slim_datapath::Status;
 use slim_datapath::api::ProtoMessage as Message;
 
@@ -17,7 +16,7 @@ pub enum ProcessingState {
 }
 
 /// Session transmitter trait
-#[async_trait]
+#[trait_variant::make(Send)]
 pub trait Transmitter: SessionInterceptorProvider {
     async fn send_to_slim(&self, message: Result<Message, Status>) -> Result<(), SessionError>;
 
@@ -29,8 +28,8 @@ pub trait Transmitter: SessionInterceptorProvider {
 ///
 /// Each layer implements this trait and can hold an inner layer.
 /// The layer decides whether to pass messages to its inner layer or handle them itself (or both).
-#[async_trait]
-pub trait MessageHandler: Send + Sync {
+#[trait_variant::make(Send)]
+pub trait MessageHandler {
     /// Init the layer.
     async fn init(&mut self) -> Result<(), SessionError>;
 
@@ -51,7 +50,7 @@ pub trait MessageHandler: Send + Sync {
         &mut self,
         _endpoint: &slim_datapath::api::Participant,
     ) -> Result<(), SessionError> {
-        Ok(())
+        async { Ok(()) }
     }
 
     /// Remove an endpoint from the session.
@@ -75,7 +74,7 @@ pub trait MessageHandler: Send + Sync {
     /// Optional hook for periodic ops (e.g. MLS key rotation)
     #[allow(dead_code)]
     async fn on_tick(&mut self) -> Result<(), SessionError> {
-        Ok(())
+        async { Ok(()) }
     }
 
     /// Get the current participants list (participants in the session)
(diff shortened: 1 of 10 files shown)


data-plane/core/session/src/traits.rs
```diff
@@ -47,11 +46,11 @@ pub trait MessageHandler: Send + Sync {
...
     /// Add an endpoint to the session.
     /// Default implementation does nothing for layers that don't manage endpoints.
-    async fn add_endpoint(
+    fn add_endpoint(
         &mut self,
         _endpoint: &slim_datapath::api::Participant,
-    ) -> Result<(), SessionError> {
-        Ok(())
+    ) -> impl std::future::Future<Output = Result<(), SessionError>> {
+        async { Ok(()) }
```
Tehsmash, line 53: Do we still need this updated with the train_variant?

data-plane/core/session/src/traits.rs
```diff
@@ -74,8 +73,8 @@ pub trait MessageHandler: Send + Sync {
 
     /// Optional hook for periodic ops (e.g. MLS key rotation)
     #[allow(dead_code)]
-    async fn on_tick(&mut self) -> Result<(), SessionError> {
-        Ok(())
+    fn on_tick(&mut self) -> impl std::future::Future<Output = Result<(), SessionError>> {
```
Tehsmash: ditto

data-plane/core/session/src/traits.rs
msardara: Yes and no:) Unfortunately the macro does not rewrite the default trait body, so we need to always return an `async { .. }`, even if the signature has the async keyword. We can do something like this:

```rust
async fn on_tick(&mut self) -> Result<(), SessionError> {
    async { Ok(()) }
}
```

It is a bit weird, but at least we don't have to manually write the signature.

Tehsmash (approved):

PR #1669, opened 2026-05-26
fix(session): apply backpressure on outbound channel send

Tehsmash:
# Description

`SessionTransmitter` and `AppTransmitter` both used `try_send` when forwarding messages to the outbound SLIM channel. `try_send` returns an error immediately when the bounded channel is full, which caused silent message drops whenever a sender produced messages faster than the SLIM internals could forward them.

This fix replaces `try_send` with `send().await` so that a full channel blocks the sending task instead of dropping the message. The sending task yields until capacity is available, providing natural backpressure through Tokio's async scheduler without any packet loss.

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/core/session/src/transmitter.rs
+++ b/data-plane/core/session/src/transmitter.rs
@@ -92,7 +92,8 @@ impl Transmitter for SessionTransmitter {
         }
 
         self.slim_tx
-            .try_send(message)
+            .send(message)
+            .await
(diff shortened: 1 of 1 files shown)


msardara (approved):
LGTM ✅

PR #1674, opened 2026-05-26
perf(session): optimise ProducerBuffer — swap field roles

Tehsmash:
## Description

`ProducerBuffer` is the fixed-capacity ring buffer used in the reliable-mode hot path to store messages for retransmission. The previous layout kept full `Message` objects in a `Vec<Option<Message>>` ring and message IDs in a `HashMap<usize, usize>` index.

The bottleneck in `push()` was the eviction step: to evict the oldest slot we had to read `buffer[self.next]` — a large protobuf struct unlikely to be cache-warm — just to extract its `.get_id()`.

This PR swaps the field roles:

```
// Before
buffer: Vec<Option<Message>>   — ring slots (large structs)
map:    HashMap<usize, usize>  — message_id → slot index

// After
messages: HashMap<usize, Message>  — message_id → Message
ring:     Vec<usize>               — compact ring of IDs (usize::MAX sentinel = empty)
```

The `ring` Vec is 512 × 8 B = 4 KB — fits entirely in L1. Eviction now reads a single integer instead of dereferencing a large struct.

**Per-operation impact:**

| Operation | Before | After |
|-----------|--------|-------|
| `push` eviction | Access `Vec<Option<Message>>[next]` → `.get_id()` | Read `Vec<usize>[next]` — integer, L1-hot |
| `get` | HashMap → Vec index → clone | HashMap → clone (one fewer indirection) |
| `clear` | Reallocates `vec![None; capacity]` | `HashMap::clear()` only, no allocation |
| `iter` | Vec scan with `Option` unwrap | Ring-ordered walk (cold flush path only) |

`iter()` is only called on the cold flush-on-connect path so its slightly more complex traversal has no hot-path impact.

All call sites in `session_sender.rs` use the same public API (`push`, `get`, `iter`, `clear`) — no changes needed outside `producer_buffer.rs`.

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [x] Other (please describe)

Performance optimisation — no behaviour change.

## Checklist

### DIFF
--- a/crates/session/src/producer_buffer.rs
+++ b/crates/session/src/producer_buffer.rs
@@ -70,8 +70,7 @@ impl ProducerBuffer {
     pub fn clear(&mut self) {
         self.messages.clear();
         self.next = 0;
-        // Stale IDs in `ring` are harmless: they are no longer in `messages` after clear(),
-        // so messages.remove is a no-op and messages.get returns None until overwritten.
+        self.ring.fill(usize::MAX);
     }
 
     pub fn get(&self, id: usize) -> Option<Message> {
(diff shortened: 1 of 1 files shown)


crates/session/src/producer_buffer.rs
```diff
@@ -44,56 +44,48 @@ impl ProducerBuffer {
...
-        self.buffer[self.next] = Some(msg);
-        // store the position of the message in the buffer
-        self.map.insert(id, self.next);
-        // increase the index to the next element in the buffer
+        // Store new message and record its ID in the ring.
+        self.messages.insert(id, msg);
+        self.ring[self.next] = id;
         self.next = (self.next + 1) % self.capacity;
         true
     }
```
micpapal, line 67: Here we can drop the return value as we always return true

micpapal (changes requested):
LGTM, with minor changes

crates/session/src/producer_buffer.rs
```diff
@@ -42,58 +42,48 @@ impl ProducerBuffer {
...
 
     /// Remove all the elements in the buffer
     pub fn clear(&mut self) {
-        self.buffer = vec![None; self.capacity];
+        self.messages.clear();
         self.next = 0;
-        self.map.clear();
+        // Stale IDs in `ring` are harmless: they are no longer in `messages` after clear(),
+        // so messages.remove is a no-op and messages.get returns None until overwritten.
```
micpapal: This may be dangerous. consider this example:

capacity=2:
1. push(id=1) → ring=[1, _], next=1
2. push(id=2) → ring=[1, 2], next=0
3. clear()    → messages={}, next=0, ring still [1, 2]
4. push(id=2) → evicts ring[0]=1, inserts id=2 at ring[0], next=1
5. push(id=3) → evicts ring[1]=2 (stale!) → removes msg id=2 that was just added

in our setting the message id should always increase so I don't think it will be a problem, but since this is not an operation done in the hot path it is safer to reset the ring vector. WDYT?

micpapal (changes requested):

crates/session/src/producer_buffer.rs
Tehsmash: Ah yeah, great catch! If the `if messages.contains_key(..)` check doesn't protect things in this case because we've cleared messages, so you could end up with an ID twice in the ring buffer!

I think this also has a implication on `iter` if you use .iter() after the 4. insert you could end up outputting message 2 twice.

micpapal: Yes I didn't notice this one

micpapal (approved):

PR #1677, opened 2026-05-27
feat: e2e header integrity validation

### LINKED ISSUE #1556: Add integrity check for packet header

markpmarton:
closes #1556 

# Description

In some cases checking the header integrity between nodes can't guarantee the header was not tempered with (e.g. in single-node deployments. To ensure the the inmutability of the  routing header fields, we need to implement an integrity check also on the client side.

# Changes
- replaced `enable_mls` flag with more complex `MlsSettings` struct (the field is optional: `None` is equal to `enable_mls = false` and `Some()` is equal to `enable_mls = true`.
- `MlsSettings` contains the percentage of messages to be validated (`header_integrity_validation_percent`).
- bench tests added testing the validation of 0%, 50% and 100% of the messages
- AAD generation is added to MLS message encryption `encrypt_output`
- header integrity check added to `decrypt_message` (only the set percentage of headers will be checked).
- dynamic port allocation added to servers running parallel in tests.

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


janosSarusiKis (approved):
lgtm

data-plane/core/datapath/src/api/gen/dataplane.proto.v1.rs
```diff
@@ -234,7 +234,8 @@ pub struct DiscoveryReplyPayload {}
 /// Join Request
 #[derive(Clone, PartialEq, Eq, Hash, ::prost::Message)]
 pub struct JoinRequestPayload {
-    /// true if mls is enabled
+    /// Deprecated: use mls_settings instead. Kept for buf wire compatibility.
+    #[deprecated]
```
micpapal: remove what is unneeded, we don't care about compatibility

data-plane/core/datapath/src/messages/utils.rs
```diff
@@ -1157,9 +1158,10 @@ impl CommandPayloadBuilder {
         };
 
         let payload = JoinRequestPayload {
-            enable_mls,
+            enable_mls: mls_settings.is_some(),
```
micpapal: same here please remove the deprecated code

data-plane/core/session/src/session_sender.rs
```diff
@@ -215,6 +230,22 @@ where
         session_header.set_session_type(self.session_type);
         message.get_slim_header_mut().set_fanout(fanout);
 
+        // Retrieve and set the identity token before building AAD / encrypting
+        let identity = self.tx.identity_provider.get_token()?;
```
micpapal: Why are we moving the encryption here? I think this is wrong. If we store already encrypted messages in the buffer and the group change, all the messages here will be useless as nobody can decrypt them. So I think this should be done later in the code not here. And the encryption of the retransmission should be handled in the same way. This will also remove the need of the mutex that was introduced in the MLSState.

data-plane/core/session/src/mls_state.rs
```diff
@@ -329,7 +393,7 @@ where
     V: Verifier + Send + Sync + Clone + 'static,
 {
     /// mls state in common between moderator and participants
-    pub(crate) common: MlsState<P, V>,
+    pub(crate) common: Arc<Mutex<MlsState<P, V>>>,
```
micpapal, line 406: This mutex here needs to be removed as it is too costly to grab it on every message. If we remove the packet encryption form the session sender I think we can come back to the previous setting.

micpapal (changes requested):

data-plane/core/session/src/mls_state.rs
markpmarton: I needed to move the encryption to the SessionSender because the AAD signature needs to be generated on the final version of the header values, and in federated routing and multicast delivery the destination field may be rewritten and it would invalidate the signature. There may be another solution (SessionSender keeps plain text message and sends on a channel to SessionModerator for encryption when needed), but I think its way more costly than the Mutex.

EDIT: another solution might be flattening the stack and move the SessionSession and Receiver as direct fields of SessionModerator. This way the MlsState can be kept in the SessionModerator and pass to the members as &mut. WDYT?

data-plane/core/session/src/session_sender.rs
markpmarton: The late-encryption is needed to ensure it will run only after all destination value changes in the header. You are right about the buffer, I wanted to avoid the additional encryption in retransmission but forgot about the epoch change. I'll move the encryption just before the timer creation. Thanks!

data-plane/core/datapath/src/messages/utils.rs
markpmarton: Done.

data-plane/core/datapath/src/api/gen/dataplane.proto.v1.rs
markpmarton: Buf was failing in the CI (non-deprecated proto fields can not be removed it seems), so I changed buf.yaml tamporarily. I'll create an issue to change it back after release.

data-plane/core/session/src/mls_state.rs
markpmarton: Encryption is moved to SessionTransmitter, this issue is resolved.

micpapal (approved):
LGTM, Thanks for all the modifications!

Issue #1555, opened 2026-04-22
Increase size of the ID in the SLIM name


micpapal:
## Description

SLIM names follow the format `org/ns/service/id`, where each component is currently 64 bits. The `id` field is derived by hashing the public key of a service. If we want the `id` to be **globally unique** (independent of the other name components), 64 bits is insufficient and should be increased to **128 bits**.

## Why 64-bit Is Not Enough

Using the birthday problem approximation, the probability of at least one collision among *n* services with a hash space of *2^b* bits is:

$$P(\text{collision}) \approx 1 - e^{-n^2 / (2 \cdot 2^b)}$$

Considering ~10¹⁰ devices (current internet scale) as a reference:

| ID size | P(collision) at 10¹⁰ services | Verdict |
|---|---|---|
| **64-bit** | **≈ 93.3%** — almost certain collision | ❌ Not suitable |
| **128-bit** | **≈ 1.47 × 10⁻¹⁹** (~1 in 6.8 quintillion) | ✅ Globally unique |

### Full collision probability table

**64-bit IDs** (*2^64 ≈ 1.84 × 10¹⁹*):

| Number of services (n) | Collision probability |
|---|---|
| 1,000,000 (1M) | ≈ 2.7 × 10⁻⁸ |
| 1,000,000,000 (1B) | ≈ 2.7 × 10⁻² (~2.7%) |
| 10,000,000,000 (10B) | ≈ 93.3% |

**128-bit IDs** (*2^128 ≈ 3.4 × 10³⁸*):

| Number of services (n) | Collision probability |
|---|---|
| 1,000,000 (1M) | ≈ 1.5 × 10⁻²⁷ |
| 1,000,000,000 (1B) | ≈ 1.5 × 10⁻²¹ |
| 10,000,000,000 (10B) | ≈ 1.47 × 10⁻¹⁹ |

128-bit provides UUID-level global uniqueness, which is the industry standard for this use case.

### Rust

Rust natively supports `u128` as a primitive type — no custom data structure needed

### Protobuf

Protobuf does not have a native 128-bit integer type. The recommended approach is to use a `bytes` field (16 bytes):

```protobuf
message Name {
  uint64 org = 1;
  uint64 ns = 2;
  uint64 service = 3;
  bytes id = 4;  // 16 bytes = 128 bits
}
```

Conversion between `bytes` and `u128` in Rust is straightforward:

```rust
// u128 → bytes (for serialization)
[...]

PR #1680, opened 2026-05-27
feat: increase Name ID from u64 to u128

micpapal:
# Description

 ### Summary
 
 This PR migrates the Name ID field across the entire SLIM stack from `u64` to `u128`, represented as UUID strings at API boundaries. This enables globally unique, collision-free agent identifiers.
 
 ### Changes
 
 **Proto definitions (`proto/`)**
 - `EncodedName.name_id`: changed from `uint64` to a `NameId` message containing two `uint64` fields (`id_0`, `id_1`) representing a u128
 - Unified `Name` message is now used everywhere — data plane, control plane, and slimctl messages all share the same `Name` type from `data-plane/v1/data_plane.proto`
 - `Route` message: consolidated from individual component fields (`component_0/1/2` + `id`) to `Name name`, `string link_id`, `Direction direction`
 - `RouteEntry` message: similarly restructured to use `Name` directly
 - Moved proto source-of-truth files to `proto/` root directory with symlinks for crate-local access
 - Import paths updated for buf module resolution
 
 **Data plane (`data-plane/core/datapath/`)**
 - `NameId`: new struct with `id_0`/`id_1` fields, u128 ↔ UUID roundtrip, reserved sentinel values (`NULL_COMPONENT`, `DATA_CHANNEL_ID`, `CONTROL_CHANNEL_ID`)
 - `ProtoName`: `with_id(u128)`, `string_id()`, `Display` outputs `org/ns/agent/<uuid>`
 - Subscription table: updated to match on u128 IDs
 - Added comprehensive unit tests for NameId and ProtoName
 
 **Control plane (`data-plane/control-plane/`)**
 - DB model: `component_id` column changed from `BIGINT` to `TEXT` (UUID string)
 - DB trait methods: `Option<i64>` → `Option<&str>`
 - Reconciler: builds `ProtoName` from DB fields, handles UUID parsing
 - Integration tests: use `match_prefix` for subscription comparison (ignoring runtime-assigned IDs)
 
 **Bindings**
 - Rust/Java/Go/C#/.NET: `App.id()` returns `String` instead of numeric type
 - Updated tests
 
 **CLI (`slimctl`)**
 - `parse_route`: accepts `org/ns/agent[/uuid]` format
 - Route list output: displays full name including UUID component
 - Route add: constructs `ProtoName` with UUID from parsed route string
 
 **Integration tests**
 - Go integration tests: updated regex to match UUID format in route list output
 - Control plane integration tests: prefix-based subscription matching
 
 ### Breaking Changes
 
 - Proto `Route` and `RouteEntry` messages are restructured (field numbers reused with different types)
 - All API consumers must update to use string-based IDs instead of numeric

ref: #1555 

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


data-plane/core/datapath/src/api/gen/dataplane.proto.v1.rs
```diff
@@ -114,15 +114,24 @@ pub struct Name {
...
     #[prost(uint64, tag = "1")]
     pub component_0: u64,
     #[prost(uint64, tag = "2")]
     pub component_1: u64,
     #[prost(uint64, tag = "3")]
     pub component_2: u64,
-    #[prost(uint64, tag = "4")]
-    pub component_3: u64,
+    #[prost(message, optional, tag = "4")]
+    pub name_id: ::core::option::Option<NameId>,
```
Tehsmash, line 134: I notice that this is now an optional because its a sub-message are we just going to work around that with validation? or can we make it required?

data-plane/core/datapath/src/api.rs
```diff
@@ -53,39 +54,103 @@ impl std::fmt::Display for ProtoName {
...
+        }
+    }
+
+    // parse Name Id from string as UUID format. Returns None if the string is not a valid UUID.
+    pub fn from_string(s: &str) -> Option<Self> {
+        // Try to parse as UUID first
+        if let Ok(uuid) = uuid::Uuid::parse_str(s) {
+            return Some(Self::new(uuid.as_u128()));
+        }
+        None
```
Tehsmash: should this be an error? Result(self, ...) it seems like it would be nice to be able to use `?` syntax instead of is_some() / match to None etc.

Also should we add support for the the reverse of id_to_string? i.e. 

```
if s == "NULL_COMPONENT" {
	return Self::NULL_COMPONENT
}
```

data-plane/core/datapath/src/api.rs
```diff
@@ -53,39 +54,103 @@ impl std::fmt::Display for ProtoName {
...
+
+    pub fn new(id: u128) -> Self {
+        NameId {
+            id_0: (id >> 64) as u64,
+            id_1: (id & 0xFFFFFFFFFFFFFFFF) as u64,
+        }
+    }
+
+    // parse Name Id from string as UUID format. Returns None if the string is not a valid UUID.
+    pub fn from_string(s: &str) -> Option<Self> {
```
Tehsmash: can we put this into a "From" trait and id_to_string into an "Into" trait?

data-plane/core/datapath/src/api.rs
```diff
@@ -53,39 +54,103 @@ impl std::fmt::Display for ProtoName {
...
+            id => uuid::Uuid::from_u128(id).to_string(),
+        }
+    }
+
+    /// Returns true if `id` is one of the reserved values
+    /// Notice that u128::MAX - 1 is not used at the moment
+    /// and it reserved for future use.
+    pub const fn is_reserved_id(id: u128) -> bool {
         id >= Self::CONTROL_CHANNEL_ID
     }
```
Tehsmash: Should we create a larger reserved block to protect ourselves in the future? i.e. id >= Self::NULL_COMPONENT - 50 for example?

data-plane/core/datapath/src/api.rs
```diff
@@ -53,39 +54,103 @@ impl std::fmt::Display for ProtoName {
...
     // CONTROL_CHANNEL_ID is the id for the control channel name.
-    pub const CONTROL_CHANNEL_ID: u64 = u64::MAX - 3; // ends with 0xfc (control)
+    pub const CONTROL_CHANNEL_ID: u128 = u128::MAX - 3; // ends with 0xfc (control))
+
+    pub fn new(id: u128) -> Self {
+        NameId {
+            id_0: (id >> 64) as u64,
+            id_1: (id & 0xFFFFFFFFFFFFFFFF) as u64,
+        }
+    }
```
Tehsmash, line 89: Can we create "From"/"Into" traits like:

```
impl Into<NameId> for u128 {
	fn into(self) -> NameId {
    	NameId.new(...)
    }
}
```

This should allow us to do things like `<u128>.into().to_string()`

data-plane/core/datapath/src/api.rs
```diff
@@ -53,39 +54,103 @@ impl std::fmt::Display for ProtoName {
...
+    pub fn id(&self) -> u128 {
+        self.name_id
+            .as_ref()
+            .map_or(NameId::NULL_COMPONENT, |nid| nid.id())
+    }
+
+    /// Returns the ID as a human-readable string.
+    pub fn string_id(&self) -> String {
+        NameId::id_to_string(self.id())
+    }
```
Tehsmash, line 142: With the above trait suggestions I think we can avoid both of these functions potentially. For example you can just do: `protoName.EncodedName.name_id.into()` to get the u128

Tehsmash (changes requested):

data-plane/core/datapath/src/api/gen/dataplane.proto.v1.rs
micpapal: That is the compilation of the protobuf that makes these fields optional. At the moment the semantic is that if the NameId is None we return NULL_COMPONENT as an ID. There is a check when you try to get the id form the name.

data-plane/core/datapath/src/api.rs
micpapal: I put the error because this will surface up to the bindings, but I can modify this.

data-plane/core/datapath/src/api.rs
```diff
@@ -48,44 +50,99 @@ pub use proto::dataplane::v1::message::MessageType::Subscribe as ProtoSubscribeT
...
+        }
+    }
+}
+
+impl EncodedName {
+    /// Returns the u128 ID from the embedded `NameId`, or `NULL_COMPONENT` if absent.
+    pub fn id(&self) -> u128 {
+        self.name_id
+            .as_ref()
+            .map_or(NameId::NULL_COMPONENT, |nid| (*nid).into())
```
Tehsmash, line 134: Can we add a None case into the From? I'm not 100% sure but will rust allow:

```
impl From<Option<NameId>> for u128 {
	fn from(name_id: Option<NameId>) -> String {
		match ...
	}
}
```

micpapal: I tried and you can't do that because of the orphan rule https://doc.rust-lang.org/reference/items/implementations.html#orphan-rules
Using ``Option<NameId>`` instead of ``NameId``, neither the from nor the target type are defined in the local crate

Tehsmash: ah thats annoying that Option isn't counted as a fundamental type but things like Box<T> and Pin<T> are :(

Tehsmash (approved):

PR #1689, opened 2026-05-28
refactor(datapath): replace is_local bool with ConnType enum

### LINKED ISSUE #1553: Support for SLIM replicas
## Description

When SLIM is deployed in a Kubernetes cluster with multiple replicas, each instance currently connects to the controller as an independent node. Route distribution between replicas goes entirely through the controller. This has several limitations.

## Current Limitations

 1. Controller dependency — Even a single-deployment scenario requires a controller for replicas to exchange routes.
 2. Reconnection penalty — If a service disconnects from one SLIM replica and reconnects to another, all route updates must 
round-trip through the controller before the new replica is aware of the subscriptions.
 3. Scalability overhead — The controller must manage N separate nodes and distribute routes individually to each replica, even 
though they belong to the same logical deployment.

## Proposed Solution

Allow replicas within the same Kubernetes deployment to communicate directly and exchange local routes among themselves (peer-to-peer within the replica set). This brings two key improvements:

 1. Intra-replica route sync — Replicas share their local subscriptions with each other directly. Any replica is immediately aware
 of routes held by its peers, so a service can connect to any replica and messages will be delivered to the correct destination.
 2. Unified view from the controller — The controller sees all replicas as a single logical SLIM node. Route updates to/from the 
controller are handled once for the entire replica set rather than per-instance, reducing traffic and simplifying state 
management.

## Benefits

 - No controller needed for single-deployment setups — Replicas synchronize routes locally without external coordination.
 - Seamless service mobility — A service can disconnect from one replica and reconnect to another without routing disruption; all 
replicas already know the full route set.
 - Reduced controller load — The controller handles one aggregated node instead of N individual ones.

msardara:
# Description

Replace the is_local: bool parameter throughout the datapath with a
ConnType enum (Local, Remote, Peer). This provides type-safe
connection categorization and prepares the subscription table for
peer-aware routing.

Key changes:
  - Add ConnType enum in tables.rs (Local, Remote, Peer)
  - Add MatchFilter struct for source-aware publish matching
  - Update SubscriptionTable trait: add_subscription,
     remove_subscription, remove_connection, match_one,
     match_all all use ConnType/MatchFilter
  - PrefixEntry supports 3-way categorization (local/remote/peer ConnLists)
  - Forwarder API updated to use ConnType and MatchFilter
  - process_stream/handle_new_message/process_message chain uses ConnType
  - Remove connection::Type enum, use ConnType directly in Connection
  - for_each callback yields peer connections as additional parameter

Part of: #1553 

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/core/datapath/src/connection.rs
+++ b/data-plane/core/datapath/src/connection.rs
@@ -31,19 +31,7 @@ pub(crate) enum Channel {
     Client(mpsc::Sender<Message>),
 }
 
-/// Connection type
-#[derive(Debug, Clone, Default)]
-pub(crate) enum Type {
-    /// Connection with local application
-    Local,
-
-    /// Connection with remote slim instance
-    Remote,
-
-    /// Unknown connection type
-    #[default]
-    Unknown,
-}
+use crate::tables::ConnType;
 
 #[derive(Clone)]
 /// Connection information
@@ -61,7 +49,7 @@ pub struct Connection {
     config_data: Option<ClientConfig>,
 
     /// Connection type
-    connection_type: Type,
+    connection_type: ConnType,
 
     /// cancellation token to stop the receiving loop on this connection
     cancellation_token: Option<CancellationToken>,
@@ -90,7 +78,7 @@ impl std::fmt::Debug for Connection {
 /// Implementation of Connection
 impl Connection {
     /// Create a new Connection
-    pub(crate) fn new(connection_type: Type, channel: Channel) -> Self {
+    pub(crate) fn new(connection_type: ConnType, channel: Channel) -> Self {
         Self {
             remote_addr: None,
             local_addr: None,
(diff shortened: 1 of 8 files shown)


data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -129,94 +128,78 @@ impl PrefixEntry {
...
-            self.remote[idx].push(conn);
-            if !self.all_remote_conns.contains(&conn) {
-                self.all_remote_conns.push(conn);
-            }
-        }
-    }
+    /// Insert or remove `conn` from the slot at `idx` for the given category.
+    /// When `insert` is true, adds the connection; otherwise removes it.
+    /// Aggregate lists are kept in sync automatically.
+    fn update_conn(&mut self, idx: usize, conn: u64, category: ConnType, insert: bool) {
```
micpapal: For readability I would keep insert_conn and remove_conn separated. In the end they have only one line in common that will be repeted

data-plane/core/datapath/src/tables/subscription_table.rs
```diff
@@ -1303,44 +1463,47 @@ mod tests {
...
         assert!(result.contains(&4), "Should match ID 2 subscription");
 
         // Specific ID 2 message should still work
-        let result = t.match_one(&enc(&name_id2), 100).unwrap();
+        let result = t.match_one(&enc(&name_id2), 100, MatchFilter::ALL).unwrap();
         assert_eq!(result, 4);
 
         // But specific ID 1 should still fail (was removed earlier)
-        let err = t.match_one(&enc(&name_id1), 100);
+        let err = t.match_one(&enc(&name_id1), 100, MatchFilter::ALL);
```
micpapal, line 1495: Probably we need at least one test for MatchFilter::EXCLUDE_PEER

data-plane/core/datapath/src/connection.rs
```diff
@@ -173,13 +161,24 @@ impl Connection {
 
     /// Get the connection type
     #[allow(dead_code)]
-    pub(crate) fn connection_type(&self) -> &Type {
+    pub(crate) fn connection_type(&self) -> &ConnType {
```
micpapal: this is now replaced by conn_category, we can remove it I think

micpapal (changes requested):

micpapal (approved):

PR #1696, opened 2026-06-01
feat(datapath): add peer discovery module with static backend

### LINKED ISSUE #1553: Support for SLIM replicas

msardara:
# Description

Add the peer discovery abstraction for SLIM replica route synchronization:

- PeerDiscovery trait (start/recv) with #[trait_variant::make(Send)]
- PeerEvent enum (Joined/Left) and PeerInfo struct
- PeerConfig with peer_group and optional dynamic discovery config
- StaticPeerDiscovery implementation (emits Joined for provided peers)
- Add ConnectionType (remote/peer) to ClientConfig in core/config
- Add peers: Option<PeerConfig> to ServiceConfiguration
- Static peers are configured as dataplane.clients with connection_type: peer

Part of: #1553 

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/data-plane/bindings/rust/src/client_config.rs
+++ b/data-plane/bindings/rust/src/client_config.rs
@@ -346,6 +346,7 @@ impl From<ClientConfig> for CoreClientConfig {
             require_header_mac: config
                 .require_header_mac
                 .unwrap_or(core_defaults.require_header_mac),
+            connection_type: core_defaults.connection_type,
         }
     }
 }
--- a/data-plane/core/config/src/client.rs
+++ b/data-plane/core/config/src/client.rs
@@ -28,6 +28,55 @@ use crate::auth::static_jwt::Config as BearerAuthenticationConfig;
 use crate::backoff::Strategy;
 use crate::backoff::exponential::Config as ExponentialBackoff;
 use crate::backoff::fixedinterval::Config as FixedIntervalBackoff;
+
+/// The type of connection a client establishes.
+///
+/// Determines how the datapath treats messages on this connection
+/// (routing rules, subscription forwarding, etc.).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default, Serialize, Deserialize, JsonSchema)]
+#[serde(rename_all = "lowercase")]
+pub enum ConnType {
+    /// Connection with a local application (agent).
+    /// Not exposed in configuration — only used internally.
+    #[serde(skip)]
+    Local,
+    /// Connection with a remote SLIM instance (other deployment, via controller).
+    #[default]
+    Remote,
+    /// Connection with a peer replica in the same deployment.
+    Peer,
+}
+
+impl ConnType {
+    /// All variants, for iteration.
+    pub const ALL: [ConnType; 3] = [ConnType::Local, ConnType::Remote, ConnType::Peer];
+
+    /// Number of ConnType variants (derived automatically).
+    pub const COUNT: usize = Self::ALL.len();
+
+    /// Index for array-based storage. Stable mapping.
+    pub const fn index(self) -> usize {
+        match self {
+            ConnType::Local => 0,
+            ConnType::Remote => 1,
+            ConnType::Peer => 2,
+        }
+    }
+
+    /// Converts from the legacy `is_local` boolean for backward compatibility.
+    pub fn from_is_local(is_local: bool) -> Self {
+        if is_local {
+            ConnType::Local
+        } else {
+            ConnType::Remote
+        }
+    }
+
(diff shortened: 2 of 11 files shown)


data-plane/core/config/src/client.rs
```diff
@@ -28,6 +28,20 @@ use crate::auth::static_jwt::Config as BearerAuthenticationConfig;
...
 use crate::backoff::exponential::Config as ExponentialBackoff;
 use crate::backoff::fixedinterval::Config as FixedIntervalBackoff;
+
+/// The type of connection a client establishes.
+///
+/// Determines how the datapath treats messages on this connection
+/// (routing rules, subscription forwarding, etc.).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default, Serialize, Deserialize, JsonSchema)]
+#[serde(rename_all = "lowercase")]
+pub enum ConnectionType {
```
micpapal: Is there a reason to do not reuse the ConnType in tables? There we have also local that here is not used but I think we can reuse the same enum.

data-plane/bindings/rust/src/client_config.rs
```diff
@@ -346,6 +346,7 @@ impl From<ClientConfig> for CoreClientConfig {
             require_header_mac: config
                 .require_header_mac
                 .unwrap_or(core_defaults.require_header_mac),
+            connection_type: core_defaults.connection_type,
```
micpapal, line 349: Not sure if this is a bug. Here connection_type is always set to remote. is this wanted?

data-plane/core/datapath/src/peer_discovery/static_list.rs
```diff
@@ -0,0 +1,134 @@
...
+
+        Ok(())
+    }
+
+    async fn recv(&mut self) -> Result<PeerEvent, PeerDiscoveryError> {
+        match self.pending.pop_front() {
+            Some(event) => Ok(event),
+            // Static discovery never produces more events after initial list.
+            // Block forever (the peer sync manager will drop the discovery on shutdown).
+            None => std::future::pending().await,
```
micpapal, line 57: here is it possible to just return instead of block forever? this function will never be called anyway

micpapal (changes requested):

data-plane/core/config/src/client.rs
msardara: Yes we can, let me fix it

data-plane/bindings/rust/src/client_config.rs
msardara: Yes, setting it to remote we get the same behavior of now, without the peers

data-plane/core/datapath/src/peer_discovery/static_list.rs
msardara: I think blocking is the right thing here, as for instance the k8s discovery will block and send a new peer when there is another replica.

In this case it is static, so no more replicas will come, but the caller does not know it. Returning means handling it in a different way at the caller depending on the implementation.

micpapal (approved):

PR #1702, opened 2026-06-02
refactor(session): remove SessionTransmitter, return SessionOutput from layers

### LINKED ISSUE #1703: refactor(session): Return SessionOutput from layers instead of sending internally
## Summary

Refactor the session layer architecture from a "push" model (layers internally send messages via channels) to a "pull" model (layers return outbound messages as SessionOutput, and the processing loop handles all sends from a **single** dispatch point).

## Motivation

Currently, helper methods deep in the moderator/participant call chain send messages directly to SLIM or the application via channel transmitters. This makes it difficult to:

1. Reason about message flow — sends are scattered across many methods
2. Intercept or transform outbound messages in a single place (e.g., for peer forwarding in a multi-replica deployment)
3. Test message output without wiring up channel receivers

## Proposed Changes

- SessionOutput should have static constructors (`to_slim()`, `to_app()`) and push methods (`push_slim()`, `push_app()`) backed by `SmallVec<[OutboundMessage; 2]>`
- Every helper method in SessionModerator and SessionParticipant should return `Result<SessionOutput, SessionError>` instead of `Result<(), SessionError>`
- `SessionControllerCommon` should no longer accumulate output in a struct field — remove `send_to_slim`, `send_to_app`, `handle_output`, `take_output` methods
- `send_with_timer` / `send_control_message` should return `SessionOutput` directly
- `on_message` in moderator/participant should create a local `SessionOutput`, extend it from each helper call, and return it to the processing loop
- `dispatch_output` in the processing loop should be the single point where identity is set and messages are sent to SLIM or the app channel
- MessageHandler trait should be simplified: no generics, no MLS parameter
- MLS encryption should be consolidated in a single `encrypt_output()` method per layer
- Remove the `SessionTransmitter` generic parameter from Session and ControllerSender

Expected Benefits

- Single dispatch point for all outbound messages (processing loop)
- Layers become pure functions: input message → output messages
- Easier to add cross-cutting concerns (e.g., peer forwarding, metrics, logging) at the dispatch point
- Simpler testing — check returned output directly without channel receivers

msardara:
# Description

Refactor the session layer architecture so that inner layers (Session,
SessionSender, SessionReceiver, ControllerSender) return outbound
messages via a SessionOutput type instead of calling transmitter methods
internally. The processing loop dispatches all outbound messages through
a centralized dispatch_output helper.

Key changes:
- Define SessionOutput (SmallVec-backed) and OutboundMessage types
- MessageHandler trait simplified: no generics, no MLS parameter, returns
  SessionOutput instead of Result<(), SessionError>
- Remove generic P and SessionTransmitter from Session, ControllerSender
- MLS encryption/decryption owned by moderator/participant layers
- Processing loop uses dispatch_output for identity-setting and sends
- Deduplicate MsgAck/RtxRequest handling in Session::on_message

Fixes: #1703 

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


data-plane/core/session/src/session.rs
```diff
@@ -134,21 +124,22 @@ where
...
+                Ok(SessionOutput::new())
             }
             _ => Err(SessionError::SessionMessageInternalUnexpected(Box::new(
                 message,
             ))),
         }
     }
 
-    pub async fn add_endpoint(&mut self, endpoint: &Participant) -> Result<(), SessionError> {
+    pub fn add_endpoint(&mut self, endpoint: &Participant) -> Result<SessionOutput, SessionError> {
```
micpapal, line 135: This function will never send a packet. is there a way to return just Ok(). It is not a big problem but it is quite misleading to see SessionOutput as result the should contain messages.

data-plane/core/session/src/session_moderator.rs
```diff
@@ -251,16 +256,24 @@ where
...
+                return Err(SessionError::SessionMessageInternalUnexpected(Box::new(
+                    message,
+                )));
             }
-            _ => Err(SessionError::SessionMessageInternalUnexpected(Box::new(
-                message,
-            ))),
         }
+
+        self.encrypt_output(&mut output)?;
```
micpapal, line 268: here it looks like that all the messages are encrypted, also the control messages generated at line 140. instead we should encrypt only application messages. Are they filtered later?

data-plane/core/session/src/session_participant.rs
```diff
@@ -191,45 +207,55 @@ where
...
+                self.pending_leave_cleanup = false;
+            }
+            _ => {
+                return Err(SessionError::SessionMessageInternalUnexpected(Box::new(
+                    message,
+                )));
+            }
         }
+
+        self.encrypt_output(&mut output)?;
```
micpapal, line 250: Same problem as for the moderator here.

data-plane/core/session/src/session_receiver.rs
```diff
@@ -81,18 +76,12 @@ where
...
     session_type: ProtoSessionType,
 
-    /// send to slim/app
-    tx: SessionTransmitter<P>,
-
     /// drain state - when true, no new messages from app are accepted
     draining_state: ReceiverDrainStatus,
 }
 
 #[allow(dead_code)]
```
micpapal: this is probably a left over of some prev version I think we can remove it now as the SessionReceiver is used

data-plane/core/session/src/session_sender.rs
```diff
@@ -39,10 +37,7 @@ struct GroupTimer {
 }
 
 #[allow(dead_code)]
```
micpapal: same for this allow dead_code here

data-plane/core/session/src/session_receiver.rs
```diff
@@ -54,10 +52,7 @@ enum ReceiverDrainStatus {
...
```
micpapal: also this allow dead_code can be removed

data-plane/core/session/src/session_sender.rs
```diff
@@ -90,17 +82,13 @@ where
...
```
micpapal: remove also this one

data-plane/core/session/src/session_controller.rs
```diff
@@ -129,6 +140,35 @@ impl SessionController {
...
+        P: slim_auth::traits::TokenProvider + Send + Sync + Clone + 'static,
+        V: slim_auth::traits::Verifier + Send + Sync + Clone + 'static,
+        M: crate::subscription_manager::SubscriptionOps,
+    {
+        for msg in output.messages {
+            match msg {
+                OutboundMessage::ToSlim(mut message) => {
+                    // Set identity on outbound SLIM messages
+                    if let Ok(identity) = settings.identity_provider.get_token() {
+                        message.get_slim_header_mut().set_identity(identity);
```
micpapal: here if get_token returns an error we send the message to slim anyway instead I think we should drop it.

micpapal (changes requested):

data-plane/core/session/src/session_moderator.rs
msardara: Yes they are filtered in mls_state in the `should_process_message` function. I also added the case for `RtxReply` there.

data-plane/core/session/src/session_participant.rs
msardara: Ditto

data-plane/core/session/src/session.rs
msardara: There is a case where we need to return messages there, I'd leave it as it is for now.

micpapal (approved):

PR #1705, opened 2026-06-03
feat(peer-sync): subscription synchronization

msardara:
# Description

Implements peer-to-peer subscription synchronization
within a deployment. Replicas discover each other, establish
direct connections, and exchange local routes so that any replica
is immediately aware of subscriptions held by its peers.

## Motivation

When SLIM is deployed with multiple replicas in Kubernetes,
each instance currently connects to the controller as an independent
node. Route distribution between replicas goes entirely through the
controller. This has several limitations:
 - Controller dependency — Even single-deployment scenarios
    require a controller for replicas to exchange routes
 - Reconnection penalty — If a service reconnects to a different
    replica, all route updates must round-trip through the controller
 - Scalability overhead — The controller manages N separate nodes
    instead of one logical deployment

### Peer Sync Manager (core/datapath/src/peer_sync/)

 - Event loop coordinating discovery events with subscription
    forwarding
 - Deterministic tie-breaking (lower ID dials) to ensure exactly
    one connection per peer pair
 - PeerState for bidirectional peer_id ↔ conn_id tracking
 - Full sync on peer join (snapshot of all local subscriptions) +
   incremental broadcast

### Link Negotiation

 - node_id and peer_group exchanged during negotiation
 - Server-side peer detection and connection type upgrade
 - Server rejects peer upgrade when peer_group doesn't match

### Configurable Topology

 - full_mesh (default): all replicas connect to each other
 - hub_and_spoke: only hub (smallest lexicographic ID) dials out,
    relays subscriptions between spokes

### Integration Tests (core/service/tests/peer_sync_test.rs)

 - Subscription propagation between peers
 - Subscription isolation from remote connections
 - Hub-and-spoke relay and message delivery
 - Full mesh message delivery

Part of: #1553 

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


proto/controller/v1/controller.proto
```diff
@@ -104,6 +105,8 @@ message ConnectionEntry {
   string config_data = 3;
   optional string link_id = 4;
   ConnectionDirection direction = 5;
+  // Remote peer's node identifier (set after link negotiation).
+  optional string peer_node_id = 6;
```
micpapal, line 109: this is the id of the node in the remote deployment?

data-plane/config/full-mesh/replica-1.yaml
```diff
@@ -0,0 +1,46 @@
...
+        - endpoint: "0.0.0.0:46357"
+          tls:
+            insecure: true
+    controller:
+      servers:
+        - endpoint: "0.0.0.0:47357"
+          tls:
+            insecure: true
+    peers:
+      peer_group: "thegroup"
```
micpapal: should we use peer_group or deployment name?

data-plane/config/hub-and-spoke/replica-1.yaml
```diff
@@ -0,0 +1,46 @@
...
+          tls:
+            insecure: true
+    controller:
+      servers:
+        - endpoint: "0.0.0.0:47357"
+          tls:
+            insecure: true
+    peers:
+      peer_group: "thegroup"
+      topology: "hub_and_spoke"
```
micpapal, line 29: how do you specify the hub in the topology?

data-plane/core/datapath/src/sync/forwarder.rs
```diff
@@ -0,0 +1,710 @@
...
+            async move { (conn_id, mp.send_msg(msg, conn_id).await) }
+        }))
+        .await;
+
+        // Count successes; log failures.
+        let mut sent_count = 0usize;
+        for (conn_id, result) in &send_results {
+            if let Err(e) = result {
+                let peer = self.peer_label(*conn_id);
+                warn!(%conn_id, %peer, error = %e, "failed to send to peer");
```
micpapal, line 744: here if we always get an error sent_count will be 0 and we return Ok() at line 524. That should be an error

data-plane/core/datapath/src/sync/manager.rs
```diff
@@ -0,0 +1,348 @@
...
+            }
+        }
+    }
+
+    /// Handle a newly discovered peer.
+    ///
+    /// Connection behavior depends on topology:
+    /// - **FullMesh**: only the node with the lexicographically smaller self_id
+    ///   initiates the outbound connection (tie-breaking for deduplication).
+    /// - **HubAndSpoke**: only the hub (smallest ID) initiates connections.
```
micpapal: for this it may be useful to give a way to select the hub in the config

data-plane/core/datapath/src/peer_discovery/config.rs
```diff
@@ -3,35 +3,93 @@
...
+    /// Topology for peer connections. Defaults to `FullMesh`.
+    #[serde(default)]
+    pub topology: PeerTopology,
+
+    /// Static list of peer connections. Each entry requires a `node_id` plus
+    /// the connection configuration fields (flattened from `ClientConfig`).
+    /// The `connection_type` field is forced to `Peer` regardless of what is set.
+    /// Must contain at least one entry.
+    #[serde(deserialize_with = "deserialize_non_empty_peers")]
+    pub static_peers: Vec<StaticPeerEntry>,
```
micpapal, line 73: in k8s if we start with a single node that is scaled up by the system this list may be empty as no peer is set at config time. So I think it may be empty or optional if only one node is specified

data-plane/core/datapath/src/sync/forwarder.rs
```diff
@@ -0,0 +1,965 @@
...
+                node_id,
+                PeerEntry {
+                    conn_id,
+                    endpoint: String::new(),
+                    is_outgoing: false,
+                },
+            );
+        }
+
+        self.add_peer_conn(conn_id);
```
micpapal: here you can directly call add_peer_conn_and_sync

data-plane/core/datapath/src/sync/forwarder.rs
```diff
@@ -0,0 +1,965 @@
...
+        self.inner.seen_sub_ids.write().insert(sub_id);
+        self.inner
+            .forwarded_sub_for_name
+            .write()
+            .insert(name.clone(), sub_id);
+    }
+
+    /// Remove a subscription from the seen set (on unsubscribe).
+    pub fn remove_forwarded_sub(&self, name: &ProtoName, sub_id: u64) {
+        self.inner.seen_sub_ids.write().remove(&sub_id);
```
micpapal, line 503: is it always safe to remove the sud_id from this list? if a new subscription is created an removed immediately, is it possible the subscription is still in flight and re-added or forwarded form some of the node as it is not anymore in the set. I am not sure this can happen

data-plane/core/datapath/src/sync/remote.rs
```diff
@@ -0,0 +1,191 @@
...
+    subscription_id: u64,
+}
+
+impl SubscriptionInfo {
+    #[cfg(test)]
+    pub(crate) fn new(
+        source: ProtoName,
+        name: ProtoName,
+        source_identity: String,
+        _conn: u64,
```
micpapal, line 36: any reason to have conn here if not used?

data-plane/core/datapath/src/sync/remote.rs
```diff
@@ -0,0 +1,191 @@
...
+/// - [`restore`](Self::restore): replay subscriptions after reconnect
+///
+/// Also owns the [`RecoveryTable`] for server-side route preservation during brief disconnects.
+#[derive(Debug)]
+pub struct RemoteSync {
+    /// Subscriptions forwarded to each remote connection, keyed by conn_id.
+    table: RwLock<HashMap<u64, HashSet<SubscriptionInfo>>>,
+
+    /// Server-side route preservation during brief disconnects (keyed by link_id).
+    pub(crate) recovery: RecoveryTable,
```
micpapal: this can now be removed

data-plane/core/datapath/src/message_processing.rs
```diff
@@ -898,145 +928,273 @@ impl MessageProcessor {
...
+
+        info!(%in_connection, %link_id, "recovering routes for reconnected peer");
+
+        // Re-add local routing entries under the new connection index.
+        for (name, sub_ids) in &entry.local_subs {
+            for &subscription_id in sub_ids {
+                if let Err(e) = self.forwarder().on_subscription_msg(
+                    name.clone(),
+                    in_connection,
+                    ConnType::Remote,
```
micpapal: is this connection type correct?

data-plane/core/datapath/src/message_processing.rs
```diff
@@ -1225,45 +1365,113 @@ impl MessageProcessor {
...
+            && sub_id != 0
+            && self.peer_sync().has_seen_sub_id(sub_id)
+        {
+            debug!(
+                %in_conn,
+                %sub_id,
+                "dropping subscription already forwarded by this node (loop prevention)"
+            );
+            if let Some(id) = subscription_id {
+                self.send_subscription_ack(in_connection, id, &Ok(())).await;
```
micpapal, line 1062: if we saw already the subscription id, it means that we also sent already the ack back. Why do we send it again?

data-plane/core/datapath/src/sync.rs
```diff
@@ -0,0 +1,90 @@
...
+use crate::api::proto::dataplane::v1::Message;
+use crate::api::{ProtoMessage, ProtoName};
+use crate::messages::utils::MessageError;
+
+/// Monotonically increasing counter for generating unique sync subscription IDs.
+/// These IDs are distinct from application-level subscription IDs.
+static SYNC_SUB_ID: AtomicU64 = AtomicU64::new(1_000_000);
+
+/// Generate a unique subscription ID for sync messages.
+pub fn next_subscription_id() -> u64 {
```
micpapal: is this used only for the test at line 70? if yes we can move it to the test section

data-plane/core/service/src/service.rs
```diff
@@ -287,6 +352,54 @@ impl Service {
         Ok(())
     }
 
+    /// Start the peer sync manager in a background task.
+    fn start_peer_sync(&self, peer_config: &PeerConfig) {
+        let self_id = self.config.node_id.clone();
+
+        // Determine if we are the hub (smallest node_id among all peers including self).
+        let is_hub = peer_config
```
micpapal, line 363: as I wrote in the config file I think is better to specify directly who is the hub

micpapal (changes requested):

data-plane/core/datapath/src/peer_discovery/config.rs
msardara: Yess, for k8s I modified the conf file format in another PR: https://github.com/agntcy/slim/pull/1724

data-plane/core/datapath/src/sync/remote.rs
msardara: Yes. I will do it in a follow up PR

data-plane/core/datapath/src/message_processing.rs
msardara: We do it because the subscription may come from a different peer

data-plane/core/datapath/src/message_processing.rs
msardara: This part will be removed in a follow up PR, we can ignore it for now

data-plane/config/hub-and-spoke/replica-1.yaml
msardara: For the moment we don't specify it, it is auto selected. Maybe we can add the explicit configuration in a follow up pr

data-plane/core/datapath/src/sync/forwarder.rs
msardara: I think it is fine, as the unsubscribe will delete the subscribe sent before.

micpapal (approved):

PR #1724, opened 2026-06-09
feat: automatic peer discovery in k8s

msardara:
# Description

Implement automatic peer discovery in k8s.

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


charts/slim/values.yaml
```diff
@@ -18,6 +18,14 @@ slim:
   # This is for overriding the default config below
   overrideConfig:
 
+  # Peer configuration for multi-replica deployments.
+  # Automatically enabled when replicaCount > 1.
+  peers:
+    # Shared group identity for mutual peer authentication.
+    deploymentName: "slim-peers"
+    # Topology for peer connections: "full_mesh" or "hub_and_spoke"
+    topology: "full_mesh"
```
micpapal: do need to add a comment here that the only supported topology using the k8s discovery is full_mesh?

micpapal: are we missing the discovery type here or is set automatically?

data-plane/core/datapath/src/peer_discovery/kubernetes.rs
```diff
@@ -0,0 +1,658 @@
...
+                            }
+                            WatcherEvent::InitDone => {
+                                // Reconcile: emit Left for peers that were known but
+                                // not seen during this re-list.
+                                if let Some(seen) = init_seen.take() {
+                                    let stale: Vec<String> = known_peers
+                                        .keys()
+                                        .filter(|k| !seen.contains_key(*k))
+                                        .cloned()
+                                        .collect();
```
micpapal, line 279: when does it happen that you can have a peer in known_peers but is not in seen?

charts/slim/values.yaml
msardara: Yes we should add a comment ✅

msardara: It is constructed by helm when `replicas > 1`. However I am thinking that just have it always set is better

data-plane/core/datapath/src/peer_discovery/kubernetes.rs
msardara: It can happen if the watcher temporarily disconnects from the apiserver and in that moment a peer is deleted. It is a super corner case:)

charts/slim/templates/deployment.yaml
```diff
@@ -107,14 +123,25 @@ spec:
...
+        {{- end }}
+        {{- with .Values.affinity }}
         {{- toYaml . | nindent 8 }}
+        {{- end }}
       {{- end }}
       {{- with .Values.tolerations }}
       tolerations:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- if .Values.slim.onePerNode }}
```
Tehsmash: I'm not sure its a good idea to mix one per node with deployment, there is no clean way from the helm side to set the replicas to the number of nodes and you also can't scale this if a new node is added.

Instead I think it would be better to offer both options in the chart, a Daemonset mode "one per node" and a Deployment mode. The deployment mode should probably still have pod anti-affinity as the default but we can make it a "preferredDuringScheduling..." so that if there are less nodes than desired replicas you can still scale up.

charts/slim/values.yaml
```diff
@@ -2,22 +2,28 @@
...
-  daemonset: false
+  # Deploy one instance per node (uses pod anti-affinity + hostPort)
+  onePerNode: false
   # This will set the replicaset count more information can be found here: https://kubernetes.io/docs/concepts/workloads/controllers/replicaset/
   replicaCount: 1
 
   # This sets the container image more information can be found here: https://kubernetes.io/docs/concepts/containers/images/
   image:
-    repository: ghcr.io/agntcy/slim
+    repository: localhost:5001/slim
```
Tehsmash: need to revert

charts/slim/values.yaml
```diff
@@ -2,22 +2,28 @@
...
     # This sets the pull policy for images.
     pullPolicy: IfNotPresent
     # Overrides the image tag whose default is the chart appVersion.
-    tag: ""
+    tag: "test8"
```
Tehsmash: ditto

charts/slim/values.yaml
```diff
@@ -73,17 +79,16 @@ slim:
           #       key_file: "/svids/tls.key"
           #       ca_file: "/svids/svid_bundle.pem"
 
-  # StatefulSet-specific settings
-  # DaemonSet-specific settings
+  # onePerNode-specific settings (used when onePerNode: true)
   hostNetwork: false
```
Tehsmash: I don't think we need this for either Daemonset mode or deployment mode, unless we expect things running on the host/accessing the host to access this SLIM deployment.

data-plane/core/datapath/src/peer_discovery/kubernetes.rs
```diff
@@ -0,0 +1,658 @@
...
+                        warn!(error = %e, "k8s watcher error, will retry");
+                        // kube-runtime watcher handles retries internally,
+                        // but if the stream terminates we exit.
+                    }
+                }
+            }
+        });
+
+        Ok(())
+    }
```
Tehsmash, line 329: We actually might be able to take advantage of k8s a little bit more and reduce the logic we have to do in rust.

In K8s you can create a Headless service (https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) this will perform all the label selecting, readyness checks etc for us, and then on the rust side we can simply query the EndpointSlice associated with the headless service to see all the currently valid pods and IP addresses, essentially giving us the peer list in one shot. No need to handle deletions etc.

You can even avoid hitting the k8s API and dealing with permissions etc all together by using the K8s DNS to the Headless service's DNS name, as it will populate A records for each ready pod associated with the headless service.

Tehsmash: In fact thinking about it DNS based discovery using A records could even work outside of k8s, WDYT?

msardara: I thought about using the headless service, but in that case we lose the ability to immediately get an event when the deployment is scaled up or down. If we rely on the DNS we need to poll it and - in any case - wait for the records to be updated.

Also when a deployment is deleted / recreated we might get the IP address of pods belonging to the old deployment (which are maybe going to be terminated) and try to connect to them, while here we limit link creation to the current deployment, as we filter by deployment hash.

A thing we could do is having both discoveries: the current one, which is querying the API server and requiring rbac, and the one based on DNS, which could work also outside k8s.

msardara: But I'd leave it for a follow up PR

charts/slim/templates/deployment.yaml
msardara: I will restore the DaemonSet. For the anti-affinity, I was thinking to leave its configuration to the user. If they want a deployment and not a daemonset, they should also able to set the `affinity` as they desire.

data-plane/core/datapath/src/peer_discovery/kubernetes.rs
Tehsmash: I think if we watch the headless service we'll see update events as soon as the pods become ready / the endpoints get updated, and when pod endpoints are removed. We'll avoid having to handle events related to pod live-cycle / readyness and leave that to the service. I think we can also replace/sync the whole discovered_peers list on every update from the endpoint slice event instead of having to manage it per-peer etc.

We can also have the headless service filter by deployment hash (although this would mean that as soon as we deploy it kicks all the old endpoints off). I think headless service will remove a pod that is getting terminated from the endpoint list automatically. 

But this opens a question: during a rolling upgrade wouldn't we want both old and new to peer together? If we filter on deployment hash then we'll end up potentially creating a network partition while the rolling upgrade is happening. If a particular pod gets stuck k8s will freeze the rolling upgrade and this could cause more significant problems.

I like the idea of adding the DNS A names one as a second discovery path too.

msardara: > I think if we watch the headless service we'll see update events as soon as the pods become ready / the endpoints get updated, and when pod endpoints are removed. We'll avoid having to handle events related to pod live-cycle / readyness and leave that to the service. I think we can also replace/sync the whole discovered_peers list on every update from the endpoint slice event instead of having to manage it per-peer etc.

Aaah ok, I thought you wanted to make a dns query to the headless service, instead you want to watch the endpointslice: we can definitely do this, even if I I think we'd lose the pod name information, but it should be ok.

> But this opens a question: during a rolling upgrade wouldn't we want both old and new to peer together? If we filter on deployment hash then we'll end up potentially creating a network partition while the rolling upgrade is happening. If a particular pod gets stuck k8s will freeze the rolling upgrade and this could cause more significant problems.

Right, I was doing it mainly for wire protocol compatibility, but we'll be mostly making rolling upgrade with compatible versions, so it makes sense to let the 2 deployment create a unique group during the upgrade.

charts/slim/templates/deployment.yaml
```diff
@@ -68,16 +84,25 @@ spec:
...
               hostPort: {{ $data.port }}
+              {{- end }}
             {{- end }}
             {{- range $i, $ctrl := .Values.slim.service.control }}
             - name: controller-{{ $i }}
               containerPort: {{ $ctrl.port }}
               protocol: TCP
+              {{- if $isDaemonSet }}
               hostPort: {{ $ctrl.port }}
+              {{- end }}
```
Tehsmash: Do we need a hostPort for a daemonset deployment of SLIM? Do we expect this to be accessed outside of the k8s network?

Maybe we can make this a configuration value? ctrl.hostPort ?

charts/slim/templates/service.yaml
Tehsmash: Do we need two different services? or can we just set internalTrafficPolicy:Local on/off if deploymentMode == daemonset?

What I'm thinking is that it would be nice if you have an application in k8s using SLIM you can configure it to a single DNS name and not have to change it if you change the SLIM deployment mode?

Tehsmash (approved):

charts/slim/templates/service.yaml
msardara: Sounds good, I'll keep one single service with the Local traffic policy for daemonsets ✅

charts/slim/templates/deployment.yaml
msardara: I removed it for now, if we need it we can reintroduce it

micpapal (approved):

PR #1740, opened 2026-06-12
feat: add header integrity check and replay protection to control messages

### LINKED ISSUE #1721: Add header integrity validation to control messages
In the current solution e2e header integrity validation is implemented only for message and retransmission reply types. To prevent replay attacks with control messages, the integrity check should be extended to those messages.

markpmarton:
closes #1721 

# Description

In the current implementation only publish message headers are e2e validated because it is built around the MLS payload encryption. To prevent malicious control message replay attacks, we need to add manual integrity check for pre- and post-session messages.

To prevent replay attacks, simple replay check is implemented based on `message_id`.

## Changes
- AAD signature `e2e_header_sig` added for header fields (signed by MLS key)
- control message header fields signing and verification
- `message_id` is now also used to detect replayed messages
- introduced `max_seen_control_message_ids_size` to limit replay cache size

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


data-plane/core/session/src/session_controller.rs
```diff
@@ -182,6 +240,56 @@ impl SessionController {
...
+                | ProtoSessionMessageType::GroupRemove
+                | ProtoSessionMessageType::GroupClose
+                | ProtoSessionMessageType::GroupProposal
+                | ProtoSessionMessageType::GroupAck
+                | ProtoSessionMessageType::GroupNack
+                | ProtoSessionMessageType::Ping
+        );
+        // Require E2E verification only when the sender included a signature (matches pre-session path).
+        let e2e_required =
+            is_post_session_control && msg.get_slim_header().e2e_header_sig.is_some();
```
micpapal: this should be always mandatory otherwise we can always send fake packets with e2e_header_sig set to None

proto/data-plane/v1/data_plane.proto
```diff
@@ -66,6 +66,8 @@ message SLIMHeader {
   optional uint64 incoming_conn = 9;
   optional bool error = 10;
   optional bytes header_mac = 11;
+  optional uint64 sequence_number = 12;
```
micpapal: why do we need a new sequence_number here? isn't enough to use the message_id in session header?

data-plane/core/session/src/session_controller.rs
```diff
@@ -182,6 +240,56 @@ impl SessionController {
...
+        // 2. Perform sequence number check for command messages
+        if e2e_required && msg.get_session_message_type().is_command_message() {
+            let sender = msg.get_source();
+            let seq = msg.get_slim_header().sequence_number;
+            if let Some(s) = seq {
+                let mut seen_control_seqs = settings.seen_control_seqs.lock();
+                let seen = seen_control_seqs.entry(sender.clone()).or_default();
+                if !seen.insert(s) {
+                    // Duplicate delivery (network retransmission or fanout). Handlers are idempotent.
+                    return Ok(());
```
micpapal: I am not sure here, but should we drop the message here instead of return ok? If it is in seen it means that the message was already processed.  Notice that if we use the message_id in the session header RTX message will have a different id with respect to the original message so the two messages will be handled as two different messages.

data-plane/core/session/src/session_controller.rs
```diff
@@ -286,7 +393,7 @@ impl SessionController {
                                     }
                                 }
                                 Err(e) => {
-                                    debug!(
+                                    tracing::error!(
```
micpapal: this should do back to debug

data-plane/core/session/src/session_controller.rs
```diff
@@ -32,14 +32,72 @@ use crate::{
...
+            pubkey: Option<String>,
+        }
+
+        #[derive(serde::Deserialize)]
+        struct IdentityClaims {
+            pubkey: Option<String>,
+            custom_claims: Option<CustomClaims>,
+        }
+
+        let claims_res = verifier.get_claims(&identity).await;
```
micpapal: here I think we can reuse the IdentityClaims in Auth. This part can be replaced with something like

```
 let claims_json = verifier.get_claims(&identity).await?;
 let identity_claims = slim_auth::identity_claims::IdentityClaims::from_json(&claims_json)?;
 let pubkey = identity_claims.public_key;
```

data-plane/core/session/src/session_controller.rs
```diff
@@ -659,7 +769,30 @@ where
...
+        for msg in &mut output.messages {
+            if let OutboundMessage::ToSlim(m) = msg
+                && m.get_session_message_type().is_command_message()
+            {
+                m.get_slim_header_mut().sequence_number = Some(self.next_control_seq);
+                self.next_control_seq += 1;
+
+                let aad = crate::mls_state::build_aad(m);
+                let private_key = identity_provider.get_signature_secret_key()?;
+                let public_key = identity_provider.get_signature_public_key()?;
```
micpapal: This can be moved out of the loop

let private_key = identity_provider.get_signature_secret_key()?;
let public_key = identity_provider.get_signature_public_key()?;

data-plane/core/session/src/session_layer.rs
```diff
@@ -547,11 +547,17 @@ where
                 Err(_) => return,
             };
 
-            if let Err(e) =
-                crate::session_controller::verify_identity(&message, &layer.identity_verifier).await
+            let e2e_required = message.get_slim_header().e2e_header_sig.is_some();
```
micpapal: here this also should always be mandatory

micpapal (changes requested):

proto/data-plane/v1/data_plane.proto
markpmarton: AFAIK message_id is getting a random u32 for control messages, that is not viable for replay protection. Maybe it can be changed to a monotonic counter, I'm going to check it.

data-plane/core/session/src/session_layer.rs
markpmarton: This block only runs on DiscoveryRequest messages, where no Session is available yet. The future SessionConfig cannot be predicted, so validation is set to a constant true.

data-plane/core/session/src/session_controller.rs
markpmarton: Now the duplicated messages are dropped. If MLS is on an error is raised, otherwise the package is dropped silently.

proto/data-plane/v1/data_plane.proto
markpmarton: Sequence_number is now removed, message_id is used for identifying already handled messages.

data-plane/core/session/src/session_controller.rs
markpmarton: It now depends on the MlsSettings value set in the SessionConfig.

crates/session/src/session_controller.rs
```diff
@@ -32,14 +33,75 @@ use crate::{
...
+) -> Result<(), SessionError>
+where
+    P: TokenProvider,
+{
+    if !msg.get_session_message_type().is_command_message() {
+        return Ok(());
+    }
+    let private_key = match identity_provider.get_signature_secret_key() {
+        Ok(k) if !k.is_empty() => k,
+        _ => return Ok(()),
```
micpapal: here if we are not able to get the private key we cannot sign the packet and I assume it will be dropped at the receiver side. Isn't it better to return an error in this case so that the sender see why the transmission failed?

crates/session/src/session_controller.rs
```diff
@@ -32,14 +33,75 @@ use crate::{
...
+    };
+    let public_key = match identity_provider.get_signature_public_key() {
+        Ok(k) if !k.is_empty() => k,
+        _ => return Ok(()),
```
micpapal: same here

crates/session/src/session_settings.rs
```diff
@@ -75,4 +81,26 @@ where
 
     /// Service ID for tracing — identifies which service instance owns this session
     pub(crate) service_id: String,
+
+    /// Seen control-message ids per remote sender (E2E replay protection).
+    pub(crate) seen_control_message_ids: Arc<Mutex<HashMap<ProtoName, HashSet<u32>>>>,
```
micpapal: This struct is populated at config time while seen_control_message_ids changes over time. It would be better to move it in session_controller so we can also avoid the lock. The insertion work as it is now. For the removal the controller sees all the messages so we can check if it is as leaveRequest or a GroupRemove and we can remove the name from the map, avoiding the need to do it on the moderator side. 

In addition this map can grow with no limit. we can check something like https://crates.io/crates/lru that has a fix length. Maybe there are better options.

crates/session/src/session_controller.rs
```diff
@@ -662,11 +790,46 @@ where
...
+        }
+        Ok(())
+    }
+
     /// Send control message through ControllerSender, returning the output.
     pub(crate) fn send_with_timer(
         &mut self,
         message: Message,
     ) -> Result<SessionOutput, SessionError> {
+        let mut output = SessionOutput::to_slim(message);
```
micpapal: isn't this enough?

```
 pub(crate) fn send_with_timer(
     &mut self,
     mut message: Message,
 ) -> Result<SessionOutput, SessionError> {
     sign_outbound_control_message(&mut message, &self.settings.identity_provider)?;
     self.sender.on_message(&message)
 }

File: crates/session/src/session_controller.rs [hunk 12]
```diff
@@ -662,11 +790,46 @@ where
...
+        let public_key = match identity_provider.get_signature_public_key() {
+            Ok(k) if !k.is_empty() => k,
+            _ => return Ok(()),
+        };
+        for msg in &mut output.messages {
+            if let OutboundMessage::ToSlim(m) = msg
+                && m.get_session_message_type().is_command_message()
+            {
+                let aad = crate::mls_state::build_aad(m);
+                let signature = slim_auth::utils::sign_header_aad(&aad, &private_key, &public_key)?;
```
micpapal: sign_outbound_control_message can be used also here to simply this function

crates/session/src/session_controller.rs
```diff
@@ -182,6 +244,57 @@ impl SessionController {
...
+                | ProtoSessionMessageType::LeaveReply
+                | ProtoSessionMessageType::GroupAdd
+                | ProtoSessionMessageType::GroupRemove
+                | ProtoSessionMessageType::GroupClose
+                | ProtoSessionMessageType::GroupProposal
+                | ProtoSessionMessageType::GroupAck
+                | ProtoSessionMessageType::GroupNack
+                | ProtoSessionMessageType::Ping
+        );
+        // Require E2E verification only when the sender included a signature (matches pre-session path).
```
micpapal: this is an old comment

crates/session/src/session_controller.rs
```diff
@@ -182,6 +244,57 @@ impl SessionController {
...
+                    slim_auth::errors::AuthError::TokenInvalid,
+                ));
+            }
+            let mut seen = settings.seen_control_message_ids.lock();
+            let seen = seen.entry(sender.clone()).or_default();
+            if !seen.insert(message_id) {
+                if e2e_required {
+                    return Err(SessionError::ControlMessageReplay { message_id });
+                } else {
+                    return Ok(false);
```
micpapal: does this drop the packet? if e2e_required is false we should just return at line 271 I think and we should not drop

micpapal (changes requested):

crates/session/src/session_controller.rs
markpmarton: for empty private and public keys or failing collection now error is raised.

crates/session/src/session_settings.rs
markpmarton: The HashMap is replaced with LruCache and moved to processing_loop.

crates/session/src/session_controller.rs
markpmarton: the function is removed now

crates/session/src/session_controller.rs
markpmarton: Updated.

crates/session/src/session_controller.rs
markpmarton: Early return for non-MLS path, otherwise packet drop

crates/session/src/session_controller.rs
```diff
@@ -182,6 +279,56 @@ impl SessionController {
...
+                return Err(SessionError::Auth(
+                    slim_auth::errors::AuthError::TokenInvalid,
+                ));
+            }
+
+            if !seen_control_message_ids.contains(&sender) {
+                seen_control_message_ids.put(
+                    sender.clone(),
+                    LruCache::new(
+                        NonZeroUsize::new(settings.max_seen_control_message_ids_size).unwrap(),
```
micpapal: this panic if settings.max_seen_control_message_ids_size is 0. move to default in this case

micpapal (approved):
lgtm, just one minor fix before marge

crates/session/src/session_controller.rs
markpmarton: the variable type is changed to NonZeroUsize. SessionBuilder falls back to default value if it is 0 in the settings.

PR #1748, opened 2026-06-17
feat(slimctl): improve bench pub/sub commands and reporting

Tehsmash:
## Description

Improves the `slimctl bench` pub/sub commands with better reporting and minor correctness fixes.

- **Richer per-worker and aggregate output**: the `SampleGroup` report now includes actual sent/received message counts alongside the throughput figures. When messages are dropped the output shows `<actual>/<expected> msgs, <N> dropped`; when there are no drops it stays compact: `(50,000 msgs)`. This makes message loss immediately visible without post-processing logs.
- **Propagate `controller.close()` await error**: the awaited result of `controller.close()` in `run_pub_worker` was silently discarded; added a trailing `?` to propagate it.
- **Remove unnecessary 5s sleep**: `run_pub` was sleeping 5 s unconditionally after all pub workers completed. Message processing tasks complete when the session closes, so the delay served no purpose.

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -569,9 +569,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.102"
+version = "1.0.103"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
+checksum = "2a4385e2e34eb35d6b3efe798b9eb88096925d87726c0798709bf56d9ed84af3"
 
 [[package]]
 name = "arc-swap"
--- a/crates/slimctl/src/commands/bench.rs
+++ b/crates/slimctl/src/commands/bench.rs
@@ -982,8 +982,6 @@ async fn run_pub(args: &BenchPubArgs, service: &Arc<Service>) -> Result<()> {
         }
     }
 
-    tokio::time::sleep(Duration::from_secs(5)).await; // give message processing tasks time to complete
-
     if group.has_samples() {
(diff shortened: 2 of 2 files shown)


data-plane/slimctl/src/commands/bench.rs
```diff
@@ -121,6 +121,19 @@ pub struct BenchSubArgs {
...
+    /// Example: process A: --count 2 --start-index 0  (registers sub-0, sub-1)
+    ///          process B: --count 2 --start-index 2  (registers sub-2, sub-3)
+    #[arg(long, default_value_t = 0)]
+    pub start_index: usize,
+
+    /// Run in reliable mode.
+    /// By default the benchmark runs in unreliable (fire-and-forget) mode. With
+    /// --reliable, the session layer retransmits unacknowledged messages.
+    #[arg(long)]
+    pub reliable: bool,
```
micpapal: This is used only as a print because it is the pub that decides if the session is reliable or not. I think we should remove this flag.

data-plane/slimctl/src/commands/bench.rs
```diff
@@ -1246,6 +1334,7 @@ async fn run_channel_pub_worker(
         channel_name.clone(),
         Duration::from_secs(5),
         "group session creation timed out unexpectedly",
+        true,
```
micpapal: Here we are ignoring the reliable flag and set reliable always to true. is this wanted?

micpapal (approved):

PR #1760, opened 2026-06-19
ci: pin image versions in trivy scan matrix

muscariello:
## Summary

- Pin `slim-node` to `2.0.0-alpha.1` (from `slim-v2.0.0-alpha.1` tag)
- Pin `slim-controller` to `1.4.0` (from `control-plane-v1.4.0` tag)


muscariello:
Closing — confirmed that `latest` exists on GHCR and correctly points to `1.4.0`. The scan was already hitting the right image. No change needed.

Issue #1729, opened 2026-06-09
Add network segmentation


micpapal:
This is step 2 in #1547

micpapal:
# Topology Configuration Examples

## 1. Full Mesh (default)

No topology config needed — all groups link to each other, all routes visible everywhere. This is the current default behavior.

```yaml
# Option A: no topology key at all (implicit full mesh)

# Option B: explicit full mesh
topology:
  links:
    - name: "*"
      peers: ["*"]
```

All registered groups connect to each other. Every agent sees every other agent.

---

## 2. Star — No Segmentation

```yaml
topology:
  links:
    - name: platform
      peers: ["*"]
```

`platform` connects to every group. Groups only connect to `platform` (not to each other directly). However, **all routes are visible everywhere** — an agent on `customer-a` can discover agents on `customer-b` (routed through `platform`). No isolation.

---

## 3. Star — With Segmentation

```yaml
topology:
  segments:
    - name: segment-$group
      links:
        - name: platform
          peers: [$group]
```

When `customer-a`, `customer-b`, `customer-c` register, this expands to:

- `segment-customer-a`: platform ↔ customer-a
- `segment-customer-b`: platform ↔ customer-b
- `segment-customer-c`: platform ↔ customer-c

### Visibility

| Group | Sees | Does NOT see |
|-------|------|-------------|
| customer-a | platform | customer-b, customer-c |
| customer-b | platform | customer-a, customer-c |
| customer-c | platform | customer-a, customer-b |
| platform | customer-a, customer-b, customer-c | — |

New groups are automatically isolated — no config change needed.

---

## 4. Arbitrary Topology — 2 Explicit Segments

Five groups: `hub`, `shared-db`, `frontend-a`, `frontend-b`, `backend`.

```yaml
topology:
  segments:
    - name: production
      links:
        - name: hub
          peers: [frontend-a, frontend-b, backend]
        - name: backend
          peers: [hub, shared-db]
        - name: shared-db
          peers: [backend]
    - name: staging
      links:
        - name: hub
          peers: [frontend-a, shared-db]
        - name: shared-db
          peers: [hub]
```

### Segment `production` — link graph

```
frontend-a ↔ hub ↔ frontend-b
                 ↔ backend ↔ shared-db
```

### Segment `staging` — link graph

```
frontend-a ↔ hub ↔ shared-db
```

### Visibility

| Group | Sees (via production) | Sees (via staging) | Does NOT see |
[...]

PR #1761, opened 2026-06-19
feat: network segmentation

micpapal:
# Description

 Implements network segmentation for the SLIM control plane, allowing partitioning of the network into independent routing domains (segments). This enables topologies like hub-and-spoke where customer groups are isolated from each other while all sharing access to a common platform group.
 
 ref: https://github.com/agntcy/slim/issues/1729
 
 ## What's New
 
 ### Segment-based topology configuration
 
 The topology config now supports three modes:
 
 1. **No topology** → full mesh
 2. **`topology.links`** → single segment with a custom link graph 
 3. **`topology.segments`** → multiple independent routing domains (new)
 
 ```yaml
 topology: segments:
     - name: segment-$group
       links:
         - name: hub
           neighbors: [$group]
```

### Dynamic $group expansion

Segment templates using $group are automatically expanded at runtime when new groups register. In the star example above, each customer group gets its own isolated segment with the hub, without requiring explicit per-customer configuration.

### Per-segment SPT routing

Each segment maintains its own link graph. When a name is announced:

 - The SPT is computed within each segment containing the announcer
 - Routes are only installed to nodes reachable within the same segment
 - A group present in multiple segments (e.g., hub) computes routes for all of them

### Bug fixs

1. Route cleanup race :

Fixed a race condition between the reconciler and the delete handler where the reconciler could overwrite a Deleted route status back to Applied after receiving a stale ack from a node.

2. NameId display fix:

Fixed NULL_COMPONENT appearing in route outlines for well-known channel IDs (DATA_CHANNEL_ID, CONTROL_CHANNEL_ID) by using NameId::try_from instead of UUID-only parsing in northbound/southbound services.

## Type of Change

- [x] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### CODE REVIEW

File: crates/session/src/controller_sender.rs [hunk 1]
```diff
@@ -196,34 +196,34 @@ impl ControllerSender {
                     ps.missing_pings.remove(&message.get_dst());
                 }
             }
-            slim_datapath::api::ProtoSessionMessageType::JoinRequest => {
+            slim_datapath::api::ProtoSessionMessageType::JoinRequest
+                if self.group_name.is_none() =>
+            {
```
Tehsmash, line 201: Is this related to this change or just a code style clean up?

crates/controller/src/service.rs
```diff
@@ -392,7 +392,7 @@ impl ControlPlane {
                             Some(res) => {
                                 match res {
                                     Ok(msg) => {
-                                        debug!("Send sub/unsub to control plane for message: {:?}", msg);
+                                        debug!("Receive message from data plane: {:?}", msg);
```
Tehsmash:
```suggestion
                                        debug!("Received message from data plane: {:?}", msg);
```

Tehsmash: Is this change correct, it seems like the original one was right? Maybe we need a mix of both:

"Received message from data plane, forwarding to control plane: {:?}"

crates/control-plane/README.md
```diff
@@ -105,19 +105,63 @@ topology:
...
+topology:
+  segments:
+    - name: tenant-1
+      links:
+        - name: cloud
+          neighbors: [customer-a]
+    - name: tenant-2
+      links:
+        - name: cloud
+          neighbors: [customer-b, customer-c]
```
Tehsmash:
```suggestion
topology:
  segments:
    - name: customer-1
      links:
        - name: cloud
          neighbors: [cluster-a]
    - name: customer-2
      links:
        - name: cloud
          neighbors: [cluster-b, cluster-c]
```

crates/control-plane/README.md
```diff
@@ -105,19 +105,63 @@ topology:
...
 
+#### Segments (route isolation)
+
+Segments partition the network into independent routing domains.
+Agents registered in one segment are **invisible** to agents in other
+segments — routes are only expanded within the segment's topology graph.
+
+Each segment has its own `links` adjacency list. The special token
+`$group` expands to every registered group, enabling template-style
+configs that create a per-group segment dynamically.
```
Tehsmash: The way I read this it sounds like $group expands to multiple groups, so `neighbours: [$group]` would include all the groups for example. I assume that isn't the case after reading further but we might need to revisit the wording here.

crates/control-plane/README.md
```diff
@@ -105,19 +105,63 @@ topology:
...
+**Per-group isolation** (each customer reaches cloud but not other customers):
+
+```yaml
+topology:
+  segments:
+    - name: segment-$group
+      links:
+        - name: cloud
+          neighbors: [$group]
+```
```
Tehsmash, line 142: I wonder if it would be better to separate the template so that we leave room for a bit more flexibility, i'm envisioning something similar to what we see in ArgoCD / External secrets operator e.g.

```
topology:
	segments-template:
		{{- range $group := .groups }}
		- name: segment-$group
		   neighbors: [$group]
		{{- end }}
```

this could allow for things like:

```
topology:
	segments-template:	
		- name: segment-customer-a
		   neighbors: [{{- range $group := .groups -}}{{if startsWith $group "customer-a-"}}{{ $group }}{{- end }}]
		- name: segment-customer-b
		   neighbors: [{{- range $group := .groups -}}{{if startsWith $group "customer-b-"}}{{ $group }}{{- end }}]	   
```

What do you think?

crates/session/src/controller_sender.rs
micpapal: It is just a style clean up due to the new version of the linter in the CI

crates/control-plane/README.md
micpapal: I agree we need to do something better here to make the segments easier to configure. However I think we can merge this one as it is and create a new issue. In this way we can start to test the current code.

micpapal: Improve segment configuration with template engine support: https://github.com/agntcy/slim/issues/1767

Tehsmash (approved):

PR #1762, opened 2026-06-19
perf(datapath): zero-copy proto layout and forwarding path improvements

Tehsmash:
# Description

Performance improvements to the data-plane datapath focused on reducing allocation and decoding overhead on the hot forwarding path. Builds on top of the `feat/slimctl-bench-improvements` branch.

### Data-plane throughput (profiling-driven)
- General throughput improvements based on profiling results
- Reverted server-side `OpenChannel` buffer from 1024 back to 128 (regression fix)

### Zero-copy proto field layout
- **`Name` proto**: replaced nested `EncodedName`/`NameId`/`StringName` message fields with flat `bytes encoded_name` (40 bytes: 3×u64 LE + u128 LE id) and `bytes str_name` (packed string components). Eliminates nested proto struct allocation on every name access.
- **`SLIMHeader` source/destination**: replaced `Name source = 1` / `Name destination = 2` message fields with flat `bytes source = 1` / `bytes destination = 2` (40-byte encoded name directly). The `str_name` parts move to new fields `source_str = 12` / `destination_str = 13`. Eliminates `Option<Name>` unwrap and nested struct indirection on every routing decision.
- **`Publish.msg`**: changed from `Content msg = 3` (message) to `bytes msg = 3`. Data-plane-only forwarding nodes now pass the application payload through without decoding it. Session-layer nodes decode on demand via `get_payload() -> Content`.

### Dead code removal
- Removed `ProtoPublish::is_command`, `get_application_payload`, `get_command_payload` (zero callers post-refactor)
- Removed duplicate `set_error_flag` (identical to `set_error`)
- Removed `ProtoName::name_id` (no callers)

> **Note:** The proto layout changes are wire-breaking. All SLIM nodes must be updated together.

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [x] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/crates/control-plane/src/route_service/commands.rs
+++ b/crates/control-plane/src/route_service/commands.rs
@@ -4,7 +4,7 @@
 //! Public API commands called by the northbound gRPC handlers.
 //! These are the entry points that `slimctl` and other clients invoke.
 
-use slim_datapath::api::NameId;
+use slim_datapath::api::NULL_COMPONENT;
 use uuid::Uuid;
 
 use crate::api::proto::controller::proto::v1::{
@@ -37,7 +37,7 @@ impl super::RouteService {
         // downward path from root to new announcer).
         let existing_root = if source_node_id == ALL_NODES_ID {
             let (c0, c1, c2) = route_name.str_components();
-            let comp_id = if route_name.id() == NameId::NULL_COMPONENT {
+            let comp_id = if route_name.id() == NULL_COMPONENT {
                 None
             } else {
                 Some(route_name.string_id())
@@ -81,7 +81,7 @@ impl super::RouteService {
             let all_nodes = self.0.db.list_nodes().await?;
             let all_links = self.0.db.list_all_links().await?;
             let (c0, c1, c2) = route_name.str_components();
-            let comp_id = if route_name.id() == NameId::NULL_COMPONENT {
+            let comp_id = if route_name.id() == NULL_COMPONENT {
                 None
             } else {
                 Some(route_name.string_id())
@@ -137,7 +137,7 @@ impl super::RouteService {
         };
 
         let (c0, c1, c2) = route_name.str_components();
-        let comp_id = if route_name.id() == NameId::NULL_COMPONENT {
+        let comp_id = if route_name.id() == NULL_COMPONENT {
             None
         } else {
             Some(route_name.string_id())
--- a/crates/control-plane/src/route_service/reconciler.rs
+++ b/crates/control-plane/src/route_service/reconciler.rs
@@ -18,7 +18,7 @@ use crate::api::proto::controller::proto::v1::{
 use crate::db::{LinkStatus, RouteStatus, SharedDb};
 use crate::node_transport::{DefaultNodeCommandHandler, NodeStatus, ResponseKind};
 use crate::workqueue::WorkQueue;
-use slim_datapath::api::{NULL_COMPONENT, ProtoName};
+use slim_datapath::api::{NULL_COMPONENT, ProtoName, id_from_str};
 
 use super::is_connection_not_found;
 
@@ -334,8 +334,7 @@ async fn build_desired_routes<'a>(
                         route
                             .component_id
                             .as_deref()
-                            .and_then(|s| uuid::Uuid::parse_str(s).ok())
-                            .map(|u| u.as_u128())
+                            .and_then(|s| id_from_str(s).ok())
                             .unwrap_or(NULL_COMPONENT),
                     ),
             ),
--- a/crates/control-plane/src/route_service/reconciler.rs
+++ b/crates/control-plane/src/route_service/reconciler.rs
@@ -18,7 +18,7 @@ use crate::api::proto::controller::proto::v1::{
 use crate::db::{LinkStatus, RouteStatus, SharedDb};
 use crate::node_transport::{DefaultNodeCommandHandler, NodeStatus, ResponseKind};
 use crate::workqueue::WorkQueue;
-use slim_datapath::api::{NameId, ProtoName};
+use slim_datapath::api::{NULL_COMPONENT, ProtoName};
 
 use super::is_connection_not_found;
 
@@ -334,9 +334,9 @@ async fn build_desired_routes<'a>(
                         route
                             .component_id
                             .as_deref()
-                            .and_then(|s| NameId::try_from(s.to_string()).ok())
-                            .map(|nid| -> u128 { nid.into() })
-                            .unwrap_or(NameId::NULL_COMPONENT),
+                            .and_then(|s| uuid::Uuid::parse_str(s).ok())
+                            .map(|u| u.as_u128())
+                            .unwrap_or(NULL_COMPONENT),
                     ),
             ),
             link_id: Some(link_id.to_string()),
@@ -440,7 +440,7 @@ async fn process_route_acks(
         let link_id = sub.link_id.clone().unwrap_or_default();
         let proto_name = sub.name.as_ref().unwrap();
         let (c0, c1, c2) = proto_name.str_components();
-        let comp_id = if proto_name.id() == NameId::NULL_COMPONENT {
+        let comp_id = if proto_name.id() == NULL_COMPONENT {
             None
         } else {
             Some(proto_name.string_id())
--- a/crates/control-plane/src/services/northbound.rs
+++ b/crates/control-plane/src/services/northbound.rs
@@ -1,7 +1,7 @@
 // Copyright AGNTCY Contributors (https://github.com/agntcy)
 // SPDX-License-Identifier: Apache-2.0
 
-use slim_datapath::api::{NameId, ProtoName};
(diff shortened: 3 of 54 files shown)


crates/datapath/src/message_processing.rs
```diff
@@ -394,7 +395,7 @@ impl MessageProcessor {
         match channel {
             TransportChannel::Grpc(grpc_channel) => {
                 let mut client = DataPlaneServiceClient::new(grpc_channel);
-                let (tx, rx) = mpsc::channel(128);
+                let (tx, rx) = mpsc::channel(1024);
```
micpapal, line 398: Is this the right value or a left over from the tests?

crates/control-plane/src/route_service/reconciler.rs
```diff
@@ -334,9 +334,9 @@ async fn build_desired_routes<'a>(
                         route
                             .component_id
                             .as_deref()
-                            .and_then(|s| NameId::try_from(s.to_string()).ok())
-                            .map(|nid| -> u128 { nid.into() })
-                            .unwrap_or(NameId::NULL_COMPONENT),
+                            .and_then(|s| uuid::Uuid::parse_str(s).ok())
+                            .map(|u| u.as_u128())
+                            .unwrap_or(NULL_COMPONENT),
```
micpapal, line 338: Here if the id is set to "DATA_CHANNEL_ID" I think uuid::Uuid::parse_str(s)  silently falling back to NULL_COMPONENT and causing incorrect routing. Use id_from_str() should solve the problem

crates/control-plane/src/services/northbound.rs
```diff
@@ -286,12 +286,9 @@ impl ControlPlaneService for NorthboundApiService {
...
                             .as_deref()
-                            .map(|s| {
-                                NameId::try_from(s.to_string())
-                                    .map(u128::from)
-                                    .unwrap_or(NameId::NULL_COMPONENT)
-                            })
-                            .unwrap_or(NameId::NULL_COMPONENT),
+                            .and_then(|s| uuid::Uuid::parse_str(s).ok())
+                            .map(|u| u.as_u128())
+                            .unwrap_or(NULL_COMPONENT),
```
micpapal, line 293: Here there is the same problem as in reconciler.rs

crates/control-plane/src/services/southbound.rs
```diff
@@ -266,12 +266,9 @@ async fn build_node_routes(db: &SharedDb, node_id: &str) -> Vec<ProtoRoute> {
...
                         .as_deref()
-                        .map(|s| {
-                            NameId::try_from(s.to_string())
-                                .map(u128::from)
-                                .unwrap_or(NameId::NULL_COMPONENT)
-                        })
-                        .unwrap_or(NameId::NULL_COMPONENT),
+                        .and_then(|s| uuid::Uuid::parse_str(s).ok())
+                        .map(|u| u.as_u128())
+                        .unwrap_or(NULL_COMPONENT),
```
micpapal, line 270: And also here

crates/proto/src/impls.rs
```diff
@@ -211,145 +207,204 @@ impl SlimHeaderFlags {
...
-    }
+/// Decode packed str_name bytes into three byte slices.
+/// Returns `None` if the slice is empty (field absent) or malformed.
+#[inline]
+pub fn decode_str_bytes(b: &bytes::Bytes) -> Option<(&[u8], &[u8], &[u8])> {
+    if b.is_empty() {
+        return None;
+    }
+    let data = b.as_ref();
+    let len0 = u32::from_le_bytes(data.get(0..4)?.try_into().unwrap()) as usize;
```
micpapal: here if len < 4 this will panic. same for all the other steps.

micpapal (changes requested):

micpapal (approved):

crates/proto/proto/data-plane/v1/data_plane.proto
```diff
@@ -26,7 +26,10 @@ message Unsubscribe {
 message Publish {
   SLIMHeader header = 1;
   SessionHeader session = 2;
-  Content msg = 3;
+  // Raw-bytes encoding of a Content message. Stored as bytes so that
+  // data-plane-only forwarding nodes can pass it through without decoding.
+  // Wire-compatible with the original `Content msg = 3` field (same wire type 2).
```
msardara, line 31: I'd remove this line, as the comment refer to the previous Publish message definition.

crates/proto/src/impls.rs
```diff
@@ -834,17 +841,11 @@ impl ProtoMessage {
...
         }
-        match &slim_header.destination {
-            None => return Err(MessageError::DestinationNotFound),
-            Some(dst) if dst.name.is_none() => {
-                return Err(MessageError::DestinationEncodedNameNotFound);
-            }
-            _ => {}
+        if slim_header.destination.is_empty() {
+            return Err(MessageError::DestinationEncodedNameNotFound);
         }
```
msardara, line 849: I think here we should also check the length of source/destination and we should also make sure that destination_str can be decoded as utf8.

If we don't do this, a peer sending a packet with a non-empty but invalid source or destination might trigger a panic in decode_name_bytes(&self.destination).expect("destination not set"), where we call expect() thinking the message was validated.

Otherwise we can remove the expect() and handle the error.

crates/proto/src/impls.rs
```diff
@@ -754,48 +786,19 @@ impl ProtoPublish {
...
-        match self
-            .get_payload()
-            .content_type
-            .as_ref()
-            .expect("content missing")
-        {
-            ContentType::AppPayload(application_payload) => application_payload,
-            ContentType::CommandPayload(_) => panic!("the payload is not an application payload"),
+    /// Decodes the raw `msg` bytes into a [`Content`], returning `None` if absent.
+    pub fn get_payload(&self) -> Option<Content> {
```
msardara, line 790: It looks like this performs a new allocation on each call. It seems `encrypt_message`/`decrypt_message` decode the payload, then build_aad(msg) decodes it again just to read payload_type.

We could pass the already decoded `payload_type` into `build_aad`, as the caller already knows it.

crates/session/src/session_receiver.rs
msardara, line 1: Here we do `get_source()/get_dst()`, and then we immediately take the components().

Do you think we can directly use `get_encoded_source()/get_encoded_dst()`?

crates/session/src/session_sender.rs
msardara, line 1: Here we do `get_source()/get_dst()`, and then we immediately take the components().

Do you think we can directly use `get_encoded_source()/get_encoded_dst()`?

crates/control-plane/src/route_service/commands.rs
```diff
@@ -37,7 +37,7 @@ impl super::RouteService {
         // downward path from root to new announcer).
         let existing_root = if source_node_id == ALL_NODES_ID {
             let (c0, c1, c2) = route_name.str_components();
-            let comp_id = if route_name.id() == NameId::NULL_COMPONENT {
+            let comp_id = if route_name.id() == NULL_COMPONENT {
                 None
             } else {
```
msardara, line 42: It looks like this pattern is repeated several times. Should we create a helper in ProtoName which does the same? Something like

```rust
impl ProtoName {
  fn opt_string_id() {
      if name.id() == NULL_COMPONENT { None } else { Some(name.string_id()) }
  }
}

File: crates/datapath/src/header_mac.rs [hunk 10]
```diff
@@ -226,36 +208,21 @@ fn push_bool_opt(buf: &mut Vec<u8>, v: Option<bool>) {
...
+        buf.push(1);
+        buf.extend_from_slice(encoded);
+    }
+    // str_name: presence byte + u32 LE total length + raw packed bytes
+    if str_name.is_empty() {
+        buf.push(0);
+    } else {
+        buf.push(1);
+        buf.extend_from_slice(&(str_name.len() as u32).to_le_bytes());
+        buf.extend_from_slice(str_name);
```
msardara, line 225: Shouldn't we reuse encode_str_bytes/decode_str_bytes here?

crates/proto/src/impls.rs
```diff
@@ -211,145 +207,204 @@ impl SlimHeaderFlags {
...
-            }
-        }
-    }
+/// Decode a 40-byte flat `Bytes` value into `(c0, c1, c2, id)`.
+/// Returns `None` if the slice is empty (field absent).
+#[inline]
+pub(crate) fn decode_name_bytes(b: &bytes::Bytes) -> Option<(u64, u64, u64, u128)> {
+    if b.is_empty() {
+        return None;
+    }
```
msardara, line 276: The != 40 check already covers the is_ampty check, so I think we can remove `if b.is_empty()...`

msardara (changes requested):

muscariello:
@Tehsmash This PR is stale since more than 30 days.
I am closing this. Consider resurrecting this work if ready to complete.

PR #1772, opened 2026-06-26
feat: config mode vs api mode for topology management

### LINKED ISSUE #1768: Config mode vs API mode for topology management
## Context

Currently the SLIM control plane topology (segments, groups, links) is entirely driven by the config file. There is no API to mutate topology at runtime. This issue proposes a clear separation between config-driven and API-driven topology management, along with the DB schema and CLI commands needed to support it.

## Design

The control plane operates in one of two mutually exclusive modes based on whether the config file contains a `topology` section:

### Config mode (topology section present in config)

- Config file is the **single source of truth**
- On restart: DB is wiped and config is re-applied
- Dynamic `$group` expansion still happens at runtime (re-evaluated when new nodes register)
- Mutation APIs (`create`, `delete`, `add`, `remove`) return an error: _"topology is config-managed, modify the config file and restart"_
- List/read APIs work normally

### API mode (no topology section in config)

- DB is the **single source of truth**
- On restart: state is loaded from DB
- Full CRUD operations available via `slimctl` and gRPC APIs
- Config file only contains non-topology settings (reconciler params, TLS, etc.)

### Why this approach

- Zero ambiguity about who owns topology state
- No reconciliation logic between config and DB
- Explicit — the operator chooses their management model upfront

### `segments` table

| Column | Type | Description |
|--------|------|-------------|
| id | TEXT PK | Unique segment ID |
| name | TEXT | Segment name |
| created_at | TIMESTAMP | Creation time |

### `segment_groups` table (join table)

| Column | Type | Description |
|--------|------|-------------|
| segment_id | TEXT FK | References segments.id |
| group_name | TEXT | Group belonging to this segment |

### `groups` table (optional, for explicit group lifecycle)

| Column | Type | Description |
|--------|------|-------------|
| name | TEXT PK | Group name |
| created_at | TIMESTAMP | First seen / created time |

> Groups could remain implicit (derived from registered nodes) in config mode, but in API mode having an explicit table allows group lifecycle management independent of node registration.

### Segment management

```
slimctl controller segment create <NAME>
slimctl controller segment delete <NAME>
slimctl controller segment add-group <SEGMENT> <GROUP>
slimctl controller segment remove-group <SEGMENT> <GROUP>
```

### Link management

```
slimctl controller link add --source-group <GROUP> --dest-group <GROUP> [--config <FILE>]
slimctl controller link delete --source-group <GROUP> --dest-group <GROUP>
```

### Corresponding gRPC RPCs

- `CreateSegment(CreateSegmentRequest) returns (CreateSegmentResponse)`
- `DeleteSegment(DeleteSegmentRequest) returns (DeleteSegmentResponse)`
- `AddGroupToSegment(AddGroupToSegmentRequest) returns (AddGroupToSegmentResponse)`
- `RemoveGroupFromSegment(RemoveGroupFromSegmentRequest) returns (RemoveGroupFromSegmentResponse)`
- `AddLink(AddLinkRequest) returns (AddLinkResponse)`
- `DeleteLink(DeleteLinkRequest) returns (DeleteLinkResponse)`

## Behavior on mode detection

On startup:
1. Parse config file
2. If `topology` section is present → **config mode**
   - Wipe segments/links from DB
   - Seed DB from expanded config
   - Mark control plane as config-managed (reject mutation APIs)
3. If `topology` section is absent → **API mode**
   - Load existing state from DB
   - Accept all CRUD APIs
   - If DB is empty, the system starts with no topology (operator uses APIs to build it)

micpapal:
# Description

## Summary

Implements dual topology management modes for the SLIM control plane: **config-managed** (YAML file is source of truth) and **API-managed** (DB is source of truth with full CRUD via gRPC/CLI). Adds topology mutation commands, persistence for config→API mode transitions, and controller restart resilience.

## Topology management modes

The control plane operates in one of two mutually exclusive modes based on whether the config file contains a `topology` section:

### Config mode (topology section present)

- Config file is the **single source of truth**
- On restart: all state (runtime + topology) is wiped and config is re-applied
- Dynamic `$group` expansion re-evaluated when new nodes register
- Mutation APIs return an error: _"topology is config-managed"_
- Topology is persisted to DB after expansion (for seamless transition to API mode)

### API mode (no topology section)

- DB is the **single source of truth**
- On restart: runtime state (nodes, links, routes) is cleared but topology config (segments, segment links) is preserved from DB
- Full CRUD operations available via `slimctl` and gRPC APIs
- Nodes re-register and links/routes are rebuilt automatically

### Topology persistence and startup logic

- **`clear_runtime_state()`** — new DB method that clears nodes, links, routes but preserves topology config. Called on API-mode startup.
- **`clear_all_state()`** — clears everything (runtime + topology). Called on config-mode startup. Internally delegates to `clear_runtime_state()`.
- **`persist_segments_to_db()`** — in config mode, writes expanded segments/links to DB after `rebuild_link_graph()`, enabling config→API mode transition without losing topology
- **`load_topology_from_db()`** — on API-mode startup, reconstructs in-memory segment graphs from DB

### Topology mutation APIs (API mode only)

- **`controller segment add <NAME>`** — create a new segment
- **`controller segment remove <NAME>`** — remove a segment and all its links
- **`controller link add --segment <S> <GROUP_A> <GROUP_B>`** — add inter-group link to a segment
- **`controller link remove --segment <S> <GROUP_A> <GROUP_B>`** — remove inter-group link from a segment
- Corresponding gRPC RPCs: `AddSegment`, `RemoveSegment`, `AddTopologyLink`, `RemoveTopologyLink`
- Guards prevent mutation in config-managed mode

### Node reconnection resilience

- On controller restart, nodes reconnect and rebuild their links/routes automatically
- Stale nodes (that don't reconnect) are no longer persisted across restarts — cleared by `clear_runtime_state()`
- Node disconnect logic properly cleans up routes and reassigns gateway links

### DB schema additions

- **`topology_segments`** table — stores segment definitions
- **`topology_segment_links`** table — stores inter-group links within each segment
- SQLite `add_link_to_segment` made idempotent (duplicate links ignored)
- Links allowed for groups that don't have registered nodes yet

## Related

- Closes #1768

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


crates/control-plane/src/route_service/routes.rs
```diff
@@ -66,11 +79,57 @@ impl super::RouteService {
         if groups_changed {
             let group_vec: Vec<&str> = new_groups.into_iter().collect();
             *current_segments = self.0.topology.build_graph(&group_vec);
+
+            // Persist the expanded topology to DB so that API mode can pick it
+            // up if the operator later switches modes.
+            tracing::debug!("groups changed, persisting topology to DB");
+            self.persist_segments_to_db(&current_segments).await;
```
muscariello: **Bug: `segment_graphs` write guard held across async DB I/O**

`current_segments` is the `tokio::sync::RwLock` write guard on `segment_graphs`, acquired at the top of `rebuild_link_graph`. It is still live here when `persist_segments_to_db` is called, and that function awaits multiple DB round-trips (one per segment, one per edge). While awaiting, the write lock is held, starving every concurrent reader: `allowed_link_pairs`, `list_segments`, and the two `expand_wildcard_routes` helpers all block on `segment_graphs.read().await` — exactly the hot paths that run while nodes are registering.

Suggested fix: clone the graph data before releasing the guard, then persist without holding the lock:
```rust
let segments_snapshot: Vec<_> = current_segments.clone();
drop(current_segments);          // release write lock before any I/O
self.persist_segments_to_db(&segments_snapshot).await;
```

crates/control-plane/src/route_service.rs
```diff
@@ -134,6 +136,310 @@ impl RouteService {
...
+            if !allowed_pairs.contains(&pair) {
+                tracing::info!(
+                    "reconcile_topology_change: removing disallowed link {}↔{} (groups {}↔{})",
+                    link.source_node_id,
+                    link.dest_node_id,
+                    link.source_group,
+                    link.dest_group,
+                );
+                // Delete routes that depended on this link.
+                if let Ok(routes) = self.0.db.get_routes_by_link_id(&link.link_id).await {
```
muscariello: **Bug: DB error silently swallowed — routes orphaned, source nodes not re-queued**

When `get_routes_by_link_id` returns `Err`, the `if let Ok(routes)` arm is skipped entirely. The physical link is still deleted on the next line, but:

1. Routes pointing to that `link_id` are never deleted from the DB.
2. `route.source_node_id` is never inserted into `reconcile_nodes`, so those nodes are not queued for reconciliation.

The result: nodes retain stale forwarding state for a link that no longer exists. The reconciler encounters the orphaned route records on every sweep, logs a warning about a missing link, and skips them — there is no garbage-collection path for this case.

Suggested fix — treat the failure as unrecoverable for this link and skip deletion so the operation can be retried:
```rust
match self.0.db.get_routes_by_link_id(&link.link_id).await {
    Ok(routes) => {
        for route in &routes {
            if let Err(e) = self.0.db.delete_route(&route.id).await {
                tracing::error!("reconcile_topology_change: delete_route: {e}");
            }
            reconcile_nodes.insert(route.source_node_id.clone());
        }
    }
    Err(e) => {
        tracing::error!("reconcile_topology_change: get_routes_by_link_id: {e}; skipping link deletion");
        continue;  // retry on next reconcile cycle
    }
}
```

crates/control-plane/src/route_service/routes.rs
```diff
@@ -66,11 +79,57 @@ impl super::RouteService {
...
         }
 
         groups_changed
     }
 
+    /// Write the current segment graphs to the DB (topology_segments +
+    /// topology_segment_links). Called after config-mode graph expansion so
+    /// the DB always reflects the latest resolved topology.
+    async fn persist_segments_to_db(
```
muscariello, line 100: **Bug: `persist_segments_to_db` only adds edges, never removes stale ones**

This function is called from `rebuild_link_graph` whenever the registered group set changes. It upserts edges from the current (possibly smaller) graph, but never deletes rows that are no longer in the graph. When a group loses its last node and disappears, the corresponding `topology_segment_links` rows survive indefinitely.

This matters if an operator switches from config mode to API mode without wiping the DB: `load_topology_from_db` on startup will see the stale config-derived pairs as authoritative API-managed topology, permitting cross-group traffic between groups that were intentionally removed from the config.

Suggested fix — after upserting edges, delete any rows for this segment that are not in the new graph:
```rust
// After the add_link_to_segment loop:
let current_edges: HashSet<(String, String)> = graph
    .edge_references()
    .map(|e| (graph[e.source()].clone(), graph[e.target()].clone()))
    .collect();
let stored = self.0.db.get_links_for_segment(&seg.id).await?;
for (src, dst) in stored {
    if !current_edges.contains(&(src.clone(), dst.clone()))
        && !current_edges.contains(&(dst.clone(), src.clone()))
    {
        let _ = self.0.db.delete_link_from_segment(&seg.id, &src, &dst).await;
    }
}
```

crates/control-plane/src/db/sqlite/mod.rs
```diff
@@ -1013,6 +1014,212 @@ impl DataAccess for SqliteDb {
 
         Ok(link)
     }
+
+    // ── Topology (API-managed mode) ───────────────────────────────────────
+
+    async fn create_segment(&self, name: &str) -> Result<TopologySegment> {
```
muscariello, line 1020: **Bug: check-then-insert TOCTOU — loser returns a UUID never stored in the DB**

The pattern is:
1. `get_segment_by_name` → `None`
2. Generate `uuid = Uuid::new_v4()`
3. `INSERT ... ON CONFLICT DO NOTHING`
4. Return `Ok(segment)` with the locally-generated `uuid`

If two callers race, both see `None` in step 1. The losing insert is silently dropped by `ON CONFLICT DO NOTHING`. The loser returns a `TopologySegment` whose `id` does not exist in the DB. Any subsequent `add_link_to_segment` using that ID will violate the FK constraint on `topology_segment_links.segment_id`.

Suggested fix — re-fetch after a silent conflict to return what was actually stored:
```rust
async fn create_segment(&self, name: &str) -> Result<TopologySegment> {
    let segment = TopologySegment {
        id: uuid::Uuid::new_v4().to_string(),
        name: name.to_string(),
        created_at: SystemTime::now(),
    };
    let mut conn = self.pool.get().await...?;
    let affected = diesel::insert_into(topology_segments::table)
        .values(segment)
        .on_conflict_do_nothing()
        .execute(&mut conn)
        .await...?;
    if affected == 0 {
        // Another writer won the race — return the row they stored.
        self.get_segment_by_name(name).await?
            .ok_or_else(|| Error::DbError { context: "create_segment", msg: "row vanished".into() })
    } else {
        // Re-fetch to get the authoritative stored record.
        self.get_segment_by_name(name).await?
            .ok_or_else(|| Error::DbError { context: "create_segment", msg: "insert lost".into() })
    }
}
```

muscariello (changes requested):
There are a few comments to consider before merging

muscariello (approved):
LGTM

PR #1775, opened 2026-06-29
feat: add websocket supports for the browser

hackeramitkumar:
# Description

Adds first-class **browser (`wasm32-unknown-unknown`) support** to
the SLIM data plane so a web client can join a channel over WebSocket and
interoperate with native gRPC and native-WebSocket participants in the same group.

Previously the data plane compiled only for native targets. This PR makes the
transport-neutral core (connection state, forwarding, link negotiation, sync,
message processing) build for wasm, adds a browser WebSocket transport
backed by `gloo-net`, and swaps the native-only `aws-lc-rs` crypto for
pure-Rust backends on wasm — while leaving every native code path unchanged.

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


crates/datapath/src/websocket/stream_wasm.rs
```diff
@@ -0,0 +1,119 @@
...
+                next = rx.recv() => next,
+            };
+
+            let msg = match next {
+                Some(msg) => msg,
+                None => break,
+            };
+
+            let payload = msg.encode_to_vec();
+            if let Err(err) = sink
```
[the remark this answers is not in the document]
hackeramitkumar: It's being used.

crates/datapath/src/websocket/stream_wasm.rs
```diff
@@ -0,0 +1,119 @@
...
+                                    err
+                                ))))
+                                .await;
+                        }
+                    }
+                }
+                Some(Ok(gloo_net::websocket::Message::Text(_))) => {
+                    warn!("ignoring text websocket frame, expected binary protobuf frame");
+                }
+                Some(Err(err)) => {
```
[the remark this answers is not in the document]
hackeramitkumar: It's being used

crates/auth/Cargo.toml
```diff
@@ -67,6 +67,9 @@ wiremock = { workspace = true }
 # `rand` works in the browser.
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 getrandom = { workspace = true }
+# ed25519-dalek/p256 still use rand_core 0.6, which pulls getrandom 0.2.
+# Enable its browser backend in addition to workspace getrandom 0.3.
+getrandom_02 = { package = "getrandom", version = "0.2", features = ["js"] }
```
msardara, line 72: I think we can remove this

crates/config/src/schema/generate_schema.rs
```diff
@@ -1,23 +1,31 @@
-use schemars::{JsonSchema, schema_for};
-use slim_config::client::ClientConfig;
-use slim_config::server::ServerConfig;
-use std::fs::File;
-use std::io::Write;
-use std::path::PathBuf;
+// The schema generator depends on the native-only `ServerConfig` and filesystem
```
msardara, line 4: Here we miss the header:

```
// Copyright AGNTCY Contributors (https://github.com/agntcy)
// SPDX-License-Identifier: Apache-2.0
```

msardara (changes requested):

msardara (changes requested):
Great job Amit! Overall this is a **high-quality, carefully-engineered PR**.

- `TransportChannel<G>` + the `WebSocketStreams { inbound, outbound }` shape let the native (`fastwebsockets`) and browser (`gloo-net`) transports share the entire `register_remote_connection` / processing machinery — the wasm branch in `message_processing.rs` stays small.
- `to_channel()` on wasm returns `TransportChannel<Infallible>`, and the gRPC arm is discharged elegantly with `TransportChannel::Grpc(never) => match never {}`.
- The crypto backend swap (`aws-lc-rs` → `hmac`/`sha2` + `x25519-dalek`/`hkdf`) is the right call, and compiling the pure-Rust backends into **native test builds** with **parity tests** (`pure_hmac_backend_matches_production_backend`, `pure_ecdh_backend_peers_agree`) directly guarantees a native↔browser link interoperates — exactly the property that matters.
- Nice detail: capping the wasm "infinite" sleep at `i32::MAX` ms to respect JS `setTimeout`'s 32-bit delay limit.

---

### Correctness / coverage

**1. The actual browser WebSocket transport has no test coverage (notable gap).**
The new wasm tests cover only crypto, `runtime::spawn`, and transport *selection*:
- `wasm_client_test.rs` only asserts gRPC endpoints are rejected — it never opens a socket.
- `wasm_browser_test.rs` covers HMAC/ECDH and the spawner.

Neither `websocket/stream_wasm.rs` (the read/write loops) nor `client_wasm.rs::to_websocket_channel` is exercised on **any** target. This is the newest, riskiest code (protobuf framing, cancellation coordination, error mapping) and it ships untested. With the `web-client` crate removed there's also no end-to-end browser smoke test. At minimum a follow-up issue; ideally a minimal loopback-WS `wasm_bindgen_test`.

**2. Possible send-before-open race with `gloo_net::WebSocket::open` (please verify).**
Unlike the native path — where `to_websocket_channel` completes the handshake before returning — `WebSocket::open()` returns immediately while the socket is still `CONNECTING`. `stream_wasm::spawn_transport_tasks` then splits it and the write loop may call `sink.send(...)` as soon as link negotiation produces its first outbound frame. `gloo-net` does not buffer pre-open sends, so the JS `send()` can throw `InvalidStateError` → `MessageSendError`, breaking the write loop before negotiation completes. Since negotiation is client-initiated here, this is reachable. Please confirm in a real browser that the first negotiation frame isn't dropped; if it is, awaiting the open event inside `to_websocket_channel` before returning the channel would fix it.

---

### Isolation of wasm vs. non-wasm code

The target split currently shows up three different ways — inline `#[cfg]` on individual items, `cfg_if!` blocks around large chunks, and paired cfg'd function bodies — and the mixing is heaviest in `header_mac.rs`, `link_ecdh.rs`, `client.rs`, and the session timer code. Suggestions to tighten it, by impact:

**a) Backend-module pattern for the crypto (biggest win).**
`header_mac.rs` interleaves six cfg-gated free functions plus a cfg'd type alias to abstract one HMAC backend. Collapse each backend into its own module exposing an identical surface, then select once:

```rust
#[cfg(not(target_arch = "wasm32"))]
#[path = "header_mac/backend_awslc.rs"]
mod backend;
#[cfg(target_arch = "wasm32")]
#[path = "header_mac/backend_pure.rs"]
mod backend;

type MacKey = backend::MacKey;
// self.key = backend::new(secret);  backend::sign(&self.key, buf);  ...
```

Now `HeaderMacSession` and the whole signing path carry **zero** `#[cfg]`. Same for `link_ecdh.rs`: move the `aws_lc_rs` bodies of `generate_x25519_ephemeral`/`derive_header_mac_from_ecdh` and the `EphemeralKey` alias into `backend_awslc.rs`, and the existing `pure_ecdh` into `backend_pure.rs`. The public fns become one-line forwarders. The "compile pure backend into native tests for parity" trick still works via a `#[cfg(any(target_arch = "wasm32", test))]` alias used only by the parity test.

**b) Collapse `ClientConfig`'s native-only methods into one gated `impl` block.**
`client.rs` puts `#[cfg(not(target_arch = "wasm32"))]` on ~7 builder/helper methods individually (`with_compression`, `with_tls_setting`, `with_proxy`, `with_auth`, `with_backoff`, `with_metadata`, `retry_connect`, native `to_channel`). Move them into a single `#[cfg(not(target_arch = "wasm32"))] impl ClientConfig { ... }` — one attribute instead of eight, and it reads as "the native surface." (The struct *fields* still need individual gates — their types are native-only — but grouping them under a comment banner beats the current scatter.)

**c) A `platform`/`runtime` module for target-conditional primitives.**
You already did this for spawning (`runtime.rs`). Absorb the inline timer cfgs the same way: the shutdown-deadline reset + `Duration::MAX` vs `i32::MAX ms` sleep in `session_controller.rs`, and the `Delay::new` vs `tokio::time::sleep` in `subscription_manager.rs`. Hoist to intent-named helpers (`runtime::infinite_sleep()`, `runtime::ack_timeout_delay()`) so business logic carries no `#[cfg]` and the `setTimeout` rationale lives in one documented place.

**d) Prefer file-split over `cfg_if!` for whole-item duplication.**
The PR already has the right template (`websocket.rs` selecting `client` vs `client_wasm`). Extend it to the two remaining in-file duplications: the two `impl Display for ClientConfig` bodies, and the two-copy `ConfigError` in `errors.rs` (move the `mod wasm` variant to `errors_wasm.rs`, native to `errors.rs`, select with one `cfg_if!`). File-level splits diff and review more cleanly than 60-line `cfg_if!` blocks, and make "what exists on wasm" answerable by `ls`.

Net effect: the leaf modules become target-agnostic, and every `#[cfg(target_arch = "wasm32")]` migrates to either a module-selection line at the top of a file or an unavoidable struct-field gate.

---

### Verdict

Architecturally sound and safe on the native side (native paths untouched; native tests exercise the new pure-Rust crypto). Main reservations before merge: **untested transport code (1)** and the **send-before-open question (2)** — I'd want (2) confirmed against a real browser, and (1) tracked as a follow-up. Also the isolation of wasm vs non-wasm code will greatly improve readability.

crates/auth/Cargo.toml
amitami2: I tried removing it, but the wasm build breaks without it.

The issue is that p256 still pulls in an older version of getrandom (0.2) through its dependencies. On wasm, that version needs the js feature enabled, or compilation fails. Our workspace getrandom 0.3 setup only covers the code we call directly - it doesn't fix the older one that p256 brings in transitively.

So getrandom_02 is basically there to turn on the browser backend for that older dependency. Once the RustCrypto crates (p256, elliptic-curve, etc.) move to the newer random stack, we can drop it.

hackeramitkumar:
> ### Isolation of wasm vs. non-wasm code

Thanks @msardara for the review. I agree with all these points and have refactored the code isolation as you suggested.

For points 1 and 2, I've added a few test cases, but they require a browser driver and a running WS echo server. So once these pass in CI, we can be confident about both points; if they don't, I'll fall back to adding end-to-end test cases.

Initially I had added a complete browser web-client application and validated everything manually, but I later decided that rather than maintaining an extra wasm-bindgen application here, we'll add example applications in the bindings repo instead.

crates/config/src/schema/generate_schema.rs
hackeramitkumar: done

msardara (approved):
LGTM, thanks @hackeramitkumar!

PR #1778, opened 2026-06-30
fix: separate server connection config

### LINKED ISSUE #1640: Separate server-side connection config from client-specific config in slim node
## Description

Currently, when a slim node runs as a server (accepting connections from other slim nodes), it provides the full connection configuration that a client would need to connect. This config is passed to the controller, which uses it to create new connections from newly registered nodes.

## Problem

The current approach has limitations because the connection config may be client-specific. Different clients may need:
- Different authentication credentials
- Proxy configuration
- Other client-local settings

Storing full client-specific configuration on the controller side is problematic — it leaks secrets and doesn't scale across heterogeneous clients.

## Proposed Solution

**Server-side config should only contain what is required to reach the server**, specifically:
- The IP address (or hostname) to connect to
- Whether TLS is required or not
- Which authentication method is required (if any)

This minimal server-side config should **not** include any client-specific details (credentials, proxy settings, etc.). This information must be provided locally by each client.

**Connection flow:**
1. The server-side slim node exposes its connection requirements to the controller (address, TLS, auth method).
2. When a new slim instance needs to connect, it receives the server-side config from the controller.
3. The new slim instance **merges** the server requirements with its own local configuration (credentials, proxy, etc.) to establish the connection.
4. Once the link is created and has an ID, the controller stores only the **link ID** and the **server-side config** in the DB — no client secrets are persisted.

markpmarton:
closes #1640 

# Description

The PR separates server-side connection requirements from client connection policy by introducing `ServerConnectionConfig` as a dedicated type, distinct from `ClientConfig`.

Previously, the full `ClientConfig` (including client-local settings like Spire socket paths, backoff, keepalive) was serialized and stored in the database and passed over the wire. Now the control plane stores and transmits only what the server requires — endpoint, TLS flag, and auth method — while the controller merges those requirements onto its own local ClientConfig when opening a connection.

# Changes
- Dedicated `ServerConnectionConfig` type containing only remote connection related fields
- In `ConnectionDetails` db schema `SpireMtls` nested field is replaced with three flat fields (`tls_required`, `auth_method` and `spire_trust_domain`)
- `generate_config_data` simplified: unused `src_node` removed
- `tls_required == true` is now mandatory if  `auth_method` is Spire

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [x] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


crates/control-plane/src/db/model.rs
```diff
@@ -220,7 +219,7 @@ pub fn has_connection_details_changed(
             None => return true,
             Some(ncd) => {
                 if ecd.external_endpoint != ncd.external_endpoint
-                    || ecd.spire_mtls != ncd.spire_mtls
+                    || ecd.spire_trust_domain != ncd.spire_trust_domain
```
micpapal: here I think we need to check also the tls_required and auth_method that may have changed after the reconnections

crates/control-plane/src/db/model.rs
micpapal: this struct is not used anymore I think as the mtls option is not there anymore

crates/controller/src/service.rs
```diff
@@ -195,15 +196,16 @@ pub(crate) fn from_server_config(server_config: &ServerConfig) -> ConnectionDeta
         }
     }
 
-    let spire_mtls = if !server_config.tls_setting.insecure || spire_socket_path.is_some() {
-        Some(v1::connection_details::SpireMtls {
-            socket_path: spire_socket_path.unwrap_or_default(),
-            trust_domain,
-        })
+    let tls_required = !server_config.tls_setting.insecure || spire_socket_path.is_some();
+    let auth_method = if spire_socket_path.is_some() || trust_domain.is_some() {
```
micpapal: here we should also match AuthenticationConfig::Spire(_) in the if

```suggestion
let auth_method = if spire_socket_path.is_some()
      || trust_domain.is_some()
      || matches!(server_config.auth, AuthenticationConfig::Spire(_))
```

crates/control-plane/src/route_service/links.rs
```diff
@@ -192,7 +193,12 @@ impl super::RouteService {
                 } else {
                     Uuid::new_v4().to_string()
                 };
-                (ep, cd, lid, LinkStatus::Pending)
+                (
+                    ep,
+                    ServerConnectionConfig::from_client_config(&cd),
```
Tehsmash: Do you think we need to compute_client_config any more inside of the controller? Something I've realised is that its generating keepalive, backoff etc settings, which might be different for different clients/nodes connecting so really should maybe be defined by the node itself not the controller.

https://github.com/markpmarton/slim/blob/28f6c3a57d62813b243505fc77ae7c15f23e9c5c/crates/control-plane/src/route_service/connection_config.rs#L160-L173

If this is the case perhaps the controller only operates on ServerConnectionConfigs, and nodes receive ServerConnectionConfigs and perform the merge on their end with these other settings like the KeepAlive etc. WDYT?

crates/proto/proto/controller/v1/controller.proto
```diff
@@ -126,15 +126,12 @@ message Node {
...
-    string socket_path = 1;
-    optional string trust_domain = 2;
-  }
-
   string endpoint = 1;
   optional string external_endpoint = 2;
-  optional SpireMtls spire_mtls = 3;
-  google.protobuf.Struct metadata = 4;
+  bool tls_required = 3;
+  string auth_method = 4;
```
micpapal: use to a proto enum instead of string so we can use it also in the code instead to compare strings to get the auth_metod

micpapal (changes requested):
Can you also add an example of config file with the new fields? Maybe by updating some of the example config files in /crates/control-plane/tests/config-files/nodes You an also add a new one there with the possible setups commented as reference.

crates/control-plane/src/route_service/links.rs
markpmarton: Yes, I think moving the config merge make sense if we want to use the node configs for keepalive and backoff.
Although I think there might be cases where having synced keepalive/backoff can be useful (e.g. low timeout value set in FW rules). For this reason I would add optional fields in ServerConnectionConfig to override local settings if set.

Tehsmash: Ah interesting yes that makes sense, so the controller can say "I know this server is behind a crappy nginx therefore you should set this keepalive setting"

crates/control-plane/src/db/model.rs
markpmarton: `has_connection_details_changed` is now simplified, all fields of the ConnectionDetails are checked.

crates/control-plane/src/route_service/links.rs
markpmarton: I checked the current implementation and I think moving the connection related configs causes quite a few changes unrelated to this PR. I'll open another issue dedicated to this one.

EDIT: Issue is created [here](https://github.com/agntcy/slim/issues/1794)

crates/control-plane/src/route_service/connection_config.rs
```diff
@@ -118,71 +114,35 @@ pub(super) fn generate_config_data(
...
-        if let Some(td) = trust_domain {
-            trust_domains.push(td.to_string());
-        }
-
-        let spire_config = slim_config::auth::spire::SpireConfig {
-            socket_path: Some(spire_socket.clone()),
-            trust_domains: trust_domains.clone(),
-            ..Default::default()
+    let (effective_endpoint, tls_required, auth_method) =
+        if detail.tls_required && detail.auth_method == model::AuthMethod::Spire {
```
micpapal: Here TLS can be enable also when Spire is off. For example also using basic auth or jwt we should be able to set TLS. In addition in the else message everything is set to AuthMethod::None while it should be equal to the one passed in the function.

crates/controller/src/service.rs
```diff
@@ -682,38 +689,45 @@ impl ControllerService {
...
-                Ok(client_config) => {
-                    match self
-                        .inner
-                        .message_processor
-                        .connect(client_config.clone(), None, None)
-                        .await
-                    {
-                        Err(e) => {
+                Ok(server_config) => {
+                    let mut client_config = ClientConfig::default();
```
micpapal: The  ClientConfig::default()  here gives  TlsSource::None , so  merge_server_requirements  will always resolve  tls_socket_path  to None. I don't know if the socket field should be populated by the data plane node with its own config or if this is actually a problem.

crates/config/src/client.rs
```diff
@@ -509,6 +510,107 @@ impl ClientConfig {
...
+
+#[derive(Debug, Default, Serialize, Deserialize, Clone, PartialEq, Eq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum RequiredAuthMethod {
+    #[default]
+    None,
+    Basic,
+    Jwt,
+    Spire {
+        trust_domain: Option<String>,
```
micpapal, line 624: SpireAuthConfig.trust_domains is a Vec<String> but here we store only one. Do we need a vec also here?

crates/control-plane/src/route_service/connection_config.rs
markpmarton: You're right, I separated the TLS check and fixed the auth method parse.

crates/controller/src/service.rs
markpmarton: Also right, I just figured out I only focused on the spire auth method, and kept the jwt and basic auth secret processing unfinished... Now all auth should work correctly, also added integration tests for each.

crates/config/src/client.rs
markpmarton: I don't think we need to handle more trust_domain in the RequiredAuthMethod as only one domain should be received from the control-plane for each node registration. Now the multi-domain handling is covered with the `outbound_clients` in Config.

micpapal (approved):

PR #1795, opened 2026-07-06
feat: group registration via slimctl

micpapal:
# Description

Adds CRUD operations for managing registration auth groups at runtime (API mode only). Previously, groups and their secrets could only be configured via the static YAML config file. Now operators can dynamically add/remove groups without restarting the control plane.

### Key Changes

**Proto & API:**
- Added `ListGroups`, `AddGroup`, `RemoveGroup` RPCs to the northbound `ControlPlaneService`
- `ListGroups` merges auth-configured groups (even those with no connected nodes) with connected node information

**Control Plane:**
- `GroupAuthenticator` refactored to use `Arc<RwLock<HashMap>>` for interior mutability — enables concurrent read (southbound auth checks) and write (northbound group management)
- `add_group`: validates input → builds verifier (fails fast on bad secret) → persists to DB
- `remove_group`: removes verifier (gates new connections) → deletes from DB → disconnects all nodes in the group → deregisters them (cleans up links/routes)
- On startup in API mode, DB-persisted secrets are restored into the live authenticator

**Database:**
- Added `registration_secrets` table (group_name PK, secret, created_at)
- 5 new `DataAccess` trait methods with in-memory and SQLite implementations

**slimctl CLI:**
- `slimctl controller group list` — shows all groups and their connected nodes
- `slimctl controller group add <name> <secret>` — registers a new group
- `slimctl controller group rm <name>` — removes a group (disconnects all its nodes)

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


crates/control-plane/src/auth.rs
```diff
@@ -127,6 +137,39 @@ impl GroupAuthenticator {
...
+                "add_verifier is only supported for shared_secret auth",
+            )),
+        }
+    }
+
+    /// Remove the verifier for a group (shared secret mode only).
+    pub fn remove_verifier(&self, group_name: &str) {
+        if let Self::SharedSecret { verifiers } = self {
+            verifiers.write().remove(group_name);
+        }
```
Tehsmash, line 177: Should we have an error message for the other verifier types too?

Tehsmash:
> ListGroups merges auth-configured groups (even those with no connected nodes) with connected node information

Does this mean that even with Auth groups configured we can have nodes register with the control plane with an unknown group? I had assumed that if group auth was enabled that only known groups would be allowed.

This raises the question: if I have a node register with group "Foo" and it joins the network, then I add/convert group "Foo" to an auth group, will the node get kicked, assuming that it doesn't have the secret?
There are two consequences I thought of depending on how we handle this:
a) if the control plane blocks you from converting a node registered group to an auth group, then you can squat on group names
b) if we don't kick the node, then we have a case were a node could guess a group name which will be created later and insert himself before the group is registered with the control plane.

crates/control-plane/src/services/northbound.rs
```diff
@@ -541,12 +547,134 @@ impl ControlPlaneService for NorthboundApiService {
...
+    // ── Registration auth group management ─────────────────────────────────
+
+    async fn list_groups(
+        &self,
+        _request: Request<ListGroupsRequest>,
+    ) -> Result<Response<ListGroupsResponse>, Status> {
+        // Start with auth groups (from DB secrets).
+        let auth_groups = self
+            .db
+            .list_registration_secret_groups()
```
msardara, line 568: `list_groups` sources the group set from the `registration_secrets` DB table, but in config-managed mode `build_authenticator` loads secrets from YAML straight into the in-memory verifiers map and never writes them to the DB (the DB-restore block in `server.rs` is gated on `if is_api_managed`).

The proto comment promises "In shared_secret mode: returns groups with configured/stored secrets", yet in config mode a configured group with no connected nodes won't appear here at all — only groups that happen to have a connected node show up. Consider sourcing the configured set from the authenticator's verifiers map so config and API mode behave consistently.

crates/control-plane/src/services/northbound.rs
```diff
@@ -541,12 +547,134 @@ impl ControlPlaneService for NorthboundApiService {
...
+            return Err(Status::invalid_argument("secret must not be empty"));
+        }
+
+        // Build the verifier first (validates the secret is usable).
+        self.authenticator
+            .add_verifier(&req.group_name, &req.secret)?;
+
+        // Persist to DB only after the verifier was successfully built.
+        self.db
+            .upsert_registration_secret(&req.group_name, &req.secret)
```
msardara, line 621: `add_verifier` mutates the in-memory verifiers map first, then this `upsert_registration_secret` persists. If the DB write fails, the function returns an error but the verifier is left active. The caller sees a failure and assumes the group wasn't created, yet nodes presenting that secret can now register — until the next restart, when the unpersisted verifier silently disappears. Consider rolling back the verifier (`remove_verifier`) on DB failure.

crates/control-plane/src/services/northbound.rs
```diff
@@ -541,12 +547,134 @@ impl ControlPlaneService for NorthboundApiService {
...
+    ) -> Result<Response<RemoveGroupResponse>, Status> {
+        self.route_service.ensure_api_mode()?;
+
+        let req = request.get_ref();
+        if req.group_name.is_empty() {
+            return Err(Status::invalid_argument("group_name must not be empty"));
+        }
+
+        // Find all nodes belonging to this group.
+        let nodes = self
```
msardara: Race window: the node list is snapshotted here *before* `remove_verifier` is called below. A node that finishes registering in the gap between this `list_nodes` and `remove_verifier` (verifier still present → auth passes) won't be in `group_nodes` and therefore never gets force-disconnected, leaving it attached to a group that no longer exists. Removing the verifier first, then listing and disconnecting, closes the window.

crates/control-plane/src/services/northbound.rs
```diff
@@ -541,12 +547,134 @@ impl ControlPlaneService for NorthboundApiService {
...
+            .list_nodes()
+            .await
+            .map_err(|e| Status::internal(format!("failed to list nodes: {e}")))?;
+        let group_nodes: Vec<_> = nodes
+            .iter()
+            .filter(|n| n.group_name.as_deref() == Some(&req.group_name))
+            .collect();
+
+        // Remove the verifier first to prevent new registrations.
+        self.authenticator.remove_verifier(&req.group_name)?;
```
msardara, line 651: Unlike `add_group`, which checks `is_shared_secret()` up front, `remove_group` reaches `remove_verifier` only after `ensure_api_mode` + `list_nodes`. For `Spire`/`Noop` auth in API mode this returns `UNIMPLEMENTED` after doing work — and the proto only documents `FAILED_PRECONDITION` (config mode), so the SPIRE-mode error code is undocumented. An up-front `is_shared_secret()` check would make it symmetric with `add_group`.

crates/control-plane/src/server.rs
```diff
@@ -59,15 +59,37 @@ impl ControlPlane {
...
         };
 
+        // In API mode, restore DB-persisted secrets into the live authenticator.
+        if is_api_managed && authenticator.is_shared_secret() {
+            let groups = db.list_registration_secret_groups().await?;
+            for group in &groups {
+                let secret = db
+                    .get_registration_secret(group)
+                    .await?
+                    .expect("group listed but secret missing from DB");
```
msardara: `.expect(...)` panics the entire control-plane startup if a listed group has no secret row. The race is effectively impossible at single-threaded startup so severity is low, but a graceful `map_err`/`continue` (log and skip) would avoid crashing the process on unexpected DB state.

msardara (changes requested):

micpapal:
> > ListGroups merges auth-configured groups (even those with no connected nodes) with connected node information
> 
> Does this mean that even with Auth groups configured we can have nodes register with the control plane with an unknown group? I had assumed that if group auth was enabled that only known groups would be allowed.
> 
> This raises the question: if I have a node register with group "Foo" and it joins the network, then I add/convert group "Foo" to an auth group, will the node get kicked, assuming that it doesn't have the secret? There are two consequences I thought of depending on how we handle this: a) if the control plane blocks you from converting a node registered group to an auth group, then you can squat on group names b) if we don't kick the node, then we have a case were a node could guess a group name which will be created later and insert himself before the group is registered with the control plane.

The comment could have been clearer. To clarify: a node can only connect to the control plane if it belongs to a pre-configured, authenticated group. What I meant is that you can register a group (and its secret) before any node from that group actually connects.

So if you run  ```slimctl controller group list ```, you might see:
```
GROUP       NODES
group-1     node-1, node-2
group-x     (no nodes connected)
```
 group-x  is fully configured and ready to accept nodes, it just doesn't have any connected yet. When a node from  group-x  connects with valid credentials, it will be accepted. Conversely, if a node claims to belong to  group-y  which hasn't been registered, the connection will be rejected regardless of what credentials it presents.

msardara (approved):
LGTM

PR #1809, opened 2026-07-07
feat(session): use uuid for channel ids

micpapal:
# Description

Replace fixed `DATA_CHANNEL_ID` and `CONTROL_CHANNEL_ID` constants with identity-derived UUIDs for multicast channel names.

Previously, multicast channels used hardcoded sentinel IDs (`u128::MAX - 2` and `u128::MAX - 3`) which caused name collisions when two channels shared the same `org/ns/channel` prefix. This change derives unique data and control channel IDs by hashing the moderator's identity token together with the channel name using `XxHash3_128`, mirroring the approach already used for agent name IDs.

## Key changes

- **Moderator side**: Derives unique data/control channel IDs from `identity_token + destination_name` when creating a multicast session
- **JoinRequest proto**: Now carries both `channel` (data) and `control` names explicitly so participants use the exact same IDs
- **Participant side**: Extracts data/control names directly from the JoinRequest payload
- **SessionBuilder**: Validates that multicast sessions have distinct, non-null data and control IDs with matching prefixes

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [x] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/crates/service/src/app.rs
+++ b/crates/service/src/app.rs
@@ -451,8 +451,8 @@ mod tests {
     use crate::SubscriptionAckError;
     use slim_auth::shared_secret::SharedSecret;
     use slim_datapath::api::{
-        CommandPayload, NameId, Participant, ParticipantSettings, ProtoMessage,
-        ProtoSessionMessageType, ProtoSessionType,
+        CommandPayload, Participant, ParticipantSettings, ProtoMessage, ProtoSessionMessageType,
+        ProtoSessionType,
     };
     use slim_datapath::messages::utils::SlimHeaderFlags;
     use slim_session::session_config::MlsSettings;
@@ -807,7 +807,7 @@ mod tests {
 
         // send join_request message to create the session
         let payload = CommandPayload::builder()
-            .join_request(None, None, None, None)
+            .join_request(None, None, None, None, None)
             .as_content();
 
         let mut join_request = Message::builder()
@@ -1224,10 +1224,10 @@ mod tests {
             // Verify it's a multicast session
             assert_eq!(session_arc.session_type(), ProtoSessionType::Multicast);
 
-            // For multicast sessions, the destination is the channel name with DATA_CHANNEL_ID
+            // For multicast sessions, the destination is the channel name
             let dst = session_arc.dst();
-            let expected_dst = channel_name.clone().with_id(NameId::DATA_CHANNEL_ID);
-            assert_eq!(dst, &expected_dst);
+            let expected_dst = channel_name.clone();
+            assert!(dst.match_prefix(&expected_dst));
 
             total_received_sessions += 1;
         }
--- a/crates/service/src/app.rs
+++ b/crates/service/src/app.rs
@@ -1228,6 +1228,10 @@ mod tests {
             let dst = session_arc.dst();
             let expected_dst = channel_name.clone();
             assert!(dst.match_prefix(&expected_dst));
+            assert!(
+                dst.has_id(),
+                "destination should have a derived ID, not NULL_COMPONENT"
+            );
 
             total_received_sessions += 1;
         }
--- a/crates/session/src/session_layer.rs
+++ b/crates/session/src/session_layer.rs
@@ -248,7 +248,40 @@ where
         let is_p2p = session_config.session_type == ProtoSessionType::PointToPoint;
         let destination_proto = destination.clone();
 
-        let session = self.create_session_internal(session_config, local_name, destination, id)?;
+        let (dest_with_id, control_with_id) =
+            if session_config.session_type == ProtoSessionType::Multicast {
+                let token_id = self
+                    .identity_provider
+                    .get_id()
+                    .map_err(|_| SessionError::SessionBuilderIncomplete)?;
+
+                let data_input = format!("{}/{}/data", token_id, destination);
+                let mut data_id = twox_hash::XxHash3_128::oneshot(data_input.as_bytes());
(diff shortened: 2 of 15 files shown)


crates/session/src/session_layer.rs
```diff
@@ -248,7 +248,40 @@ where
...
+            if session_config.session_type == ProtoSessionType::Multicast {
+                let token_id = self
+                    .identity_provider
+                    .get_id()
+                    .map_err(|_| SessionError::SessionBuilderIncomplete)?;
+
+                let data_input = format!("{}/{}/data", token_id, destination);
+                let mut data_id = twox_hash::XxHash3_128::oneshot(data_input.as_bytes());
+                if NameId::is_reserved_id(data_id) {
+                    data_id %= u128::MAX - NameId::RESERVED_IDS;
```
msardara: `is_reserved_id` covers `id >= u128::MAX - 50`, so the modulus is `u128::MAX - 50`. If the hash lands on exactly that boundary:
```
(u128::MAX - 50) % (u128::MAX - 50) == 0
```
The session builder only rejects `id == NULL_COMPONENT` (`u128::MAX`), not `0`. Two different `(token, channel)` pairs that both reduce to `0` will silently share a subscription-table entry.

Suggested fix: after the modulo, also exclude `0`, or use `(id % (u128::MAX - RESERVED_IDS - 1)) + 1` to keep the result in `[1, u128::MAX - RESERVED_IDS - 1]`.

crates/session/src/session_layer.rs
```diff
@@ -248,7 +248,40 @@ where
...
+                }
+
+                let ctrl_input = format!("{}/{}/control", token_id, destination);
+                let mut ctrl_id = twox_hash::XxHash3_128::oneshot(ctrl_input.as_bytes());
+                if NameId::is_reserved_id(ctrl_id) {
+                    ctrl_id %= u128::MAX - NameId::RESERVED_IDS;
```
msardara: If both `data_id` and `ctrl_id` reduce to the same value after the modulo, `SessionBuilder::build()` returns `SessionBuilderIncomplete` permanently for that `(identity, channel)` pair.

The retry loop in `create_session_internal` only handles session-ID pool collisions, not hash collisions — there is no recovery path. The probability is negligible, but when it does happen the error is completely opaque (same variant as a genuinely misconfigured builder). Consider logging the collision and salting the hash input (e.g. appending a counter) to retry.

crates/session/src/session_layer.rs
```diff
@@ -248,7 +248,40 @@ where
         let is_p2p = session_config.session_type == ProtoSessionType::PointToPoint;
         let destination_proto = destination.clone();
 
-        let session = self.create_session_internal(session_config, local_name, destination, id)?;
+        let (dest_with_id, control_with_id) =
+            if session_config.session_type == ProtoSessionType::Multicast {
+                let token_id = self
+                    .identity_provider
+                    .get_id()
+                    .map_err(|_| SessionError::SessionBuilderIncomplete)?;
```
msardara: `SessionError::SessionBuilderIncomplete` means "not all required builder fields were set", which is misleading when the actual cause is `get_id()` failing. `SessionError` already has `Auth(#[from] AuthError)`, so bare `?` would propagate the real error and let callers distinguish an auth failure from a misconfigured session:

```rust
let token_id = self.identity_provider.get_id()?;
```

crates/session/src/session_layer.rs
```diff
@@ -248,7 +248,40 @@ where
...
+
+                let data_input = format!("{}/{}/data", token_id, destination);
```
msardara: `ProtoName::Display` emits `org/ns/name/{id}` — the ID is included. All current callers pass bare names (ID = `NULL_COMPONENT`), so today this is deterministic. But there is no guard preventing a caller from passing a `destination` with a non-null ID already set, which would silently produce different channel IDs than any participant reconstructing the name from the wire.

Consider hashing only the three string components to make the derivation ID-agnostic by construction.

crates/service/src/app.rs
```diff
@@ -1224,10 +1224,10 @@ mod tests {
             // Verify it's a multicast session
             assert_eq!(session_arc.session_type(), ProtoSessionType::Multicast);
 
-            // For multicast sessions, the destination is the channel name with DATA_CHANNEL_ID
+            // For multicast sessions, the destination is the channel name
             let dst = session_arc.dst();
-            let expected_dst = channel_name.clone().with_id(NameId::DATA_CHANNEL_ID);
-            assert_eq!(dst, &expected_dst);
+            let expected_dst = channel_name.clone();
+            assert!(dst.match_prefix(&expected_dst));
```
msardara, line 1230: The old assertion verified the exact data-channel ID:
```rust
let expected_dst = channel_name.clone().with_id(NameId::DATA_CHANNEL_ID);
assert_eq!(dst, &expected_dst);
```
The new assertion only checks the three string components and would pass even if `dst` carries `NULL_COMPONENT` as its ID (i.e. the hash derivation was skipped or produced an unset name). Consider additionally asserting `dst.id() != NameId::NULL_COMPONENT` to catch that regression.

msardara (commented):
LGTM, with some nits

crates/session/src/session_layer.rs
micpapal: A collision here should be impossible, but it is easy to add a retry loop

msardara (approved):

PR #1822, opened 2026-07-10
fix: remove unused pub modifier from test-gated functions

### LINKED ISSUE #1821: fix: remove unused pub modifier
# Description
`is_endpoint` and `parse_endpoint` are both declared `pub fn` but are gated with `#[cfg(test)]`, meaning they only exist in test builds. The pub modifier is a no-op here — test-gated items are never part of the crate's public API — but it misleads readers into thinking these are intentionally exported helpers rather than test-only utilities.

# Proposed Solution
Remove the pub modifier from both functions. No other file references them, so this is a visibility-only change with no functional impact.

ahyeon2626:
Removes the unused pub modifier from is_endpoint and parse_endpoint in crates/slimctl/src/utils.rs. Both functions are gated with #[cfg(test)] and are not referenced outside this file, so this is a visibility-only change with no functional impact. Closes #1821

### DIFF
--- a/crates/slimctl/src/utils.rs
+++ b/crates/slimctl/src/utils.rs
@@ -56,14 +56,14 @@ pub fn parse_route(route: &str) -> Result<(String, String, String, u128)> {
 
 /// Return true if the string looks like an endpoint URL (contains ':' or starts with http/https).
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


markpmarton (approved):
LGTM! Thanks for the contribution.

Issue #1807, opened 2026-07-07
feat(channel): offline membership state and session persistence — core crates


Tehsmash:
## Context

Closes part of #1786.

Channel membership is currently tied to an open session — closing a session removes the participant from the channel. This makes membership ephemeral and prevents agents from rejoining a channel they were already a member of without going through the full invite flow again. This issue solves the problem at the core Rust crate level.

### Offline membership state

- Closing a session transitions the participant to an **offline** state rather than removing them from the channel.
- Reconnecting (opening a new session for the same topic) transitions the participant back to **online**.
- Add an explicit **leave channel** operation, separate from closing a session.

### Session state persistence

- Define exactly what must be persisted for an agent to rejoin a channel: identity credentials and MLS group state (epoch, key material).
- Implement `session_from_state(state)` in the core session crates to restore a session from persisted state.
- Define the storage format (encrypted file, SQLite, etc.) and default location.

### Moderator re-key on epoch mismatch

- If an agent rejoins with an out-of-sync MLS epoch the channel moderator (channel-manager) detects the mismatch and triggers a group re-key so all members share a fresh epoch before the rejoining agent proceeds.

## Acceptance Criteria

- [ ] Participant offline/online state model implemented in core crates
- [ ] Session close no longer removes participant from channel membership
- [ ] Explicit leave-channel operation added
- [ ] `session_from_state(state)` API implemented
- [ ] Persistent state format and default storage location defined and documented
- [ ] Moderator re-key flow implemented for epoch-mismatch on rejoin
- [ ] Existing session and channel tests updated; new tests for offline/rejoin/leave flows

micpapal:
## Current Behavior

### Session Close

#### Initiated by Moderator

1. The moderator sends a **group close** message on the channel.
2. All participants send an ACK and close their sessions.
3. The moderator collects all ACKs and closes its own session.

```mermaid
sequenceDiagram
    participant M as Moderator
    participant P1 as Participant 1
    participant P2 as Participant 2

    M->>P1: Group Close
    M->>P2: Group Close
    P1->>M: ACK
    P2->>M: ACK
    Note over P1,P2: Participants close sessions
    Note over M: Moderator closes session
```

#### Initiated by a Participant

1. The participant sends a **leave request** to the moderator.
2. The moderator acknowledges the request.
3. The participant closes its session.
4. The moderator proceeds based on session type:

**P2P Session:**
- The moderator closes the session.

**Group Session:**
[...]

Issue #1817, opened 2026-07-08
feat: Update ping disconnection to use offline state


micpapal:
Change the ping behavior: instead of closing the session when a participant or moderator is unresponsive, move the unresponsive node to offline state. This allows recovery via  session_resume  after crashes.

See full design: #1807

Tehsmash:
Discussion on a heartbeat option is here: https://github.com/agntcy/slim/pull/1823#discussion_r3570199716

Main downside to members heartbeating is the extra traffic it could generate
Benefit is that all members can self-detect when other members go offline/crash etc. even if the moderator goes offline.

I think we could probably optimise this in a group setting for example there is no need to heartbeat if you've sent a broadcast message to the group within the heartbeat window, the listening agent can take the fact you've sent a message as a liveness indicator.

micpapal:
Related to #1807 and PR #1823 (close/rejoin session).

The current ping mechanism is centralized: the moderator (or initiator) sends periodic ping messages to all participants, and participants reply. If a participant doesn't reply after N consecutive pings, it is considered disconnected and removed from the group. Conversely, participants detect moderator disconnection if no ping is received within the timeout.

This design has a gap: **if the moderator goes offline, no one sends pings, and further participant disconnections go undetected** until a new moderator is established.

## Proposal: Decentralized Heartbeats

Replace the moderator-sent ping/reply mechanism with **one-way heartbeat broadcasts from every participant**. Each participant periodically sends a heartbeat message to the group and independently tracks when it last heard from every other participant. If no heartbeat is received from a participant within a configurable timeout, that participant is marked **offline** locally (not removed from the group).

### Key design decisions

1. **One-way messages, no replies** — heartbeats are fire-and-forget broadcasts. No ACK/reply is needed, simplifying the protocol.
2. **Fully decentralized** — every participant (including the moderator) sends heartbeats and monitors others. No single point of failure.
3. **Offline, not disconnected** — when a heartbeat timeout is detected, the participant is moved to `OFF_LINE` state (from PR #1823) rather than being removed from the group. This aligns with the persistent session model where participants can rejoin.

micpapal:
## Heartbeat-Based Disconnection Detection with Epoch (see PR https://github.com/agntcy/slim/pull/1868)

### Overview

[...]

PR #1823, opened 2026-07-10
feat: close and rejoin session functions

### LINKED ISSUE #1815: feat: Implement  session_pause  and  session_resume
Add new proto messages ( UpdateParticipantStatePayload ,  RejoinRequestPayload ,  RejoinReplyPayload ) and implement  session_pause  and  session_resume.

See full design: #1807

micpapal:
This PR implements the ability for participants (or moderator) in a group session to go offline and rejoin later, making sessions persistent rather than ephemeral.

The public API provides:
- **`session.close()`** — notifies the group that this participant is going offline. The session state can be persisted and the application closed. The participant remains a member of the group.
- **`session.rejoin()`** — reconnects using stored state, validates with the moderator, and re-enters the group as online.
- **`session.leave()`** — permanently exits the group (cancels the session). This is the old `close()` behavior.

### Protocol Changes

**New protobuf types:**
- `ParticipantState` enum (`ON_LINE`, `OFF_LINE`)
- `UpdateParticipantStatePayload` — carries the participant name and new state
- `RejoinRequestPayload` — carries participant name, session_id, and MLS epoch for validation
- `RejoinReplyPayload` — success/failure response from the moderator

**New message types:**
- `SESSION_MESSAGE_TYPE_UPDATE_PARTICIPANT_STATE` (19)
- `SESSION_MESSAGE_TYPE_REJOIN_REQUEST` (20)
- `SESSION_MESSAGE_TYPE_REJOIN_REPLY` (21)

#### Close (Participant → Offline)

1. Participant sends `UpdateParticipantState(OFF_LINE)` broadcast on the channel
2. All participants update the sender's state to offline
3. Participant collects ACKs from all group members
4. Participant stores session state and can close the app

#### Rejoin (Participant → Online)

1. Participant sends `RejoinRequest` to the moderator with session_id and MLS epoch
2. Moderator validates: participant is known, was offline, session_id matches, MLS epoch matches
3. On success: moderator sends `RejoinReply(true)` and broadcasts `UpdateParticipantState(ON_LINE)` on behalf of the participant
4. On failure: moderator sends `RejoinReply(false)` — participant must rejoin via a fresh invite

### Related Issue

close: #1815
related to: #1807

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


crates/session/src/session_controller.rs
Tehsmash, line 653: I had envisioned that session.close() would be the primary interface for "pausing" so then an A2A client can do something like:

```
channel_manager = request_invite(name="...")

#  Join a session the first time
session = app.listen_for_session(...) 
session.publish(...)
session.close() 

# Rejoin a session
session = app.rejoin_session(name="...", session_id="...") <-- not the real UX for this but something similar 
session.publish(..)
session.close()

# Rejoin again then quit the channel offically
session = app.rejoin_session(name="...", session_id="...")
session.leave() <-- quit the channel for real
```

The pause / resume is entirely handled under the hood as part of the creation of the session / closing of the session. I guess close and whatever the rejoin function ends up being can use these functions.

As a side note: we should also probably drain as part of pausing so remaining pending messages are sent and then the offline notification goes out.

crates/session/src/session_moderator.rs
```diff
@@ -403,7 +431,33 @@ where
...
+            for name in missing {
+                if let Err(e) = self
+                    .common
+                    .settings
+                    .tx_session
+                    .try_send(SessionMessage::ParticipantDisconnected { name: Some(name) })
+                {
+                    debug!(error = %e, "failed to send participant disconnected message");
+                }
+            }
```
Tehsmash: How does this handle the scenario:
moderator goes offline -> participant a goes offline so misses offline notification -> moderator comes back online?

In that case the participant shouldn't be disconnected from the channel right? The moderator should just notice that the participant is not responding and move it to offline.

This also makes me wonder about how we handle crashes / unexpected disconnects etc.

Today we have the ping/echo mechanism so I was thinking something like:

Happy path:
Participant closes session -> broadcasts to everyone UpdateState(offline) -> everyone updates their state (no need for the moderator to even be up)

Crash Path:
Participant crashes -> participant stops responding to echos from moderator -> moderator broadcasts that Participant has gone offline
(this requires the moderator to be online)

If we want the crash path to work without the moderator then either:
1) all participants can broadcast heartbeats if online and if any participant sees a missed heartbeat from another participant then they mark the participant offline locally, if heartbeats resume then the participant is moved back online. Downside is the traffic scales with the number of participants. 

2) The next time any participant tries to send a broadcast message on the channel if another participant fails to respond / ack then they can broadcast to the group that participant is offline. When the participant comes back up it should go through the rejoin process anyway and the moderator will broadcast back online.

After writing this all down, it seems to me like 1) makes the most sense as a solution which can handle both going offline and coming back online without the moderator or substitute pseudo-moderator needing to be involved as long as the MLS state doesn't need re-keying. WDYT?

Actually expanding on this:

If the heartbeat was to include the MLS epoc then a participant could just start heartbeating again when they come online, if the moderator is online then it can notice the mismatch and send the rekey to the group.

A participant can also see the heartbeats of the other participants and if the epoc doesn't match then it can do two things:
If self.epoc less than incoming_heartbeat.epoc then I know I need to wait for a rekey.
If self.epoc is greater than incomming_heartbeat.spec then mark the participant offline locally same as a missed heartbeat.

Rekey brings everyone back up to the same epoc. Then everyone will see the heartbeats with matching epocs and everyone will be back online.

crates/session/src/session_controller.rs
micpapal: What I wanted here was a distinction between the two possible options: 
- leave a channel for some time and rejoin later
- leave the channel forever
What you propose works as well and it is kind of transparent for the user unless he really wants to leave the channel. Then the resume/rejoin will also be handle inside the restore_session_from_file or how we will call that function so the whole cycle  will not be visible to the user. 

And I agree on the drain. I will add that

crates/session/src/session_moderator.rs
micpapal: for the disconnection detection on this PR I can change it and declare the participant as offline instead of remove it from the group.

For the ping part I will modify that in another PR to make things a bit more simple to review. I have the issue here https://github.com/agntcy/slim/issues/1817 however I want to modify the current ping behavior. If someone fails to reply to the moderator we need to move the participant offline instead of disconnect it, which match exactly what you are describing.

For the ping when the moderator is offline I am not yet totally sure what is the best option. Send an heartbeat from every participant seems simpler, but It may generate a lot of useless traffic, especially in big groups. Now we are using 2N messages at every ping, using heartbeat we need 2NxN messages as each participant needs to send out N messages, that can be reduced to NxN as most likely we don't need the ack back. This seems a lot. I will try to put a proposal in the related issue (https://github.com/agntcy/slim/issues/1817) and we can discuss abot this there.

PR #1825, opened 2026-07-10
feat(mls): share one signing identity across an app and its sessions

### LINKED ISSUE #1824: MLS signing identity is per-session, not per-app — make it app-scoped
## Summary

The MLS signature keypair is effectively **per-session** rather than **per-app**. An app that runs multiple MLS sessions ends up with multiple,  independently-generated signing identities, and no signing identity is stable across a process restart. This blocks a coherent "one identity per app" model and is a prerequisite for persisting/restoring an app's sessions.

## Current behavior

- `SharedSecret` stores its MLS signing material `(secret, public)` in a **by-value field**, so every clone owns an *independent* copy. The `SessionLayer` clones the app's identity provider into each session, so each session starts from its own copy.
- `Mls::initialize` calls `generate_and_install_signing_identity(false)` on **every** init, generating a fresh keypair via the MLS crypto provider and overwriting that session's provider clone.

Net effect: N sessions ⇒ up to N different signing identities for one app. Nothing survives a restart consistently.

## Proposed change

Make the signing identity app-scoped, generated once via mls-rs and shared:

- Hold `SharedSecret`'s signing material behind a shared `Arc<RwLock<_>>`, so an app and every session cloned from it present **one** identity; a rotation/restore is visible to all clones at once.
- Add `TokenProvider::mls_signature_keys_installed()` (default `false`) to distinguish construction-time placeholder keys from MLS-installed, ciphersuite-correct keys.
- `Mls::initialize` generates a fresh keypair via its own crypto provider **only on first use**, then reuses the installed one for all subsequent sessions.
- Propagate the new trait method to the other `TokenProvider` impls: `AuthProvider` (delegates to the inner provider — load-bearing, since most callers use the enum), `SignerJwt` and `SpireIdentityManager` (installed ⇔ keys non-empty). `OidcTokenProvider`/`StaticTokenProvider` don't support MLS, so the default `false` is correct.

## Scope / non-goals

- **In scope:** the provider + MLS foundation above (branch `feat/shared-mls-signing-identity`, off `main`).
- **Out of scope (follow-up):** persisting the app identity and re-installing it on restart. That needs the encrypted `kv_store` that only exis
ts on the session-persistence branch, and it surfaces a separate open question — the persistent store is keyed by `identity_provider.get_id()`,
 which currently includes a random suffix, so restore needs a **stable app id** before it can locate the store across restarts.

## Acceptance criteria

- [ ] An app and all sessions cloned from it share a single MLS signing identity.
- [ ] The keypair is generated once (via the MLS crypto provider) and reused across sessions; rotation still works

msardara:
# Description

The MLS signature keypair was effectively per-session: `SharedSecret`
held it in a by-value field, so every clone owned an independent copy,
and each `Mls` init regenerated and installed a fresh pair. An app with
multiple MLS sessions therefore had multiple, diverging signing
identities, and nothing could survive a restart consistently.

Fixes: #1824 

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


micpapal (approved):
lgtm

Tehsmash (approved):
Great catch! This would have broken my identity cache+reload logic in slim-bindings too!

PR #1828, opened 2026-07-12
fix(config): require tokio-retry >=0.3.2 for RetryIf::start

muscariello:
## Summary

Raise the `tokio-retry` version floor from `"0.3"` to `"0.3.2"` so the
**published** `agntcy-slim-config` crate is buildable by external consumers.

```diff
-tokio-retry = "0.3"
+tokio-retry = "0.3.2"
```

## Why

`ClientConfig::retry_connect` (`crates/config/src/client.rs`) uses
`tokio_retry::RetryIf::start`, which was introduced in **tokio-retry 0.3.2** —
that release deprecated the old `spawn` constructor and renamed it to `start`.

The workspace declared the loose requirement `tokio-retry = "0.3"`. In this repo
the lockfile pins `0.3.2`, so it builds fine (and `main` is green). **But
downstream Rust consumers of the published crate can resolve an older 0.3.x
(`0.3.0` / `0.3.1`) that has no `start`**, so `agntcy-slim-config` fails to
compile from crates.io — which breaks the entire v2 dependency chain:

```
slim-config → slim-datapath → slim-controller → slim-service
            → slim → slim-bindings → a2a-slimrpc
```

i.e. nothing currently builds on the `2.0.0-alpha` line from crates.io.

### Reproduction

From a separate workspace whose lockfile pins `tokio-retry 0.3.0`, building
anything that pulls `agntcy-slim-config` fails:

```
error[E0599]: no function or associated item named `start` found for struct
`RetryIf<I, A, C>` ...
   --> agntcy-slim-config-0.12.0/src/client.rs
```

## Fix

Pin the floor to the version that actually provides `RetryIf::start`. No code
change — `client.rs` is unchanged; `start` is already correct for tokio-retry
≥0.3.2.

## Follow-up

This unblocks cutting a **buildable stable `2.0.0`** from crates.io. Recommend a
CI job that builds the *published* crates from a clean external workspace (not
just the monorepo, whose lockfile masks loose version floors) to catch this
class of "green in-tree, broken when published" regression.


muscariello:
Added a second commit  — 
**test(control-plane): de-flake auth-link integration tests**. 

The `coverage` / `unittest-matrix` reds were a pre-existing repo-wide flake in 
`control-plane/tests/integration_test.rs` (a different `test_link_*` failed each run, at the same two assertion sites, on `main` too), independent of the tokio-retry floor change. 

Two fixes: 
(1) `auth_link()` now selects the connector-sourced link directly, so a transient protected-sourced link during reconciliation no longer trips `wait_auth_link`'s source assertion; 

(2) doubled the wait timeouts (30→60s / 15→30s) for headroom under llvm-cov/parallel instrumentation (happy path returns immediately). 

Verified locally with `cargo check` + `cargo clippy --tests -D warnings`.

PR #1842, opened 2026-07-14
ci: make wasm tests standalone nightly jobs, not skipped on PRs

muscariello:
Follow-up to #1841. Disabling the wasm suites via reusable-workflow toggles left a conditional job that still renders as **skipped** in the PR checks. This removes the wasm jobs and their toggle inputs from `reusable-rust-build-and-test` and defines `wasm-test`/`wasm-browser-test` directly in `ci-nightly` (like the integration and sessions suites). PRs no longer show a wasm job at all.

actionlint clean; no dangling `rust-test-wasm` references.


muscariello:
Closing: this extraction breaks the reusable workflow's modularity. wasm is already a modular job there, toggled by `rust-test-wasm` exactly like lint/test/cross-build (merged in #1841). A toggled-off job rendering as "skipped" is uniform across every job in that workflow (nightly skips unittest-matrix the same way), so it's not worth de-modularizing wasm to hide it.

PR #1847, opened 2026-07-15
feat(rpc-compiler): add slimrpc protoc plugin crate

muscariello:
Brings the slimrpc code generator (`agntcy-protoc-slimrpc-plugin`) from `slim-bindings` into the workspace, so slimrpc client stubs + server handlers can be generated in-repo for Python, Go, Java, C#, Kotlin, and TypeScript/Node.

- New `crates/rpc-compiler` workspace member; deps are workspace `prost`/`prost-types`/`heck`/`anyhow` only (no coupling to the FFI crate or the runtime crate).
- Registered in `.release-plz.toml`.
- Builds, `clippy -D warnings`, and `fmt --check` all clean in-workspace.

Companion to the slimrpc runtime crate (#1830). Independent of it — no dependency, so it can land on its own.


muscariello:
Closing: leaving the slimrpc compiler out of the slim repo for now (per discussion). The dedup (#1850) is independent of it and does not need it. Can revisit bringing the compiler in later.

PR #1864, opened 2026-07-16
refactor: remove slimrpc from bindings

msardara:
# Description

Completely remove slimrpc from slim bindings.

Slimrpc has now its own crates with a feature enabling/disabling uniffi exports.

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


crates/rpc/src/channel.rs
```diff
@@ -914,7 +916,52 @@ impl Channel {
...
+        connection_id: Option<u64>,
+    ) -> Result<Self, RpcError> {
+        let slim_app = app.inner_app().clone();
+        let slim_names = members.iter().map(|n| n.as_slim_name().clone()).collect();
+        Self::new_with_members_internal(slim_app, slim_names, true, connection_id)
+    }
+}
 
+#[cfg_attr(feature = "uniffi", uniffi::export)]
+impl Channel {
```
Tehsmash, line 958: Its a shame we need to duplicate the constructors for uniffi, if we can do what we said in the meeting about uniffi exporting "App" directly from datapath then it would be significantly cleaner. As this a breaking public API change if we can do it for v2.0 then it would be best.

msardara: +1, I'll try to handle it in a follow-up PR 👌

Tehsmash (approved):

Issue #1865, opened 2026-07-16
Strict require_header_mac: messages dropped when routed to a connection whose link session isn't negotiated yet


muscariello:
### Symptom

With `require_header_mac = true` (strict), establishing a session / adding a participant **intermittently** fails with `message send retries exhausted`. With `false` it is reliable. Repro: agntcy/shadi#91.

### Root cause

Strict mode **fatally drops a message that reaches a remote connection whose link-HMAC session (or `link_id`) isn't installed yet** — a transient window during link negotiation. Non-strict tolerates it; strict returns `HeaderMacAwaitingLinkNegotiation` / "strict header MAC required but link HMAC session is not installed" (crates/datapath/src/message_processing.rs, `verify_remote_header_mac` ~L351/368 and `send_msg_raw` ~L669). The message is dropped, no ACK comes back, and the session layer retries (`max_retries`) and gives up → the error above. The intermittency is the timing of when a connection is used to route vs. when its negotiation completes.

Important: strict-dropping a message that has **no** link session is **deliberate** — see the existing test `verify_remote_header_mac_strict_rejects_publish_without_mac_session`. Fully-negotiated links sign and verify correctly (client always sends an ECDH key; server always installs the session + `link_id`), so steady state is fine. So the defect is **not** "strict is too strict" — it is that **a message is routed to a connection before that connection's link negotiation has completed.**

### Open question / fix direction

Should a remote connection be routable before its link negotiation completes?
- **If no:** gate routing/forwarding on negotiation-complete, so a message can never reach a not-yet-negotiated connection.
- **If yes (tolerate the window):** hold/queue messages for a connection until its session is installed (or the negotiation timeout fires) instead of dropping.

Either way this needs a **deterministic reproduction** to confirm the exact ordering before patching — it is timing-dependent and does not reproduce reliably.

### Related

#1866 makes `run_server` honor `ServerConfig.require_header_mac` (needed for *consistent* config) — separate from this and not a fix for it.

msardara:
## Bug

#1740 put MLS-key-based signing/verification + replay protection onto the control-message path, which is exactly the session add-participant (JOIN) handshake. Under concurrent session-establishment load it fails two ways (both captured in trace logs):

[...]

PR #1866, opened 2026-07-16
fix(datapath): honor ServerConfig.require_header_mac in run_server

muscariello:
Refs #1865 (partial — the config-honoring piece; not the full fix).

### Problem

The node `MessageProcessor` fixes `require_header_mac` at construction. On the dynamic path (`Service::new` then a later `run_server`, as slim-bindings uses), the processor defaults to `false`, so a `ServerConfig` requesting a different value is silently ignored for inbound connections — `run_server` only *warned* about the mismatch.

### Fix

Make `server_require_header_mac` an `AtomicBool` and have `MessageProcessor::run_server` store `config.require_header_mac` before serving, so inbound connections honor the `ServerConfig` regardless of how the processor was constructed. Replaces the mismatch warning with the correct behavior.

### Scope

This makes configuration consistent, which is a prerequisite for enabling strict header MAC — but it does **not** resolve the root cause in #1865 (subscription/`subscribe`/`unsubscribe` control messages are not consistently header-MAC-signed, so strict mode still breaks session/route setup). That needs a separate control-plane-integrity fix. Deliberately using `Refs` (not `Fixes`) so #1865 stays open.

### Test

`run_server_applies_require_header_mac_from_config`: a processor built via `new_with_service_id` (default false) adopts `require_header_mac=true` after `run_server`. Full datapath suite green (158 tests); `agntcy-slim-service` + `agntcy-slim-bindings` build clean; fmt + clippy clean.

### DIFF
--- a/crates/datapath/src/message_processing.rs
+++ b/crates/datapath/src/message_processing.rs
@@ -4,6 +4,7 @@
 use std::collections::{HashMap, HashSet};
 use std::net::SocketAddr;
 use std::sync::Arc;
+use std::sync::atomic::{AtomicBool, Ordering};
 
 use display_error_chain::ErrorChainExt;
 use parking_lot::RwLock;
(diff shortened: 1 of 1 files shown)


muscariello:
Closing. This was motivated by the flaky handshake, which is actually fixed by #1869 (MLS signing-key layer), unrelated to header MAC. The `run_server` config-honoring behavior here is a minor standalone correctness point and not worth pursuing on its own right now.

PR #1869, opened 2026-07-16
fix(auth,mls): stop mid-handshake MLS signing-key rotation

msardara:
# Description

Concurrent session establishment (the add-participant / JOIN handshake) failed intermittently with auth error: MLS signature key generation failed, stalling the handshake until reliable-send retries were exhausted (~11s) and the RPC returned Unavailable. This surfaced as flaky slim-rpc integration tests (~16% failure per run).

Session control messages are signed and verified with the identity's MLS signature key. Two latent issues combined:

1. Ciphersuite mismatch → mid-handshake key rotation. SharedSecret (and SignerJwt, SpireIdentityManager) generated placeholder keys of the wrong ciphersuite — Curve25519 while MLS defaults to P-256 — or started empty. So Mls::initialize rotated the shared signing identity via set_signature_keys during the handshake. Racing a concurrent control-message signature on the same cloned identity produced mismatched key material and a signing failure. Because the reply was never produced, retransmits (reusing the same message_id) timed out.
2. AWS-LC strips leading zeros from P-256 secret scalars (~1/256 keys are 31 bytes). p256_signing_key only accepted exactly-32-byte raw scalars, so a short scalar fell through to the PKCS#8 branch and failed — a latent bug the ciphersuite switch would otherwise have exposed.

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


micpapal (approved):
lgtm

PR #1875, opened 2026-07-17
feat(session): add mls re-key on rejon

### LINKED ISSUE #1816: feat: Implement group rekey
Add  GroupUpdatePayload  message type. Implement a  rekey_group()  method on the moderator that creates a rotation proposal, commits it, and broadcasts the MLS commit to the group. This is triggered during  session_resume  when the rejoining participant's MLS epoch does not match the current one.

See full design: #1807

micpapal:
# Description

This PR extends the close/rejoin feature (from `feat/close-rejoin`) with **MLS re-key on rejoin**, ensuring that participants who go offline and come back can securely re-establish their MLS key material when their epoch is stale.

### New Protocol Messages & Payloads
- **`GroupUpdate`** — replaces the separate `GroupAdd` / `GroupRemove` messages with a unified message carrying a `GroupUpdateOp` (`ADD`, `REMOVE`, `REJOIN`) and the full participant list.
- **`RejoinRequest`** — sent by a participant to the moderator when it receives a NACK on rejoin due to epoch mismatch. Contains a fresh MLS key package.
- **`RejoinReply`** — sent by the moderator back to the rejoining participant, containing the MLS welcome/commit produced by the re-key operation.
- `Participant` proto now includes a `status` field (`ParticipantState`) directly, replacing the separate online/offline wrapper.

### MLS Re-key (`crates/mls`)
- Added `Mls::rejoin_member()` — performs a single MLS commit that removes and re-adds a member with a fresh key package, producing both a commit (for existing members) and a welcome (for the rejoining member).
- Added `MlsState::rejoin_participant()` wrapper that updates local participant identity after re-key.
- Commit processing now uses the unified `GroupUpdate` payload.

### Moderator Changes
- Group roster now stores `Participant` directly with inline `status` field.
- Sending `GroupAdd`/`GroupRemove` replaced by unified `send_group_update(op, participant, participants, mls)`.
- New `on_rejoin_request()` flow: validates participant, starts `ModeratorTask::Rejoin`, calls MLS rejoin, sends `RejoinReply` to the participant, and broadcasts `GroupUpdate(REJOIN)` to the group.
- New `ModeratorTask::Rejoin` task type with welcome/commit phases.

### Participant Changes
- `GroupUpdate` handler replaces separate add/remove handlers, now also handles `REJOIN` operations (self-rejoin refreshes entire group list, other-rejoin updates the specific member).
- `RejoinReply` handler processes the MLS welcome, updates epoch, resolves pending rejoin ack, and sends `GroupAck`.
- On `GroupNack` for rejoin: instead of failing, generates a fresh key package and sends `RejoinRequest` to the moderator to trigger re-key.

### Overall Rejoin + MLS Re-key Flow
1. Participant sends `UpdateParticipantState(ONLINE)` to rejoin.
2. Other participants detect epoch mismatch and reply with `GroupNack`.
3. Participant generates a fresh MLS key package and sends `RejoinRequest` to the moderator.
4. Moderator performs MLS re-key (remove + re-add in one commit), sends `RejoinReply` (with welcome) to the participant and broadcasts `GroupUpdate(REJOIN)` to the group.
5. Participant processes the welcome, resets MLS state, and resumes messaging with synchronized epoch and refreshed key material.

close: https://github.com/agntcy/slim/issues/1816

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


crates/session/src/session_moderator.rs
```diff
@@ -1452,6 +1454,147 @@ where
...
+        debug!(
+            from = %source,
+            id = %msg.get_id(),
+            "received rejoin request",
+        );
+
+        // Extract the rejoin request payload
+        let rejoin_payload = msg
+            .get_payload()
+            .unwrap()
```
msardara: Panic on malformed `RejoinRequest`. `msg.get_payload().unwrap()` will panic the moderator task if a `RejoinRequest` arrives without a payload (malformed or malicious peer), taking down the whole group's control plane. Every other control handler uses `.ok_or(SessionError::MissingPayload { .. })` — please do the same here.

crates/session/src/session_moderator.rs
```diff
@@ -1452,6 +1454,147 @@ where
...
+
+        if participant_entry.status != ParticipantState::Offline as i32 {
+            debug!(from = %participant_name, "rejoin request from online participant, ignoring");
+            return Ok(SessionOutput::new());
+        }
+
+        // 2. Check if there is a current task
+        if self.current_task.is_some() {
+            // 3. Drop the message — the participant will retry
+            debug!(from = %participant_name, "moderator is busy, dropping rejoin request (participant will retry)");
```
msardara, line 1500: The comment assumes the participant will retry, but the participant sends `RejoinRequest` fire-and-forget (`create_control_message` + `to_slim`, bypassing `sender.on_message`), so there is no retransmission timer and no re-trigger — the heartbeat path is gated on `pending_status_update.is_none()`.

If the request is dropped here (moderator busy) or the request/reply is lost, the participant is stuck offline permanently. Either register the request for retransmission on the participant side, or queue the rejoin here instead of dropping it.

crates/session/src/session_moderator.rs
```diff
@@ -1452,6 +1454,147 @@ where
...
+
+        // 7. Start the welcome phase in rejoin task
+        self.current_task
+            .as_mut()
+            .unwrap()
+            .welcome_start(reply_msg_id)?;
+
+        // Mark participant as online in group list
+        if let Some(entry) = self.group_list.get_mut(&name_no_id) {
+            entry.status = ParticipantState::Online as i32;
```
msardara, line 1555: Endpoint never restored for the rejoining participant.

This marks the roster entry `Online` but never calls `add_endpoint` (`sender.add_participant` + `inner.add_endpoint`). When the participant went offline the moderator called `remove_endpoint`, and every other bring-online path re-adds it — `on_update_participant_state` Online (~L1401) and heartbeat-back-online (~L1846) both call `add_endpoint`. `on_group_ack` for the `Rejoin` task doesn't add it either.

Result: after a successful re-key the moderator has no datapath route to the participant and doesn't track it in `sender.group_list`, so it can't deliver application data, won't monitor heartbeats, and won't expect its ACKs. Because the roster now reads `Online`, the heartbeat-back-online branch is skipped too, so it never self-heals. Add `self.add_endpoint(&participant_entry).await?` (and the route) on the rejoin path.

crates/session/src/session_participant.rs
```diff
@@ -233,11 +228,23 @@ where
...
+                            // treat as partial success: non-responders are offline, notify app OK
+                            if pending_task.message_type
+                                == ProtoSessionMessageType::UpdateParticipantState
+                            {
+                                if let Some(tx) = pending_task.ack_tx {
+                                    let _ = tx.send(Ok(()));
+                                }
+                                self.common.online = true;
+                                self.common.sender.restart_heartbeat();
+                            } else {
```
msardara, line 244: This `RejoinFailed` fallback is unreachable. The outer guard at L225 requires `message_type == UpdateParticipantState`, but the rejoin's pending task has `message_type == RejoinRequest`, and — more fundamentally — the `RejoinRequest` is sent fire-and-forget (see `on_group_nack`/`on_heartbeat`), so no timer is ever registered for it and no `TimerFailure` fires. Consequently the app's `resume()` `CompletionHandle` (`ack_tx`) never resolves on failure. The rejoin needs a real timeout/retry mechanism for this branch to be meaningful.

crates/session/src/session_participant.rs
```diff
@@ -1010,26 +1190,69 @@ where
...
+                };
+
+                // send a rejoin request to the moderator
+                let moderator = self
+                    .moderator_name
+                    .clone()
+                    .ok_or(SessionError::ModeratorNotFound)?;
+                let msg_id = rand::random::<u32>();
+                let participant = self.common.settings.source.clone();
+                let rejoin_msg = self.common.create_control_message(
```
msardara, line 1222: Fire-and-forget send. The `RejoinRequest` is built with `create_control_message` and returned via `SessionOutput::to_slim` without going through `self.common.sender.on_message`, so it's never registered for retransmission or ACK tracking. Combined with the moderator dropping the request when busy and the unreachable failure branch above, any loss or contention leaves the participant permanently offline. Consider routing it through the sender (like `UpdateParticipantState`) so it retransmits and can time out.

msardara (approved):

PR #1883, opened 2026-07-20
fix(auth): verify JWT against every JWKS candidate key when no `kid` is present

### LINKED ISSUE #1882: JWT auth: kid-less tokens can't be verified against a multi-key JWKS (first-key selection) — breaks multi-member group sessions
## Summary

When JWT identities are used with a **static, multi-key JWKS allow-list** (several members, each an `EdDSA`/`did:key` key), a member can only be verified if its key happens to be the **first** key in the JWKS. Every other member's token fails with `InvalidSignature`, because slim mints tokens without a `kid` and the verifier does not disambiguate kid-less tokens across multiple keys.

This breaks any group session with ≥2 members using per-identity JWT credentials: the moderator's `DiscoveryRequest` is dropped by the participant as `identity verification failed`, so the participant never replies and `AddParticipant` fails with `message send retries exhausted`.

Affects `agntcy-slim-auth` 0.13.0 (also present in the 0.12.x line).

## Root cause (two parts, both in the JWT layer)

**1. The signer never sets `kid`.** `SignerJwt::get_token()` (`src/jwt.rs`) mints the JWT with a header built as:

```rust
// src/jwt.rs — sign_claims()
let header = JwtHeader::new(self.validation.algorithms[0]);
```

i.e. `{"typ":"JWT","alg":"EdDSA"}`, with no `kid` — even though the signer knows its own subject/key.

**2. The verifier can't select a key for a kid-less token from a multi-key JWKS.** `KeyResolver::get_decoded_key_from_jwks()` (`src/resolver.rs`):

```rust
if let Some(kid) = &token_header.kid {
    // match by kid ...
} else {
    // If no key ID is specified, use the first suitable key
    for key in &jwks.keys {
        if /* key alg == token alg */ {
            return self.jwk_to_decoding_key(key);   // <-- first EdDSA key wins, no fallback
        }
    }
}
```

For a kid-less token it returns the **first** algorithm-matching key and commits to it — there is no "try all candidate keys" fallback. With an all-`EdDSA` JWKS, every kid-less token resolves to `keys[0]`, so any sender that is not `keys[0]` fails signature verification.

This is self-inflicted: slim's own signer produces exactly the kid-less tokens its own resolver cannot disambiguate.

## Observed failure

Two members (`A`, `B`) derive per-identity `did:key`/Ed25519 identities; both DIDs are placed in a shared static JWKS allow-list (`IdentityVerifierConfig`/`Jwt` with a JWKS). `A` opens a `PointToPoint` (moderated) session to `B`.

Participant `B` side:

```
slim_session::session_layer: dropping pre-session message: identity verification failed
  error=auth error: token invalid
  -> InvalidSignature   msg_type=SESSION_MESSAGE_TYPE_DISCOVERY_REQUEST
```

Moderator `A` side (repeats ~1/s, then gives up):

```
slim_session::controller_sender: timeout for message id=... msg_type=DiscoveryRequest
...
Session handshake failed: failed to add participant to session: message send retries exhausted
```

A single-key JWKS (or making the sender happen to be `keys[0]`) works — which is why single-member verification tests pass while any 2+ member mesh fails.

Note this is **not** an MLS problem: `SlimIdentityProvider::validate_member` (`agntcy-slim-mls`) correctly verifies the JWT and binds the MLS signing key to the token's `pubkey` claim, and `SignerJwt` self-generates a matching MLS keypair. The break is purely JWKS key selection at verify time.

## Impact

Per-identity JWT auth with a static multi-key JWKS allow-list — the natural configuration for a multi-party mesh — is unusable beyond one member. Group/moderated sessions cannot complete their discovery/add-participant handshake.

## Suggested fixes (any one resolves it; 1+2 is the JWS-standard combination)

1. **Signer sets `kid`**: `SignerJwt::get_token()` sets `header.kid = sub` (or the key's DID), so kid-based matching in the resolver applies.
2. **Verifier tries all candidate keys** for a kid-less token instead of only the first algorithm match, returning success on the first key that validates.
3. For `did:key` specifically: a **self-resolving verifier** that derives the Ed25519 key directly from the token's `iss`/`sub` (`did:key` embeds the public key), reducing the allow-list to a membership check and removing JWKS key-selection entirely.

muscariello:
## Problem

With a static **multi-key JWKS** allow-list (per-identity `did:key`/EdDSA members), only the *first* key was ever usable. Every other member's token failed with `InvalidSignature`, so multi-member group sessions could not form: the moderator's `DiscoveryRequest` was dropped at the participant and add-participant timed out with `message send retries exhausted`.

Root cause (two behaviors that contradict each other, both in the JWT layer):
1. slim's signer emits tokens with **no `kid`** (`sign_claims` → `JwtHeader::new(alg)`).
2. the verifier's resolver, for a kid-less token, returned the **first** algorithm-matching key with no fallback (`get_decoded_key_from_jwks`).

Every kid-less EdDSA token therefore resolved to `keys[0]`. Only single-key JWKS setups worked (the first key is the only key), which is why it went unnoticed — all group tests use a single shared `SharedSecret`, never the multi-key JWT path.

Fixes #1882.

## Fix

Resolve **all** candidate keys (exact `kid` match first, then every algorithm-compatible key) and try each during verification, advancing to the next key only on a **signature mismatch** and returning any other validation error (expiry, audience, issuer) immediately — so the real reason surfaces instead of a misleading `InvalidSignature` from an unrelated key.

- `resolver.rs`: `candidate_keys_from_jwks` returns `Vec<DecodingKey>`; new `resolve_keys` / `get_cached_keys`. The single-key `resolve_key` / `get_cached_key` are kept as thin wrappers (public API unchanged).
- `jwt.rs`: verification collects candidate keys and tries each (`verify_with_keys`).

Single-key JWKS and OIDC/`kid` flows are unchanged.

## Test

Added `test_multi_key_jwks_verifies_token_from_non_first_key`: a two-member EdDSA JWKS verifies a kid-less token from the **non-first** member, and still rejects a token signed by a non-member key. All 158 auth unit tests pass.

### DIFF
--- a/crates/auth/src/jwt.rs
+++ b/crates/auth/src/jwt.rs
@@ -347,8 +347,12 @@ impl<S> Jwt<S> {
             .as_ref()
             .ok_or(AuthError::JwtMissingPrivateKey)?;
 
-        // Create the JWT header
-        let header = JwtHeader::new(self.validation.algorithms[0]);
+        // Create the JWT header, stamping `kid = sub` so verifiers can select the
+        // right key in O(1) from a multi-key JWKS instead of trying each. `sub` is
+        // this identity's stable id (its DID in a did:key mesh); the matching JWKS
+        // entry is expected to carry the same `kid`.
+        let mut header = JwtHeader::new(self.validation.algorithms[0]);
+        header.kid = self.claims.sub.clone();
 
         // Encode the claims into a JWT token (use #[from] JwtError for propagation)
         let token = encode(&header, claims, &encoding_key.read())?;
@@ -386,6 +390,17 @@ impl<P> Jwt<P> {
 #[derive(Clone)]
 pub struct V {}
 
+/// True if `err` is specifically a signature-verification failure, i.e. the token
+/// was checked against the wrong key. Used to decide whether to try the next
+/// candidate key in a multi-key JWKS; any other error is terminal.
+fn is_signature_error(err: &AuthError) -> bool {
+    matches!(
+        err,
+        AuthError::JwtTokenInvalid(e)
+            if matches!(e.kind(), jsonwebtoken::errors::ErrorKind::InvalidSignature)
+    )
+}
+
 impl<V> Jwt<V> {
     fn try_verify_claims<Claims: serde::de::DeserializeOwned>(
         &self,
@@ -398,11 +413,11 @@ impl<V> Jwt<V> {
             return Ok(cached_claims);
         }
 
-        // Try to decode the key from the cache first; if this would require async, return WouldBlockOn
-        let decoding_key = self.decoding_key(token)?;
+        // Collect candidate keys from the cache first; if resolution would require
+        // async, return WouldBlockOn. Try each until the signature verifies.
+        let decoding_keys = self.decoding_keys(token)?;
 
-        // If we have a decoding key, proceed with verification
-        self.verify_internal::<Claims>(token, decoding_key)
+        self.verify_with_keys::<Claims>(token, decoding_keys)
     }
 
     async fn verify_claims<Claims: serde::de::DeserializeOwned>(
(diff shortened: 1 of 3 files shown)


msardara (approved):
LGTM ✅

PR #1886, opened 2026-07-21
feat(slim-bindings): browser/WASM support

hackeramitkumar:
## Summary

Adds browser/WASM support to `agntcy-slim-bindings` so the same UniFFI surface used by Go, Swift, Kotlin, and React Native can run in the browser via WebAssembly.

After moved bindings into the slim core repo, the React Native WASM work needs a **`web` feature** in `crates/slim-bindings`. This PR adds that feature while keeping the default **`native`** build unchanged.

### What changed

- **`web` feature** on `agntcy-slim-bindings`
  - Builds with `--no-default-features --features web` for browser/WASM consumers
  - `native` and `web` are mutually exclusive
  - Native-only modules (service, config, identity, transport, etc.) are excluded from web builds

- **Browser `App` implementation** (`src/web/app.rs`)
  - Native apps use `slim-service`; browsers assemble the client stack directly:
    - WebSocket `MessageProcessor`
    - `SessionLayer` + subscription manager
  - Exposes `App.connect_with_secret()` for `ws://` / `wss://` endpoints
  - Supports shared-secret identity (WebCrypto-compatible)
  - Session lifecycle: create/delete, subscribe/unsubscribe, routing, incoming session notifications

- **Shared bindings updated for wasm32**
  - **`session.rs`**: async-only API on web (blocking helpers return a clear error); wasm-friendly timeouts via `tokio::time::sleep`; MLS settings expose `max_seen_control_message_ids_size`
  - **`completion_handle.rs`**: wasm timeout handling; blocking `wait()` disabled on web
  - **`errors.rs`**: web-safe error mapping

- **Dependency layout**
  - **Native (`not wasm32`)**: existing workspace deps including `agntcy-slim-service`
  - **WASM (`wasm32`)**: browser-capable crates only (`auth`, `config`, `datapath`, `session`) + `tokio_with_wasm` + `wasm-bindgen-futures`
  - UniFFI enabled with `wasm-unstable-single-threaded`

- **Workspace UniFFI pin**: `0.31.1` → `0.31.0` to align with `uniffi-bindgen-react-native@0.31.0-3` used by the slim-bindings React Native WASM generator

### Architecture

```
Native (default)                    Web (feature = "web")
─────────────────                   ────────────────────
App via slim-service        →       App via WebSocket + SessionLayer
Full config/service API     →       App / Session / Name / CompletionHandle only
tokio (native)              →       tokio_with_wasm (browser event loop)
```

Language-specific wrappers (FFI adapter, JSI, TypeScript) remain in the **[slim-bindings](https://github.com/agntcy/slim-bindings)** repo and depend on this crate via `agntcy-slim-bindings-ffi`.

### Non-goals (this PR)

- No changes to `slim-service` or server-side components
- No SlimRPC exports in web builds
- No new CI workflow for wasm32 (can follow in a separate PR)

---


crates/slim-bindings/Cargo.toml
```diff
@@ -9,7 +9,27 @@ description = "UniFFI bindings and FFI adapter for SLIM data plane - enables lan
 [lib]
 name = "slim_bindings"
 
+[features]
+default = ["native"]
+# Full desktop/mobile bindings, including services, configuration, and SlimRPC.
+native = []
+# Browser-safe UniFFI surface. Mutually exclusive with `native`; do not enable
+# both at once (CI runs native and web clippy as separate `cargo clippy` passes).
+web = []
```
msardara, line 15: Do we need this web feature? Normally we just want to target the wasm architecture and that would be used as library by the slim-bindings generated code 🤔

hackeramitkumar: wasm32 is correct for the final library, but UniFFI metadata is extracted from a host build where target_arch != wasm32. The web feature forces the browser module tree + stub App during that host bindgen step (ubrn passes --features web --no-default-features). Without it, we’d generate native Service/config/RPC types into the browser bindings. Target cfg inside web/app.rs then switches stub vs real implementation for the actual wasm32 compile.

Cargo.toml
```diff
@@ -206,7 +206,7 @@ tracing-subscriber = "0.3.19"
 tracing-test = { version = "0.2.5", features = ["no-env-filter"] }
 trait-variant = "0.1"
 twox-hash = "2.1.1"
-uniffi = "0.31.1"
+uniffi = "0.31.0"
```
amitami2, line 220: This was needed for making the uniffi compatible with the ubrn latest version.

msardara (approved):
LGTM ✅

PR #1887, opened 2026-07-21
feat: increase post-quantum crypto coverage for wasm and add hybrid key exchange

markpmarton:
## Summary
Adds an opt-in, deployment-wide post-quantum policy (`enforce_pqc`) that applies consistently across TLS, inter-node link negotiation, and MLS. When enabled, all three layers use hybrid classical + ML-KEM-768 algorithms; when disabled, behavior stays classical.

### Config
  • Introduces `EnforcePqcPolicy` (`slim_config::pqc`) as the single source of truth PQC
    switch.
  • `ServiceConfiguration::prepare()` / `normalize_pqc()` propagate the policy
    to every TLS endpoint’s runtime config and validate that `enforce_pqc`
    requires tls_version: "tls1.3".
  • Example config: `config/pqc/server-config.yaml`.

### TLS
  • When enforced, rustls is built with only the `X25519MLKEM768` hybrid
    key-exchange group via `crypto_provider_for_config()`.
  • TLS 1.3 is required and validated at config load time when PQC is
    enforced.

### Link negotiation
  • Extends `LinkNegotiationPayload` with optional `link_kem_payload`.
  • Hybrid flow: initiator sends an ML-KEM-768 public key;
    responder returns ciphertext; both sides derive the
    per-link HMAC from hybrid X25519 + ML-KEM-768 key agreement.
  • When `enforce_pqc` is set, negotiation fails closed if KEM material is
    missing, wrong length, or hybrid HMAC installation fails.
  • When `enforce_pqc` is off, responders still accept hybrid negotiation if
    the initiator sends a valid ML-KEM public key (backward-compatible
    opt-in per link).

### MLS
  • Replaces the compile-time CIPHERSUITE constant with runtime
    `ciphersuite_for(enforce_pqc)`.
  • When enforced on native, uses ML_KEM_768_X25519; `mls-rs-crypto-awslc`
    gains the post-quantum feature.
  • `enforce_pqc` MLS remains native-only — wasm32 returns a clear error
    because browser WebCrypto cannot service the PQ ciphersuite.

### WASM / browser
  • Link-level hybrid KEM is supported on wasm32 via the `ml-kem`
    backend, alongside existing X25519/HKDF.
  • Native↔browser interoperability is covered by cross-backend tests
    (aws-lc ↔ pure) and a wasm browser test for hybrid link MAC agreement.

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


crates/service/src/service.rs
```diff
@@ -265,13 +279,18 @@ impl Service {
     }
 
     /// Create a new Service with configuration
-    pub fn new_with_config(id: ID, config: ServiceConfiguration) -> Self {
+    pub fn new_with_config(id: ID, mut config: ServiceConfiguration) -> Self {
+        config
+            .prepare()
+            .expect("invalid service configuration: enforce_pqc policy");
+
```
micpapal: here either we expect a valid config, see for example the config.prepare()? in fn build_server that is called before new_with_config, and so we can skip the prepare(),  or we need to return a Result to avoid to panic.

crates/datapath/src/link_ecdh/backend_pure.rs
```diff
@@ -4,15 +4,59 @@
...
+
+    fn try_next_u64(&mut self) -> Result<u64, Self::Error> {
+        let mut buf = [0u8; 8];
+        self.try_fill_bytes(&mut buf)?;
+        Ok(u64::from_le_bytes(buf))
+    }
+
+    fn try_fill_bytes(&mut self, dest: &mut [u8]) -> Result<(), Self::Error> {
+        if let Err(e) = getrandom::fill(dest) {
+            self.error = Some(e);
```
micpapal, line 56: is it ok here to store the error but return ok()? for the caller this try will always succeed.

crates/datapath/src/link_ecdh/backend_awslc.rs
```diff
@@ -45,12 +50,78 @@ pub fn derive(
...
+    let pk = ek
+        .key_bytes()
+        .map_err(|_| HeaderMacError::KeyGenerationFailed("ml-kem-768 pk bytes".into()))?;
+    Ok((dk, pk.as_ref().to_vec()))
+}
+
+pub fn encapsulate_mlkem768(pk: &[u8]) -> Result<(Vec<u8>, [u8; 32]), HeaderMacError> {
+    let ek =
+        EncapsulationKey::new(&kem::ML_KEM_768, pk).map_err(|_| HeaderMacError::KeyAgreement)?;
+    let (ct, ss) = ek.encapsulate().map_err(|_| HeaderMacError::KeyAgreement)?;
```
micpapal, line 80: here ss is always at least 32 bytes? because otherwise shared.copy_from_slice(ss.as_ref()) may fail

crates/datapath/Cargo.toml
```diff
@@ -85,7 +87,9 @@ criterion = { workspace = true }
 getrandom = { workspace = true }
 hkdf = { workspace = true }
 hmac = { workspace = true }
+ml-kem = { workspace = true }
 opentelemetry_sdk = { workspace = true }
+rand_core = "0.10"
```
micpapal, line 97: should we add also rand_core to the workspace instead to track it here?

crates/config/src/tls/client.rs
```diff
@@ -296,8 +297,11 @@ impl TlsClientConfig {
         // Load CA root store
         let root_store = self.config.load_ca_cert_pool().await?;
 
-        // Base builder
-        let config_builder = RustlsClientConfig::builder_with_protocol_versions(&[tls_version])
+        // Validate TLS version and create base builder
+        self.config.validate_pqc()?;
+        let crypto_provider = crypto_provider_for_config(self.config.enforce_pqc);
+        let config_builder = RustlsClientConfig::builder_with_provider(crypto_provider.clone())
+            .with_protocol_versions(&[tls_version])?
```
micpapal, line 304: the clone here most likelly is not needed

crates/datapath/src/negotiation.rs
```diff
@@ -223,34 +273,58 @@ where
...
-                                        .to_string(),
-                                ));
-                            }
-                        }
-                    }
+            let (server_sk, server_pk) = link_ecdh::generate_x25519_ephemeral().map_err(|_| {
+                DataPathError::NegotiationError("failed to generate server exchange key".into())
+            })?;
+
+            let hybrid = params.enforce_pqc || kem_len(&payload) == ML_KEM768_PUBLIC_KEY_LEN;
```
micpapal, line 282: is it safe to rely only in the len of payload here?

micpapal (changes requested):

crates/service/src/service.rs
markpmarton: Yes, you're right, the prepare here would be only necessary if we would like to define `enforce_pqc` through the bindings, and currently it is not the case. I'll remove this one.

crates/datapath/src/link_ecdh/backend_pure.rs
markpmarton: Unfortunately the `ml-kem` lib needs the RNG to be `Infailable`
https://github.com/agntcy/slim/pull/1887/changes#diff-e5f4407394f22fa530448669c721e80b91563ef5c820dbcad32c39cb2bf717beR37

The existance of the error is checked after the generator is called:
https://github.com/agntcy/slim/pull/1887/changes#diff-e5f4407394f22fa530448669c721e80b91563ef5c820dbcad32c39cb2bf717beR93

crates/datapath/src/link_ecdh/backend_awslc.rs
markpmarton: Yes, for this algo the secret will always be 32 bytes on every architecture. In `aws-lc-rs` it is a const:
`const ML_KEM_768_SHARED_SECRET_LENGTH: usize = 32;`

crates/datapath/src/negotiation.rs
markpmarton: Yes, as `encapsulate_mlkem768` revalidates pk before parse. Invalid key with same length will fail.
https://github.com/agntcy/slim/pull/1887/changes#diff-cdbbde742ed65bff7d84d06ca72b40497fb3a102980f5a8b0cc9386ef464be6bR292

micpapal (approved):

PR #1895, opened 2026-07-23
fix: prevent fail if a participant is offline

micpapal:
# Description

When a reliable message times out after all retries, some participants in the group may have silently gone offline without sending an ack. Previously, the session treated this as a hard send failure and reported an error to the application, even though the message was successfully delivered to all reachable peers.

This change introduces the following improvements:

- SessionSender::missing_acks_for(id) — queries which participants have not yet acked a given message id, by reading the existing GroupTimer.missing_timers set before it is cleared on failure.
- MessageHandler::missing_acks_for — default no-op trait method so all layers can expose this without breaking changes; Session overrides it to delegate to SessionSender.
- SessionParticipant — on TimerFailure for a data message, marks each non-acking participant Offline in group_list and calls remove_endpoint, before forwarding the failure to the inner layer.
- SessionModerator — same logic via a shared mark_missing_offline helper, called in two places: on TimerFailure for data messages, and in handle_failure when a control-message task partially succeeded (some participants acked, some did not).
- SessionSender::on_failure — now signals Ok(()) to the app-level ack notifier instead of Err, since a missing ack means the peer was unreachable (now marked offline), not that the send failed.
- TaskUpdate::partial_success — new trait method that each moderator task implements to signal whether the main goal was achieved even if not all acks arrived (e.g. AddParticipant succeeded if welcome + commit were received, regardless of whether every group-update ack arrived).

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


crates/session/src/session_sender.rs
```diff
@@ -468,9 +468,11 @@ impl SessionSender {
 
         self.pending_acks.remove(&id);
 
-        // Signal failure to the ack notifier if present
+        // Signal success: the message reached all reachable participants. Non-replying
+        // participants are marked offline (in participant or moderator), so a
+        // missing ack means the peer was unreachable, not that the send failed.
         if let Some(tx) = self.ack_notifiers.remove(&id) {
-            let _ = tx.send(Err(error));
+            let _ = tx.send(Ok(()));
```
msardara: I'd limit the `Ok(())` to the timer-failure only, and keep `on_slim_failure` returning `Err`.

crates/session/src/session_sender.rs
```diff
@@ -454,7 +454,7 @@ impl SessionSender {
     pub fn on_failure(
         &mut self,
         id: u32,
-        error: SessionError,
+        _error: SessionError,
```
msardara: on_failure is also called by `on_slim_failure`, should we handle that error?

crates/session/src/moderator_task.rs
```diff
@@ -302,6 +318,13 @@ impl TaskUpdate for AddParticipant {
             && self.welcome.received
             && self.commit.received
     }
+
+    fn partial_success(&self) -> bool {
```
msardara, line 322: Two problems:

partial_success It's identical to task_complete. Both check discovery && join && welcome && commit (lines 315-319), so partial_success is never true unless the task is already fully complete. That means Add has no real "partial" state, which contradicts the PR description ("succeeded if welcome + commit were received").

The branch that uses it can never run for Add. handle_failure only enters its partial_success branch when ack_tx is still present (session_moderator.rs:581-582). But update_phase_completed already takes ack_tx and sends Ok(()) as soon as welcome+commit arrive (lines 304-310). So by the time a failure reaches handle_failure:
- if welcome+commit arrived → ack_tx is already gone → branch skipped;
- if they didn't → partial_success is false → Err path.

The result: for Add, mark_missing_offline is never called from this path, so missing members are never marked offline here.

crates/session/src/moderator_task.rs
```diff
@@ -27,6 +27,11 @@ pub(crate) trait TaskUpdate {
     fn proposal_start(&mut self, timer_id: u32) -> Result<(), SessionError>;
     fn update_phase_completed(&mut self, timer_id: u32) -> Result<(), SessionError>;
     fn task_complete(&self) -> bool;
+    // this returns true if the main goal of the task was achieved but not all the phases
+    // were completed. for example in add participant the task partially succeed if the welcome
+    // message if received but some ack for the update group is not received. this may happen
```
msardara:
```suggestion
    // message is received but some ack for the update group is not received. this may happen
```

crates/session/src/session_moderator.rs
```diff
@@ -579,13 +581,37 @@ where
         if let Some(task) = self.current_task.as_mut()
             && let Some(ack_tx) = task.ack_tx_take()
         {
-            let _ = ack_tx.send(Err(task.failure_message(error)));
+            // check it the task was partially successful. if yes send ok to the app and
```
msardara:
```suggestion
            // check if the task was partially successful. if yes send ok to the app and
```

crates/session/src/session_moderator.rs
```diff
@@ -579,13 +581,37 @@ where
...
+            } else {
+                let _ = ack_tx.send(Err(task.failure_message(error)));
+            }
         }
 
         self.current_task = None;
         self.pop_task().await
     }
 
+    fn mark_missing_offline(&mut self, missing: &[ProtoName]) {
```
msardara: This looks like the loop in `session_participant.rs` (the participant version does the same marking), with the difference that here we call`reset_id()` before the `group_list` lookup.

We could use this helper in both and call reset_id based on a flag passed by the caller.

crates/session/src/session_participant.rs
```diff
@@ -283,6 +283,22 @@ where
                         }
                     }
                 } else {
+                    // here we are missing an ack from some one in the group, so we move it
+                    // offline (if not yet) and notify the inner layer that the message failed to be delivered
+                    let missing = self.inner.missing_acks_for(message_id);
+                    for participant_name in &missing {
```
msardara: This looks like the same of`mark_missing_offline` helper. Should we try to have a common helper here?

crates/session/src/session_moderator.rs
```diff
@@ -579,13 +581,37 @@ where
         if let Some(task) = self.current_task.as_mut()
             && let Some(ack_tx) = task.ack_tx_take()
         {
-            let _ = ack_tx.send(Err(task.failure_message(error)));
+            // check it the task was partially successful. if yes send ok to the app and
+            // put offline all the participant that did not ack
```
msardara:
```suggestion
            // put offline all the participants that did not ack
```

msardara (changes requested):

msardara (approved):

PR #1901, opened 2026-07-24
feat(channel-manager): add storage

micpapal:
# Description

Integrates agntcy-slim-persistence into the channel-manager so that sessions survive a restart.

Session persistence on creation — SessionModerator::init() now calls persist_state() immediately when a session starts, writing the session record and app identity to the encrypted SQLite store. Previously, persistence only triggered when the first participant joined, so channels with no active participants were never durably recorded.

Session restore on startup — after connecting to SLIM and creating the app, the channel-manager reads all session records from the store and re-registers each session. Config-driven channels are skipped if they were already restored, avoiding duplicates.

Graceful shutdown mode — a new delete-sessions-on-shutdown flag (under persistence:) controls shutdown behaviour:
- true (default): sessions are fully deleted from SLIM on shutdown (same as before persistence was added)
- false: sessions are closed with UpdateParticipantState::Offline and kept in the store, so they are automatically restored on the next startup

KV cleanup on delete — session_layer::remove_session() now deletes the session record from the KV store when a session is explicitly removed, so a restart after a delete-sessions-on-shutdown: true shutdown does not attempt to restore stale sessions.

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/crates/channel-manager/base-config.yaml
+++ b/crates/channel-manager/base-config.yaml
@@ -16,6 +16,11 @@ channel-manager:
 
   auth:
     type: shared_secret
-    secret: "a-very-long-shared-secret-abcdef1234567890"
+    secret: "1234567890-1234567890-1234567890-123456"
 
   channels: []
+
+  persistence:
+    path: /tmp/slim/channel-manager
+    encryption-passphrase: "a-strong-passphrase"
+    delete-sessions-on-shutdown: false
--- a/crates/channel-manager/base-config.yaml
+++ b/crates/channel-manager/base-config.yaml
@@ -16,11 +16,33 @@ channel-manager:
 
   auth:
     type: shared_secret
+    # In production inject the secret at runtime, e.g.:
+    #   secret: ${env:CM_AUTH_SECRET}
+    #   secret: ${file:/run/secrets/cm-auth-secret}
     secret: "1234567890-1234567890-1234567890-123456"
 
+  # Channel definition controls the operating mode:
+  #
+  # Config mode (channels non-empty): the config file is the source of truth.
+  #   Channels and participants are reconciled against the DB on startup.
+  #   gRPC write APIs (create/delete channel, add/remove participant) are
+  #   disabled — modify this file and restart to change the topology.
+  #
+  # API mode (channels empty): the DB is the source of truth.
+  #   All sessions restored from the DB are used as-is and the full gRPC
+  #   API is available for dynamic channel management.
   channels: []
-
+    
   persistence:
+    # Both path and passphrase support ${ENV_VAR} substitution and can be
+    # injected at runtime via a secrets manager (Kubernetes Secret, Vault,
+    # External Secrets Operator, etc.), for example:
+    #   path: ${env:CM_PERSISTENCE_PATH}
+    #   encryption-passphrase: ${env:CM_PASSPHRASE}
+    # A file-based secret is also supported:
+    #   encryption-passphrase: ${file:/run/secrets/cm-passphrase}
+    # In production use a private directory with restrictive permissions
+    # and never store the passphrase in the config file directly.
     path: /tmp/slim/channel-manager
-    encryption-passphrase: "a-strong-passphrase"
+    insecure: true
     delete-sessions-on-shutdown: false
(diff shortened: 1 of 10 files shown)


crates/channel-manager/src/bin/channel_manager.rs
```diff
@@ -44,6 +45,12 @@ async fn create_channels_from_config(
     config: &Config,
 ) -> anyhow::Result<()> {
     for channel_cfg in &config.manager.channels {
+        // Skip channels already restored from persistent storage.
+        if sessions.get_session(&channel_cfg.name).await.is_some() {
```
muscariello: Restoring by name skips all participant reconciliation. A config change—or a crash after the creation-time checkpoint but before every invite completes—leaves the stored roster incomplete indefinitely. Please reconcile configured participants instead of skipping the channel wholesale.

crates/channel-manager/src/bin/channel_manager.rs
```diff
@@ -235,11 +289,24 @@ async fn main() -> anyhow::Result<()> {
...
-        Ok(()) => info!("All sessions cleaned up"),
-        Err(_) => warn!("Session cleanup timed out, forcing shutdown"),
+    let delete_on_shutdown = config
+        .manager
+        .persistence
+        .as_ref()
+        .map(|p| p.delete_sessions_on_shutdown)
+        .unwrap_or(true);
+    if delete_on_shutdown {
+        match tokio::time::timeout(Duration::from_secs(5), sessions.delete_all(&arc_app)).await {
```
muscariello: This 5s timeout wraps a sequential loop whose individual deletions can each wait 5s. One slow channel prevents later records from being deleted, so `delete-sessions-on-shutdown: true` can still restore them next boot. Please run cleanup concurrently or bound each session independently.

crates/channel-manager/src/sessions.rs
```diff
@@ -109,16 +97,52 @@ impl SessionsList {
...
+        for name in self.list_channel_names().await {
+            let Some(weak) = self.take_channel(&name).await else {
+                continue;
+            };
+            let Some(session) = weak.upgrade() else {
+                continue;
+            };
+            match session.close().await {
+                Ok(completion) => {
+                    if tokio::time::timeout(std::time::Duration::from_secs(5), completion)
```
muscariello: `timeout` returns `Ok(Err(e))` when closing completes with a session error, but `.is_err()` treats that as success. Please handle `Ok(Ok(()))`, `Ok(Err(e))`, and `Err(_)` separately.

muscariello (changes requested):
Persistence lifecycle needs these cases addressed before merge.

muscariello (commented):
Security concerns for persisted MLS state:

crates/channel-manager/src/config.rs
```diff
@@ -44,6 +52,35 @@ pub struct Config {
...
+/// group state and session records in an encrypted SQLite database, and the
+/// channel manager restores all previously-known sessions on startup.
+#[derive(Clone, Debug, Deserialize)]
+pub struct PersistenceConfig {
+    /// Directory where the SQLite database will be written.
+    pub path: PathBuf,
+
+    /// Optional passphrase used to derive the AES-256 encryption key.
+    /// Without one, values are still encrypted but the key is derived from
+    /// the (public) app name — adequate for at-rest protection but not a
```
muscariello: The fallback key is derived from the public app name, so it provides no confidentiality for stored MLS epoch secrets. Please require a secret when persistence is enabled, or gate this behind an explicit insecure option.

crates/channel-manager/base-config.yaml
```diff
@@ -16,6 +16,11 @@ channel-manager:
...
   auth:
     type: shared_secret
-    secret: "a-very-long-shared-secret-abcdef1234567890"
+    secret: "1234567890-1234567890-1234567890-123456"
 
   channels: []
+
+  persistence:
+    path: /tmp/slim/channel-manager
+    encryption-passphrase: "a-strong-passphrase"
```
muscariello: This enables persistence with a repository-known passphrase under `/tmp`. Anyone who can read the database can decrypt the session and MLS state. Please use runtime secret injection and a private state directory with restrictive permissions.

muscariello:
passphrases still use fixed-salt HKDF, which is fast and offers little resistance to offline guessing of weak passwords. At minimum, documentation should require a high-entropy random secret;

muscariello (approved):
LGTM

PR #1902, opened 2026-07-24
fix(session): remove record + MLS state + pool entry on close (both roles)

msardara:
## Summary

When a session is closed it should leave no trace: removed from the session-layer pool, its persisted record deleted, **and** its MLS group state deleted — so it is neither left dangling in memory nor restored on the next start. A **graceful shutdown** must instead preserve everything so the session can be restored. This now works uniformly for **both moderator and participant** sessions.

## Design

The `DeleteSession` signal is emitted **once, by the session controller's processing loop** (right after `on_shutdown`), for **both roles**. Previously only the moderator sent it, via a role-specific `send_close_signal`; consolidating it in the controller means a participant's own hard close cleans up too.

The layer's `DeleteSession` handler — the single teardown point — does the full cleanup: removes the pool entry, reads the session record for its `group_id` and deletes the **MLS group state**, then deletes the **session record**. Pool removal is now centralized here: `clear_all_sessions` and the restore-failure path no longer mutate the pool directly — they `terminate()` (→ `DeleteSession`) instead, so every removal goes through the one path that also cleans up the record + MLS state.

This is safe for restore because the signal is a **no-op on a graceful shutdown**: when the whole app/service is dropped, the layer's `listen_from_sessions` receiver is already gone, so the send isn't processed and the persisted state is preserved. On a hard close (runtime alive) the send is processed and the footprint is removed.

## Behavior

| Path | Pool entry | Session record | MLS group state |
|------|------------|----------------|-----------------|
| Hard close — moderator or participant (runtime alive) | removed | **deleted** | **deleted** |
| Group leave (`on_leave_request`) | removed | **deleted** | **deleted** |
| Graceful shutdown / `Drop` | (torn down) | **kept** — restore works | **kept** |

## Changes

- `session_controller.rs`: emit `DeleteSession` from the processing loop after `on_shutdown`, for both roles.
- `session_moderator.rs`: drop the role-specific `send_close_signal` (now handled by the controller).
- `persistence/group_storage.rs`: add `SlimGroupStateStorage::delete_group` (delegates to the in-memory and SQLite providers).
- `session_layer.rs`: the `DeleteSession` handler deletes the MLS group state and the persisted record in addition to removing the pool entry; add `stored_mls_group_ids` introspection.
- `service/app.rs`: expose `App::get_session(id)` and `App::stored_mls_group_ids()` (introspection / test helpers).
- `session_restore_test.rs`: `test_hard_close_deletes_persisted_session` (moderator) and `test_participant_hard_close_deletes_persisted_session` (participant) assert the pool entry + record + MLS group state are present before the close and gone afterwards.

## Note

The SQLite provider's `delete_group` removes the group snapshot (which makes the group unloadable); a few prior-epoch rows may remain as harmless dead weight.

## Testing

- Workspace builds; 250 `session` unit tests pass.
- All 6 `session_restore_test` e2e tests pass: hard-close deletion for **both** moderator and participant, and the graceful-restart tests confirming shutdown preserves record **and** MLS state so restore works.

### DIFF
--- a/crates/session/src/session_controller.rs
+++ b/crates/session/src/session_controller.rs
@@ -504,6 +504,19 @@ impl SessionController {
         if let Err(e) = inner.on_shutdown().await {
             tracing::error!(error = %e.chain(), "error during shutdown of session");
         }
+
+        // Signal the session layer that this session is done, so it removes it
+        // from the pool and deletes its persisted footprint (record + MLS
+        // state). Done here — once, for both moderator and participant — so the
+        // teardown is uniform regardless of role. On a graceful process
+        // shutdown the layer's receiver is already gone, so the signal is a
+        // no-op and the persisted state is preserved for restore.
+        let _ = settings
+            .tx_to_session_layer
+            .send(Ok(SessionMessage::DeleteSession {
+                session_id: settings.id,
+            }))
+            .await;
     }
 
     /// getters
--- a/crates/persistence/src/group_storage.rs
+++ b/crates/persistence/src/group_storage.rs
@@ -109,6 +109,23 @@ impl SlimGroupStateStorage {
             SlimGroupStateStorage::Sqlite(_) => true,
         }
     }
+
+    /// Delete a group's stored MLS state so it can no longer be loaded. Used
+    /// when a session is permanently closed, so its MLS group is not left behind
+    /// in the store. Deleting an unknown group is a no-op.
+    pub fn delete_group(&self, group_id: &[u8]) -> Result<(), PersistenceError> {
+        match self {
+            SlimGroupStateStorage::InMemory(s) => {
+                s.delete_group(group_id);
+                Ok(())
+            }
+            #[cfg(not(target_arch = "wasm32"))]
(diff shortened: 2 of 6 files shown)


crates/session/src/session_controller.rs
```diff
@@ -504,6 +530,19 @@ impl SessionController {
...
         }
+
+        // On an explicit hard close (not a plain Drop / graceful shutdown), ask
+        // the session layer to drop us: it removes us from its pool and deletes
+        // our persisted record so we are not restored on the next start. A Drop
+        // leaves `hard_close` false, so a graceful shutdown preserves the record.
+        if hard_close.load(Ordering::SeqCst) {
+            let _ = settings
+                .tx_to_session_layer
+                .send(Ok(SessionMessage::DeleteSession {
```
micpapal: this done already in the moderator at line send_close_signal and in the participant in on_leave_request. Do we need it also here? if we do and hard_close both function should be called.

micpapal (approved):

PR #1913, opened 2026-07-29
feat: add control-plane side override of node connection data

### LINKED ISSUE #1794: Move client connection params to node config with CP override
# Summary
In the current state `compute_client_config` builds a `ServerConnectionConfig` for every link with hardcoded 2s backoff values. When a node received this config, `merge_server_requirements` unconditionally overrides the local settings with the CP supplied backoff.

# Problem
The backoff strategy is only decided by the CP, overriding local settings unconditionally. With a heterogenous infrastructure there might be need to have custom configuration values for backoff, timeout and keepalive.

# Proposed change
Local settings should be the source of truth for the nodes. The node's own `ClientConfig` should apply keepalive, timeout and backoff unless the control-plane overrides it.
CP config should have optional fields for processing the enforced synced values.

markpmarton:
closes #1794 

# Description

The CP unconditionally overrode every connecting node's backoff with a hardcoded 2 000 ms value (see generate_config_data), ignoring whatever the node had configured locally. There was no mechanism for operators to intentionally enforce connection parameters cluster-wide.

# Changes

- Remove hardcoded backoff: `Some(2000)` in `generate_config_data` — reads from `ConnectionDetails` instead- Node default backoff changed to 2s fixed interval (preserves old behaviour)
- `ConnectionDetails` gains optional `backoff`, `timeout`, `keepalive` fields
- `ServerConnectionConfig` gains `keepalive` field; `merge_server_requirements` applies it
- Add `NodeConnectionParams` struct to CP config, exposed as `enforce_node_connection` in YAML
- On node registration, stamp `NodeConnectionParams` values onto `ConnectionDetails` (survives reconnects)

# Config example
```
enforce_node_connection:
  backoff: 3000        # ms, fixed-interval
  timeout: 5000        # ms
  keepalive:
    http2_keepalive: "45s"
    timeout: "8s"
    keep_alive_while_idle: true
```

All three fields are optional; omitting the entire block (or the block itself) leaves node-local settings untouched.

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


docs/content/slim/components/controller/config.md
```diff
@@ -108,6 +108,38 @@ topology:
           neighbors: [$domain]
 ```
 
+## Node Connection Enforcement
```
micpapal: move this config at the end of the topology session as now it is in the middle.

crates/control-plane/src/config.rs
```diff
@@ -8,10 +8,25 @@ use serde::Deserialize;
...
+/// CP-enforced connection parameters applied to all nodes on registration.
+/// Any field set here overrides the node's local configuration.
+/// Omitted fields leave the node's local settings untouched.
+#[derive(Debug, Clone, Default, Deserialize)]
+#[serde(default)]
+pub struct NodeConnectionParams {
+    /// Fixed-interval backoff to enforce (milliseconds).
+    pub backoff: Option<u32>,
+    /// Connect timeout to enforce (milliseconds).
+    pub timeout: Option<u32>,
```
micpapal, line 25: is it ok if backoff or timeout are set to 0 or do we need to check the input?

micpapal (approved):
check the small comments before merge

crates/control-plane/src/config.rs
markpmarton: The 0 value is safe, but `from_secs()` gave wrong results for <1000 values. FIxed.

PR #1960, opened 2026-08-10
feat(slimctl): OIDC token refresh for long-lived nodes

markpmarton:
# Description

`slimctl` login did not request a refresh token from the IdP, so long-lived client nodes had no way to recover when their access token expired. After a crash and restart with an expired token, the node would fail to reconnect. Additionally, IdPs that omit group membership from the JWT payload were not usable with claim-based policy filters.

# Changes

- Added offline_access and groups to the OIDC authorization scope so the IdP issues a refresh token and includes group membership
- Added token_endpoint field to `OidcCredentials` (persisted in `credentials.yaml`) so the refresh grant can be called without re-running discovery
- On startup, if credentials contain a refresh token, a background task is spawned that reads the exp claim from the access token and sleeps until 60 s before expiry, then calls `grant_type=refresh_token` against the stored token_endpoint
- Refreshed tokens are written via `save_credentials()`, which already writes `~/.slimctl/token` ; the existing notify file watcher in the gRPC middleware picks up the new token automatically with no connection restart
- On startup, if the stored access token is already expired (e.g. after a crash), the refresh is performed synchronously before any subcommand runs, so the initial connection attempt uses a valid token
- `OidcVerifier` now fetches the OIDC `userinfo` endpoint after every JWT verification and merges the response claims into the claims visible to the policy engine 
- enabling group-based CEL/Rego policies against IdPs that omit groups from the token payload

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


crates/auth/src/oidc.rs
```diff
@@ -595,15 +610,12 @@ impl Verifier for OidcVerifier {
...
     where
         Claims: serde::de::DeserializeOwned + Send,
     {
-        if let Some(cached_jwks) = self.jwks_cache.get(&self.issuer_url) {
-            self.verify_token_util(token.as_ref(), &cached_jwks)
-        } else {
-            Err(AuthError::WouldBlockOn)
-        }
+        // ponytail: always async — userinfo fetch has no sync path
+        Err(AuthError::WouldBlockOn)
```
msardara, line 651: Do we need to fetch the userinfo here? I think the token already contains all the claims we need 🤔

crates/auth/src/refresh_token.rs
```diff
@@ -0,0 +1,255 @@
...
+        let c = cached.as_ref().ok_or(AuthError::GetTokenError)?;
+        let token = c.token.clone();
+        let now = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .map(|d| d.as_secs())
+            .unwrap_or(0);
+        let needs_refresh = now >= c.refresh_at;
+        drop(cached);
+
+        if needs_refresh && !self.refreshing.swap(true, Ordering::Relaxed) {
```
msardara: What happens if someone calls this on an expired token? I guess it will get the expired and it will trigger the refresh token task in background, so the first call will get an expired token and the second will get a fresh one.

I think in the initialize() call we should start a background task that refresh the token asynchronously, so that get_token always get a valid one.

crates/slimctl/src/config.rs
```diff
@@ -215,16 +227,32 @@ pub fn save_credentials(creds: &OidcCredentials) -> Result<()> {
     std::fs::create_dir_all(dir)
         .with_context(|| format!("failed to create config directory: {}", dir.display()))?;
     let data = serde_yaml::to_string(creds).context("failed to serialize credentials")?;
-    std::fs::write(&path, data)
+    std::fs::write(&path, &data)
         .with_context(|| format!("failed to write credentials: {}", path.display()))?;
+    std::fs::set_permissions(&path, Permissions::from_mode(0o600))
+        .with_context(|| format!("failed to set permissions on: {}", path.display()))?;
```
msardara: Here we should probably write directly with the permission 600 (rather than issuing 2 calls) to avoid a window where the file is stored with the wrong permissions

msardara (changes requested):

crates/auth/src/refresh_token.rs
markpmarton: Right, I moved the refresh spawn to a background task in initialize. Also added fail-fast for expired token into `get_token`, so the bg task will be the only one handling the refresh. Thank you!

crates/auth/src/oidc.rs
markpmarton: Some IdP-s don't provide group claims, only is the userinfo. Should we add a flag into the config if we want to validate in the userinfo also?

crates/slimctl/src/config.rs
markpmarton: Done.

msardara (approved):
LGTM ✅

PR #1961, opened 2026-08-11
fix(rpc): signal end-of-stream with the status header, not an empty payload

janosSarusiKis:
# Description

is_eos() treated any Ok message with an empty payload as end-of-stream, and send_response_stream stamped slimrpc-code = Ok onto every response data frame. A response that encodes to zero bytes was therefore indistinguishable from the terminator and got consumed as one — Channel::unary returned Internal: "No response received".

This hits every RPC returning google.protobuf.Empty, and any response with no fields set (an empty list result, for example). It surfaced in the A2A Go SDK, where DeleteTaskPushNotificationConfig fails on every successful delete.

Response data frames no longer carry a status, so the presence of the status header is the terminal signal — matching gRPC's grpc-status trailer, and matching what the client side already did via first_msg_metadata / continuation_metadata. is_eos() keeps the empty-payload check alongside the status-presence check: presence alone would make a client on this version read the first data frame from an older server as EOS, since those stamp Ok on every frame. With both clauses, behaviour against older peers is unchanged rather than broken — no wire break.

Also updates the demux filter in server.rs, which identified echoed server responses by status presence; it now keys off RPC_DIR_KEY directly, which is what it meant.

Adds regression coverage for zero-byte unary and streamed responses, and extends test_request_stream to cover all four frame shapes including a legacy-style data frame carrying both a status and a payload.

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/crates/rpc/src/channel.rs
+++ b/crates/rpc/src/channel.rs
@@ -63,7 +63,7 @@ use super::{
     Context, METHOD_KEY, Metadata, RPC_DIR_KEY, RPC_DIR_REQ, RPC_ID_KEY, ReceivedMessage, RpcCode,
     RpcError, SERVICE_KEY, STATUS_CODE_KEY, calculate_timeout_duration,
     codec::{Decoder, Encoder},
-    mark_empty_data_frame, send_eos,
+    send_eos,
     session_wrapper::{SessionRx, SessionTx, new_session},
 };
 
@@ -457,13 +457,12 @@ impl Channel {
         let mut first = true;
         while let Some(request) = request_stream.next().await {
             let request_bytes = request.encode()?;
-            let mut msg_meta = if first {
+            let msg_meta = if first {
                 first = false;
                 first_msg_metadata(ctx, service_name, method_name, rpc_id)
             } else {
                 continuation_metadata(rpc_id)
             };
-            mark_empty_data_frame(&mut msg_meta, &request_bytes);
             let handle = session
                 .publish(
                     session.destination(),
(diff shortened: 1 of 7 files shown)


Tehsmash:
I did some double checking on the gRPC implementation to see if we could learn more about how they handle this scenario as our code was originally based on theirs.

It appears that they use the grpc-status trailer(header) as the primary indicator that the stream is finished: Normal stream packets do not include the "grpc-status" header, final packet includes the grpc-status header either OK (stream completed successfully) or one of the error codes (stream/server encountered an error) so its the end of the stream.

I think instead of adding a separate marker we could probably do the same logic WDYT?

Tehsmash:
Actually looking at the original code what I described is how its supposed to work already, but perhaps we're accidentally adding the srpc-status header somewhere we're not supposed to be

Edit: yup found it here: https://github.com/agntcy/slim/blob/main/crates/rpc/src/rpc_session.rs#L223 status shouldn't be here I think.

janosSarusiKis:
Thanks for the info! With that, I gave it a new look and get to this point:

The client side already omits the status on data frames (first_msg_metadata / continuation_metadata); only send_response_stream was stamping one. Fixed that instead of adding a marker, so no new metadata key.

One tweak: is_eos keeps the empty-payload check alongside the status-presence check. Presence alone would mean a client on this version reads the first data frame from an older server as EOS, since those stamp Ok on every frame. With both clauses it degrades to current behaviour against old peers instead of truncating streams — no wire break.

Also had to adjust the demux filter in server.rs, which identified echoed server responses by status presence; it now keys off RPC_DIR_KEY directly.

I also updated PR title and description.

msardara (approved):
LGTM ✅

PR #1964, opened 2026-08-11
chore(config): use DurationString for timeout and jwks_ttl in OIDC config

msardara:
## Summary

- Change `timeout` and `jwks_ttl` fields in the OIDC `Config` struct from `Option<Duration>` to `Option<DurationString>` for consistent serde-friendly duration handling
- Update default functions and builder methods to convert with `.into()`
- Update `to_auth_config` and `create_verifier` to convert back to `Duration` at the call site
- Fix test assertions to compare against `DurationString` values

### DIFF
--- a/crates/config/src/auth/oidc.rs
+++ b/crates/config/src/auth/oidc.rs
@@ -4,6 +4,7 @@
 use std::path::PathBuf;
 use std::time::Duration;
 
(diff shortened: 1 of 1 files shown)


markpmarton (approved):
LGTM 👍

PR #1965, opened 2026-08-11
feat(slimctl): record login credentials in the connection config

msardara:
# Description

`slimctl login` stored tokens in `~/.slimctl/credentials.yaml` but left the connection config untouched, so authentication relied on an implicit credentials-file lookup inside `resolve_config`. Nothing on disk showed which identity a command would use.

Follow-up to #1960, which added the OIDC refresh machinery.

### `slimctl login` writes an explicit `auth` block

Merged into the config `--config` targets (default `~/.slimctl/config.yaml`), so `endpoint`, `tls`, timeouts and everything else are preserved. Secrets are never inlined — the block only names the file holding the credential:

```yaml
auth:
  type: static_jwt
  file: /home/user/.slimctl/token
```

### slimctl authenticates with the access token only

It deliberately does not consume the refresh token. A refresh token serves one process at a time and every exchange invalidates the stored copy, so a CLI spending a rotation per invocation would race any long-lived process seeded from the same token — and would gain nothing, since it exits long before the background renewal task could run.

When the access token expires the user re-runs `slimctl login`. `Unauthenticated`/`PermissionDenied` failures now carry that hint instead of a bare status code.

The refresh token is still written to `~/.slimctl/refresh_token` (mode 0600) for long-lived processes to consume.

### `oidc::Config`: file-backed refresh tokens

Two new optional fields, both absent from serialized output when unset, so existing configs are unaffected:

- **`refresh_token_file`** — a *read-write* binding. The token is read from the file and rotations are written back to the same path (mode 0600). The existing inline `refresh_token` cannot survive rotation by construction: `get_client_layer` passed `persist_credentials: None` on that branch, so an IdP issuing single-use refresh tokens invalidated the configured value after one exchange, breaking the next start.
- **`access_token_file`** — optional cache. A still-valid token found here seeds the provider, sparing short-lived processes a token-endpoint round-trip *and* a refresh-token rotation on every start.

```yaml
auth:
  type: oidc
  issuer_url: https://issuer.example.com
  client_id: myclient
  refresh_token_file: /var/lib/slim/refresh_token
  access_token_file: /var/lib/slim/access_token
```

Note that a refresh token serves one process at a time and nothing here enforces that — two processes sharing a file will fight and both eventually fail with `invalid_grant`. Documented on the field; point each long-lived consumer at its own file.

### `${env:...}` / `${file:...}` in slimctl configs

`ConfigResolver` is now wired into slimctl's config loader, matching the node, control-plane and channel-manager loaders.

This exposed a hazard worth calling out: read-modify-write paths (`login`, `config set`) must **not** resolve before serializing back, or a user's `${...}` reference gets replaced by its expansion — silently baking a secret read from a file into the YAML. Those paths use a separate non-resolving load, pinned by `load_config_for_update_preserves_references_verbatim`.

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Testing

30 new unit tests. Notable coverage:

- refresh-token file is written 0600, and neither token ever appears in the config
- `save_login_auth` preserves `endpoint`/`tls`/timeouts, creates nested paths, overwrites a prior `basic` block
- re-login leaves neither the old nor the new token in the config
- the written block survives `resolve_config` rather than being masked by the implicit fallback
- `refresh_token_file` takes precedence over a stale inline `refresh_token`; an unreadable `access_token_file` is only a cache and must not fail the build
- `${env:...}`/`${file:...}` resolution, and unresolvable references erroring

`cargo test -p agntcy-slim-config -p agntcy-slimctl`: 230 + 205 unit, 25 e2e, 5 tls, 0 failures. clippy and fmt clean on `--all-targets`.

## Not included

The checked-in `client-config.schema.json` is not regenerated. It is already ~370 lines stale on `main` (missing `connection_type`, `require_header_mac`, the `oidc` auth variant, `policy`), it is generated manually rather than in CI, and `link_id`'s default is a random UUID so every regeneration produces churn. Worth a dedicated PR.

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -178,6 +178,7 @@ dependencies = [
  "serde_json",
  "serde_yaml",
  "sha1",
+ "tempfile",
  "thiserror 2.0.19",
  "tokio",
  "tokio-retry",
--- a/crates/auth/src/refresh_token.rs
+++ b/crates/auth/src/refresh_token.rs
@@ -32,6 +32,11 @@ pub struct RefreshTokenProviderConfig {
     /// Called after a successful token exchange when the IdP issues a new
     /// refresh token (rotation). Arguments: (new_access_token, new_refresh_token).
     /// Use this to persist the rotated tokens so process restarts don't lose them.
+    ///
+    /// Runs on the blocking pool and is awaited before the exchange completes, so
+    /// the rotated token is on disk before the new access token is handed out.
+    /// Implementations may block (file I/O) but should not be slow: they sit in
+    /// the path of every renewal.
     pub persist_credentials: Option<Arc<dyn Fn(String, String) + Send + Sync>>,
 }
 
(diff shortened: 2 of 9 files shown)


markpmarton (approved):
LGTM 👍, thanks!

PR #1966, opened 2026-08-11
feat(control-plane): support multiple northbound and southbound listeners

msardara:
# Description

The control plane bound exactly one address per API surface, so a deployment could not expose the same API on several addresses — e.g. plaintext on loopback for local tooling alongside mTLS on a routable address, or one listener per network interface. Working around it meant running a second control-plane instance, which does not share state.

`northbound` and `southbound` now each accept **either** a single server mapping **or** a list of them, and the Helm chart exposes one Service port per listener.

## Control plane

```yaml
northbound:
  - endpoint: "127.0.0.1:50051"    # local tooling, plaintext
    tls:
      insecure: true
  - endpoint: "0.0.0.0:50451"      # routable, TLS
    tls:
      insecure: false
      source:
        type: file
        cert: /etc/slim/tls.crt
        key: /etc/slim/tls.key
```

Every listener serves the same service backed by the same state, so a node registering through one southbound listener is immediately visible from every northbound listener. The single-mapping form is untouched, so **existing configs keep working** — `ServerConfigs` deserializes from either a map or a sequence. Omitting a key still yields the default single listener on `0.0.0.0:50051` / `0.0.0.0:50052`.

### Failure modes made explicit

- **An empty list is rejected at deserialization.** Silently binding nothing would leave the API unreachable with no diagnostic; the error points at removing the key to get the default instead.
- **A bind failure names the offending endpoint.** With several listeners configured, a duplicated or already-occupied address is otherwise hard to attribute.

### Implementation notes

- `NorthboundApiService` / `SouthboundApiService` derive `Clone` — one instance per listener. Every field was already a handle to shared state (`SharedDb`, `RouteService`, `DomainAuthenticator`, `SharedDrain`) and already cloned when the services were constructed, so this adds no new sharing semantics.
- `Config { northbound: ServerConfig::.. }` call sites now need `.into()`; applied to the six in the integration tests. This is the only source-level break, and it is confined to test setup.

## Helm chart

`service.north` and `service.south` become lists of ports to expose — one entry per listener — in the same shape the slim data-plane chart already uses for `service.data` / `service.control`. The Service and container ports range over them, and `values.yaml` templates the port into each config entry, exactly as the data-plane chart does with `endpoint: '0.0.0.0:{{ (index .Values.slim.service.data 0).port }}'`:

```yaml
service:
  north:
    - port: 50051
      name: north
    - port: 50451
      name: north-tls

config:
  northbound:
    - endpoint: "0.0.0.0:{{ (index .Values.service.north 0).port }}"
      tls:
        insecure: true
    - endpoint: "0.0.0.0:{{ (index .Values.service.north 1).port }}"
      tls:
        insecure: false
        source:
          type: spire
          socket_path: "unix:///run/spire/agent-sockets/api.sock"
```

`name` is optional and defaults to `<bound>-{index}`. The Ingress backends and `NOTES.txt` use `(index .Values.service.<bound> 0).port`, the same idiom as the slim chart's readiness probe. `_helpers.tpl` and `configmap.yaml` are untouched.

Adding a listener means one entry in `service.<bound>` and one in `config.<bound>`. An earlier revision of this branch derived the second from the first via template helpers; that was dropped in favour of matching the existing chart convention.

**This changes the values shape**: `service.north.port` becomes `service.north[0].port`, as in the data-plane chart.

### Exposing only some listeners

An Ingress publishes exactly one listener. `ingressNorth.servicePort` / `ingressSouth.servicePort` select it by port name or number; unset keeps the previous behaviour of taking the first, so existing values render unchanged.

```yaml
service:
  type: ClusterIP
  north:
    - port: 50051          # in-cluster only
      name: north
    - port: 50451          # published
      name: north-tls

ingressNorth:
  enabled: true
  servicePort: north-tls
```

With `ClusterIP` the unpublished listeners are reachable only inside the cluster, which is how one address stays public while another stays internal. Auto-generated names (`north-1`) work as selectors too, and a value matching no listener fails rendering with the available names listed rather than producing an Ingress that 503s.

Documented limitation: `LoadBalancer` and `NodePort` publish **every** Service port regardless of the Ingress, so the internal-only property depends on `ClusterIP`. Splitting listeners across trust boundaries with a LoadBalancer needs a second Service selecting the same pod, which this chart does not create.

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Testing

**`test_multiple_northbound_and_southbound_listeners`** — two northbound and two southbound listeners; one node registered through *each* southbound listener; both northbound listeners queried. Asserts both return identical node sets and that a node registered via southbound #2 is visible from northbound #1, i.e. the listeners share state rather than merely both being bound.

I verified this test has teeth by mutating `server.rs` to start only the first listener of each bound — it fails with `ConnectionRefused` on the second northbound port, and passes again once reverted.

**`test_duplicate_listener_endpoint_fails_to_start`** — two listeners on the same address must fail startup with the endpoint named, rather than silently binding only the first.

**12 config unit tests** covering: the single-mapping form still parsing to one listener; a sequence parsing to N; per-listener TLS staying independent rather than inherited from the first; empty-list and scalar forms erroring; mixed single/multiple across the two bounds in one file; omitted keys falling back to defaults; and the two exact YAML shapes the chart emits, so a drift on either side is caught here rather than at deploy time.

**Chart verification** — rendered the default and multi-listener shapes and parsed each resulting `config.yaml` with the Rust `Config` type, confirming the templated endpoints are what the binary accepts (`nb=["0.0.0.0:50051", "0.0.0.0:50451"]`). `helm lint` passes for both. The default `Service` output is unchanged from `main`.

`cargo test -p agntcy-slim-control-plane`: 254 unit + 33 integration, 0 failures. clippy and fmt clean on `--all-targets`.

## Docs

- `docs/content/slim/components/controller/config.md` gains a "Multiple Listeners" section covering the list form, the per-listener TLS use case, both failure modes, and the chart derivation.
- `charts/slim-control-plane/README.md` documents the new `service.*` fields, the derivation, and reworks the mTLS examples to set TLS/auth on the port rather than on `config`.
- `charts/slim-control-plane/values.yaml` documents both shapes inline.

## Incidental fix

`useSpiffe: true` appeared in a commented mTLS example in `values.yaml` on `main` and is not a field on `TlsServerConfig`. Because `TlsServerConfig` does not deny unknown fields, the key is silently dropped and the config resolves to TLS enabled with `source: None` — no certificate — instead of failing with a clear error. The third commit corrects it to the tagged `source: {type: spire, socket_path: …}` form throughout, and adds two tests pinning the documented TLS forms against the real deserializer so an invented field name can't survive review again.

## Checklist


msardara:
Note: integration test is currently failing as the image needs a version bump. It will be done after releasing the new image.

markpmarton (approved):
LGTM ✅

PR #1967, opened 2026-08-11
fix(auth): await the rotated-credential persist instead of detaching it

msardara:
# Description

`RefreshTokenProvider::fetch_new_token` updated `current_refresh_token` in memory and then dropped the `spawn_blocking` handle for `persist_credentials`, leaving the write detached from the exchange:

```rust
if let Some(new_rt) = resp["refresh_token"].as_str() {
    *self.current_refresh_token.write() = new_rt.to_owned();
    if let Some(cb) = self.config.persist_credentials.clone() {
        tokio::task::spawn_blocking(move || cb(at, rt));   // handle dropped
    }
}
```

Two consequences:

- **Ordering.** The function returned — and the caller started using the new access token — while the rotated refresh token might not have reached disk. A crash in that window leaves memory ahead of disk, so the next process start replays a token the IdP has already spent and fails with `invalid_grant`. That is exactly what `persist_credentials` exists to prevent.
- **Shutdown.** Dropping a `spawn_blocking` handle does not cancel the task, and `#[tokio::main]` waits for the blocking pool on runtime drop, so in practice short-lived processes did complete the write. But that relied on an incidental property of the shutdown path, and anything calling `std::process::exit` lost it.

## Changes

The handle is awaited, and a failed join is logged instead of ignored. `spawn_blocking` stays because the callback does synchronous file I/O (`std::fs::write` + `set_permissions` in `oidc.rs`) — running it inline would block an async worker. No lock is held across the new await point: the `current_refresh_token` write guard above it is a statement-level temporary.

An async callback would also have worked:

```rust
Option<Arc<dyn Fn(String, String) -> BoxFuture<'static, ()> + Send + Sync>>
```

at the cost of a boxed future every caller has to produce. Nothing needs to persist over the network today, so the sync callback is kept and the doc comment now states that implementations run on the blocking pool, are awaited, and sit in the path of every renewal.

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Testing

Four tests against a wiremock IdP whose token endpoint rotates the refresh token on every exchange:

- **`persist_completes_before_fetch_returns`** — the callback sleeps 150 ms, as a real file write would; asserts the completion flag is set once `fetch_new_token` resolves, and that the callback received the new access *and* refresh token.
- **`persisted_token_matches_in_memory_token`** — the persisted value and `current_refresh_token` must agree, which is the whole point of ordering the write.
- **`rotation_without_persist_callback_is_fine`** — no callback configured is not an error; rotation still updates memory and the access token is served.
- **`no_rotation_means_no_persist`** — a response without a `refresh_token` must not invoke the callback; there is nothing new to persist.

I verified the first two actually have teeth by reverting the change to the detached form: both fail (`fetch_new_token returned before the rotated token was persisted`), and pass again once restored.

`cargo test -p agntcy-slim-auth --lib`: 166 passed, 0 failed. clippy and fmt clean on `--all-targets`.

Note: `cargo test -p agntcy-slim-auth` also reports 8 failing doctests in `auth_provider.rs`, `jwt.rs`, `resolver.rs` and `spire.rs`. These fail identically on clean `main` (verified by stashing this change) and are unrelated — none are in files touched here.

### DIFF
--- a/crates/auth/src/refresh_token.rs
+++ b/crates/auth/src/refresh_token.rs
@@ -32,6 +32,11 @@ pub struct RefreshTokenProviderConfig {
     /// Called after a successful token exchange when the IdP issues a new
     /// refresh token (rotation). Arguments: (new_access_token, new_refresh_token).
     /// Use this to persist the rotated tokens so process restarts don't lose them.
+    ///
+    /// Runs on the blocking pool and is awaited before the exchange completes, so
+    /// the rotated token is on disk before the new access token is handed out.
+    /// Implementations may block (file I/O) but should not be slow: they sit in
+    /// the path of every renewal.
     pub persist_credentials: Option<Arc<dyn Fn(String, String) + Send + Sync>>,
 }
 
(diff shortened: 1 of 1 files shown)


markpmarton (approved):
LGTM 👍 Thank you Mauro for fixing this bug!

PR #1971, opened 2026-08-12
fix(auth): serialize concurrent refresh-token exchanges with file lock

markpmarton:
# Description

Two processes sharing the same `refresh_token_file` could both read the same refresh token and send it to the IdP concurrently. The first succeeds and rotates the token; the second gets `invalid_grant` and its background renewal loop exits permanently.

# Changes
- Add `lock_and_reload` callback to `RefreshTokenProviderConfig` that acquires an exclusive lock and returns the current tokens from the backing store
- In `fetch_new_token`, acquire the lock before the IdP call and hold it until `persist_credentials` completes
- Skip the IdP call if the access token from disk is still fresh (within the first 2/3 of its lifetime), consistent with the background task's refresh threshold
- Add `extract_exp_iat_claims_unsafe` to the JWT module for the freshness check
- Wire up the `lock_and_reload` closure in the `refresh_token_file` branch of `get_client_layer`, using a companion `.lock` file as the locking target so `write_secret_file` can open the token file freely without a re-entrant lock
- Use `fs2::FileExt::lock_exclusive` for cross-platform file locking without unsafe
- Add a test verifying that concurrent callers of the lock closure are serialized and read distinct tokens

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -155,6 +155,7 @@ dependencies = [
  "drain",
  "duration-string",
  "fastwebsockets",
+ "fs2",
  "futures",
  "getrandom 0.4.3",
  "gloo-net",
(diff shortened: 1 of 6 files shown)


msardara (approved):
LGTM ✅

PR #1983, opened 2026-08-13
fix: propagate OIDC auth through merge and diff_connections

msardara:
## Summary

- `RequiredAuthMethod` lacked an `Oidc` variant, causing OIDC client configs to be silently folded into `Jwt` in `from_client_config`, with no matching arm in `merge_server_requirements`
- `diff_connections` had no access to `controller.clients` credentials as a fallback when no `outbound_clients` entry matched the CP-assigned endpoint
- `RequiredAuthMethod::None` was wiping local auth (`self.auth = None`), so any locally configured OIDC credentials were destroyed when the CP sent `None` for the auth method — which it does whenever it doesn't model the auth explicitly
- The proto `AuthMethod` enum and `model::AuthMethod` had no `Oidc` variant, so the CP could never signal OIDC to nodes

## Changes

**`crates/config/src/client.rs`**
- Add `RequiredAuthMethod::Oidc`; fix `from_client_config` to map `AuthenticationConfig::Oidc` → `RequiredAuthMethod::Oidc` (was `Jwt`)
- Add `RequiredAuthMethod::Oidc => {}` arm to `merge_server_requirements` (preserves local auth, matching `Jwt` behaviour)
- Change `RequiredAuthMethod::None` arm from `self.auth = None` to a no-op — `None` means "no auth required", not "force no auth", so locally configured credentials survive

**`crates/controller/src/service.rs`**
- Store `controller.clients` on `ControllerServiceInternal` for auth fallback
- `diff_connections` inherits auth from the first non-`None` `clients` entry when no `outbound_clients` entry matches the CP-assigned endpoint
- Credential guard updated to cover `Oidc` and check post-merge `client_config.auth` instead of `outbound_clients` membership

**`crates/control-plane/`**
- Add `AUTH_METHOD_OIDC = 4` to the controller proto enum and regenerate bindings
- Add `model::AuthMethod::Oidc` with full southbound (proto→model) and northbound (model→proto) round-trip
- Wire `model::AuthMethod::Oidc → RequiredAuthMethod::Oidc` in `generate_config_data`

## Test plan

- [ ] Build `agntcy-slim-config`, `agntcy-slim-controller`, and `agntcy-slim-control-plane` with no errors
- [ ] All 261+ `agntcy-slim-config` tests pass
- [ ] Node configured with OIDC in `controller.clients` can establish CP-managed outbound links without a duplicate `outbound_clients` entry
- [ ] Node with explicit `outbound_clients` entries continues to use those credentials (fallback not triggered)
- [ ] Node with `auth_method: None` on a link no longer has its local OIDC credentials wiped
- [ ] CP registered with `auth_method: oidc` correctly sends `RequiredAuthMethod::Oidc` to nodes

### DIFF
--- a/crates/config/src/client.rs
+++ b/crates/config/src/client.rs
@@ -496,6 +496,7 @@ impl ClientConfig {
             }
             RequiredAuthMethod::Basic => {}
             RequiredAuthMethod::Jwt => {}
+            RequiredAuthMethod::Oidc => {}
         }
         if let Some(ms) = server.timeout {
             self.connect_timeout = Duration::from_millis(ms as u64).into();
@@ -632,6 +633,7 @@ pub enum RequiredAuthMethod {
     None,
     Basic,
     Jwt,
+    Oidc,
     #[cfg(not(target_family = "windows"))]
     Spire {
         trust_domain: Option<String>,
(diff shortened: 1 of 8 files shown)


markpmarton (approved):
LGTM 👍

PR #1986, opened 2026-08-13
fix: prevent unbounded span nesting on reconnect cycles

msardara:
## Summary

- On every reconnect cycle, the span hierarchy grew by two levels: `process_stream → reconnect → process_stream → reconnect → …`
- `#[tracing::instrument]` on `reconnect` creates its span **synchronously when the function is called**, making it a child of the current `process_stream` span before `.instrument(Span::none())` at the call site could have any effect
- The new `process_stream` spawned inside `reconnect` then inherited that span as its parent, and its own reconnect loop continued the chain

## Fix

Remove `#[tracing::instrument]` from `reconnect` and create the span manually **inside the async body**. Code inside `async fn` runs when the future is first polled, not when the function is called. The call site wraps the future with `.instrument(Span::none())`, so `Span::none()` is the active context at poll time — any span created inside the body becomes a root span.

Result: every reconnect cycle starts a fresh `reconnect` root span. Depth is capped at `reconnect → process_stream` regardless of how many cycles occur.

## Before / After

**Before** (after 11 reconnects):
```
process_stream:reconnect:process_stream:reconnect:process_stream:reconnect:…:reconnect: connection re-established
```

**After** (any number of reconnects):
```
reconnect{conn_index=0}: connection re-established
```

## Test plan

- [ ] Build `agntcy-slim-datapath` with no errors
- [ ] Run a node against a peer that repeatedly drops — confirm log spans stay flat (`reconnect{…}: …`) and do not accumulate nesting

### DIFF
--- a/crates/datapath/src/message_processing.rs
+++ b/crates/datapath/src/message_processing.rs
@@ -1346,71 +1346,85 @@ impl MessageProcessor {
         }
     }
 
-    #[tracing::instrument(skip_all, fields(service_id = %self.internal.service_id, conn_index))]
     async fn reconnect(
         &self,
         client_conf: ClientConfig,
         conn_index: u64,
         cancellation_token: &CancellationToken,
     ) -> bool {
-        info!("connection lost with remote endpoint, attempting to reconnect");
+        // The span is created here, inside the async body, so it is created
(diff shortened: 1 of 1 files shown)


markpmarton (approved):
LGTM ✅

PR #1989, opened 2026-08-13
fix: OIDC credentials for CP-managed links using dataplane.clients fallback

msardara:
## Summary

- Fix `from_server_config` so `AuthenticationConfig::Oidc` maps to `AuthMethod::Oidc` instead of falling through to `None`, ensuring the CP stores and forwards the correct auth method after node registration.
- Fix `merge_server_requirements` so `RequiredAuthMethod::None` from the CP no longer wipes local OIDC credentials (now a no-op).
- In `diff_connections`, after failing to find credentials in `outbound_clients` (exact endpoint match or empty-endpoint default), fall back to the node's own `dataplane.clients` by endpoint. This means a node that already authenticates to an endpoint via `dataplane.clients` (e.g. with OIDC) does not need to duplicate those credentials under `controller.outbound_clients` to satisfy CP-managed Remote links to the same endpoint.
- Normalize endpoints before comparison (`https://host` and `https://host:443` are treated as equivalent) so the fallback lookup matches even when the CP records an explicit port but the local config omits it.

## Test plan

- Build passes (`cargo build --bin slim`).
- All existing unit tests pass.
- Manually verified: CP-ordered Remote link to an endpoint that is already reachable via `dataplane.clients` with OIDC now inherits those credentials and connects successfully without duplicating config.

### DIFF
--- a/crates/controller/src/config.rs
+++ b/crates/controller/src/config.rs
@@ -129,6 +129,11 @@ impl Config {
         // Used to extract connection type information required to connect to the node
         // (e.g., TLS settings). This information is used by the control plane.
         dataplane_servers: &[ServerConfig],
+        // List of client configurations for the dataplane services.
+        // Used as a credential fallback for CP-managed outbound links when no
+        // matching outbound_clients entry exists, so the client does not need to
+        // duplicate credentials it already has for a given endpoint.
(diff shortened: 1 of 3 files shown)


janosSarusiKis (approved):
lgtm

PR #1995, opened 2026-08-14
fix(session): accept multicast join requests in unreliable mode

### LINKED ISSUE #1988: Multicast join fails in default unreliable mode
**Version**

slimctl 2.2.0 on Linux x86_64

**Description**

slimctl bench channel fails in the default mode.

The subscriber receives the join request, but rejects it with:

missing payload: timer options

Looking at the request, timer_settings is None.

The publisher then retries and eventually fails with:

message send retries exhausted

If I run the same test with --reliable, the subscriber joins correctly and messages are exchanged.

So it looks like multicast join requires timer settings, while the default unreliable mode doesn't provide them.

markpmarton:
fixes #1988 

# Description

The `timer_settings` guard in `handle_join_request` was introduced when timer fields were mandatory. After unreliable mode was added (making them `Option`), `from_join_request` already handled `None` correctly, but the guard was never removed — causing the subscriber to reject every join attempt that didn't use `--reliable`.

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/crates/session/src/session_layer.rs
+++ b/crates/session/src/session_layer.rs
@@ -1090,7 +1090,9 @@ mod tests {
     use crate::test_utils::{MockTokenProvider, MockVerifier};
     use slim_auth::shared_secret::SharedSecret;
     use slim_datapath::Status;
-    use slim_datapath::api::{CommandPayload, NameId, ProtoName, ProtoSessionMessageType, ProtoSessionType};
+    use slim_datapath::api::{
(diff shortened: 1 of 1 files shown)


muscariello (approved):
LGTM

PR #2000, opened 2026-08-19
feat(auth): bind OIDC identity to the MLS signing key via DPoP

markpmarton:
# Description

Adds an OIDC credential method where the SLIM identity is the **person** in your SSOprovider, not the application instance — so every app a user runs present the same`sub`, instead of each getting an unrelated identity.

SLIM needs the app's MLS signing public key inside the identity token, and an external IdP won't mint a SLIM-specific claim. Instead of requiring a custom protocol mapper per deployment, this uses **DPoP (RFC 9449)**, which providers already implement:
 `slimctl login --dpop` generates an MLS key pair, proofs the authorization-code exchange with it, and the IdP returns a token carrying `cnf.jkt` —the RFC 7638 thumbprint of that key. The app installs the key as its MLS signing key; a receiving peer validates the token against JWKS, re-hashes the presented key, and requires it to equal `cnf.jkt`.

A stolen token replayed under a different key fails. After the initial binding there is no per-message DPoP overhead — MLS leaf-node signatures, made with the same key, are the continuing proof of possession. 

Renewal re-proves the same key, so the binding survives token rotation without disturbing group membership. Requires an RFC 9449 provider (Keycloak 24+) and a signing key with a JOSE mapping:P-256 (`ES256`) or Ed25519 (`EdDSA`). 

`enforce_pqc` needs the `curve25519` build feature — `ML_KEM_768_X25519` is post-quantum in its KEM only and still signs with Ed25519.

## Changes

- **New `crates/auth/src/dpop.rs`** — RFC 9449 proof construction and RFC 7638  thumbprints over MLS keys. Free of `jsonwebtoken`/`reqwest` so it compiles for  `wasm32`, where `validate_member` still needs thumbprints.
- **Credential format carries the presented key** as `<access token>~<b64url pubkey>`  in the existing opaque `SLIMHeader.identity` field. No new proto field: `prost`  drops unknown fields and relays re-encode forwarded messages, so a new field would  be silently stripped by any node on an older build. A credential with no `~` is a  plain bearer token on the unchanged path.
- **Verification stays DPoP-unaware downstream** — `OidcVerifier` checks the  thumbprint and then surfaces the key as an ordinary `pubkey` claim, so  `IdentityClaims`, MLS `validate_member`, and the e2e header-signature check are  untouched.
- **`OidcTokenProvider` becomes an identity provider**, not just a transport one:  implements `get_signature_keys`/`set_signature_keys`, gains  `exchange_authorization_code`, `install_signature_keys`, `adopt_refresh_token`.
- **Renewal delegated to `RefreshTokenProvider`** for user identities, which owns the  key, schedule, persistence and cross-process rotation lock. Exactly one token cache  is live per mode — reading the wrong one would serve a service token under a user's  MLS key.
- **One shared token-endpoint POST helper** (`post_token_request_with_dpop`) across  the authorization-code grant, the refresh grant, and `slimctl login`, so the proof  cannot drift between the grant that mints the binding and the one that renews it.  Includes the RFC 9449 §8 nonce retry.
- **Identity uniqueness now salted in three places**, since one `sub` is shared by  many instances: the MLS leaf identity is `subject:public_key` (mls-rs requires it  unique per member; `valid_successor` still matches on subject so key rotation  remains a succession), the app id hash includes the app's own name components, and  derived session data/control channel addresses include the app id.
- **Split provider constructors**: `create_identity_provider()` adopts the key and refresh token from `slimctl login`; `create_provider()` (transport auth) deliberately does not, or a daemon would authenticate as whoever last logged in on that host and load their private key. A configured `client_secret` also suppresses adoption, so a service identity can't be overridden by a stray personal login.
- **`AuthConfig::Oidc` / `IdentityProviderConfig::Oidc` / `IdentityVerifierConfig::Oidc`** plus the `slim-bindings` equivalents — selectable from YAML like any other variant.
- **`slimctl login --dpop`** generates the key, proofs the exchange, verifies `cnf.jkt` before saving, and stores the pair in `~/.slimctl/credentials.yaml` (0600) as optional fields — older files still parse. Adds `--allow-http` for a loopback test IdP.
- **JWKS keys prepared once per fetch** instead of re-parsed per verification, and held behind an `Arc` so a verify takes a refcount rather than a deep clone.
- **Token and JWKS caches reduced to a single slot each**, dropping the `HashMap`-keyed indirection that never held more than one entry.

## Bug fixes

- **Claim cache outlived the token.** A cache hit short-circuits verification, so an  entry with the plain TTL kept an *expired* token verifying until the TTL ran out.  Now clamped to the token's own `exp`.
- **`try_get_claims` always returned `WouldBlockOn`**, which failed every member of an  OIDC-backed MLS group — `validate_member` is sync, so this is the only path MLS has.  It now verifies against the cached JWKS, as `try_verify` does.
- **Background refresh stuck forever after a failed initial fetch.** The initial fetch  deliberately doesn't fail startup, but with no cache entry the loop had no  `refresh_at` to reach and never retried.
- **JWKS read lock held across verification**, so a refresh writer blocked every  in-flight signature check. The lock is now released before verifying.
- **Identity verified twice per message** on the e2e path (`try_verify`/`verify`  followed by the claims fetch). Now once.
- **`invalid_grant` misreported as a revoked refresh token** on the  authorization-code grant, where it actually means the code was stale or replayed —  sending the caller down the wrong recovery path.
- **OIDC config required a `client_secret`**, rejecting the public  authorization-code + PKCE clients this flow depends on. Now optional;  `AuthOidcEmptyClientSecret` → `AuthOidcEmptyIssuerUrl`, with an explicit failure  when there is no working renewal path at all (no secret and no usable stored login,  e.g. the IdP granted no `offline_access`) instead of failing at the first token  fetch behind a swallowed warning.
- **Empty `client_secret` sent to the token endpoint** made Keycloak reject the  request as malformed; the field is now omitted for public clients.
- **Ciphersuite/key mismatch failed deep inside a group operation.**  `Mls::build_client` and `create_update_proposal` now derive the public key from the  installed secret under the active ciphersuite and reject a mismatch at startup, with  a message naming the `curve25519`/`enforce_pqc` cause.
- **An MLS key rotation could silently break an OIDC identity** — `cnf.jkt` can't be  re-bound locally. `install_signature_keys` now refuses a key whose thumbprint  doesn't match the live token.
- **Unverified `pubkey` claims were adoptable** from a token or a userinfo merge.  They're now stripped from both places `IdentityClaims::from_json` looks, so a  `pubkey` can only ever come from a verified `cnf.jkt`.
- `h2` bumped to 0.4.16 (unrelated to the feature — split it out if you prefer clean  scope).

## Type of Change

- [x] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


Tehsmash (dismissed):

muscariello (changes requested):
I am not sure why we have this feature

muscariello:
moving this to AgentBridge https://github.com/agntcy/shadi

PR #2001, opened 2026-08-23
fix: update SPIRE to 1.15.3

muscariello:
Trivy reported 26 findings against the pinned 1.15.2 images, most of what the
code-scanning page shows. Scanning 1.15.3 locally:

| image | 1.15.2 | 1.15.3 |
|---|---|---|
| spire-agent | 14 | 2 |
| spire-server | 12 | 0 |

The 2 left are `golang.org/x/mod` v0.38.0 needing 0.40.0 — upstream's to fix.

Bumps the charts as well as the scan matrix, so the deployed version moves too
rather than only the version being scanned.

**Second commit:** 1.15.3 renamed the agent startup log from `Starting Workload
and SDS APIs` to `Starting agent APIs`
([spiffe/spire@v1.15.2...v1.15.3](https://github.com/spiffe/spire/compare/v1.15.2...v1.15.3)
`pkg/agent/endpoints/endpoints.go`), so the test harness timed out on a line the
agent no longer emits. Verified in the shipped binaries: 1.15.2 has only the old
string, 1.15.3 only the new. The matcher now accepts either.

### DIFF
--- a/.github/workflows/container-security-scan.yml
+++ b/.github/workflows/container-security-scan.yml
@@ -33,10 +33,10 @@ jobs:
             version: latest
           - image: spire-server
             repo: ghcr.io/spiffe/spire-server
-            version: 1.15.2
+            version: 1.15.3
           - image: spire-agent
             repo: ghcr.io/spiffe/spire-agent
-            version: 1.15.2
+            version: 1.15.3
     steps:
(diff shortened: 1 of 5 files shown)


msardara (approved):
LGTM ✅

PR #2005, opened 2026-08-25
chore: reusing slim sdk docs

### LINKED ISSUE #1892: chore(docs): document .NET bindings
Add documentation on the .NET SDK to the [SLIM bindings section](https://slim.agntcy.org/latest/slim/slim-howto/#slim-bindings).

keraron:
# Description

Fixes #1892

Updates the SLIM SDK docs in place: adds standalone per-language guides (Python, Go, .NET, Java, Kotlin, Node.js, React Native), refreshes the SLIMRPC compiler page with the Kotlin and Node.js plugins, and corrects package versions and toolchain requirements. The docs stay in this repo — the earlier plan to source them from agntcy/slim-bindings is dropped.

Removing the docs from slim-bindings: https://github.com/agntcy/slim-bindings/pull/86

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [x] Documentation
- [ ] Other (please describe)

## Checklist


muscariello (dismissed):
LGTM

Tehsmash (changes requested):
I have a couple issues with this PR. The original docs pre-2.0.0 did include the docs like this but there are a few issues with that so I removed it and inlined the docs.

First this change makes the docs track SDK documentation from slim-bindings main. This results in unreleased doc changes going out.

Related to this structure of the docs is defined in this repo so if we want to change that we have an issue because we can't merge changes to main docs in slim-bindings to move to the new structure without breaking the existing docs.

I would ask what we're gaining from this change, we're introducing a weird dependency cycle, today the docs are monorepoed here but they are versioned/released separately.

If we want to do this "split" properly then I think we need to use either split the docs out of SLIM into a separate slim-docs repo to make the separate versioning obvious.
 OR
We should use GitHub actions to have slim-bindings publish it's pages into the slim/gh-pages branch itself.

muscariello:
@janosSarusiKis @Tehsmash should we just move back all slim related doc in here?
This requires two issues to fix this.

janosSarusiKis (approved):

PR #2045, opened 2026-09-08
rpc: add shared-responses multicast mode

Tehsmash:
## Description

Adds shared-responses multicast mode to SlimRPC. In a GROUP session each server
normally handles the client's request in isolation. Shared-responses mode lets
every server also receive the response payloads of its peers, enabling
coordination patterns (aggregation, quorum, fan-in) without an external
message bus.

**Design** — per-call opt-in via metadata:

- Pass `SHARED_RESPONSES_KEY = "true"` in the per-call `metadata` argument
  to enable shared-responses for a single RPC call on any GROUP channel.
  Plain multicast and shared-responses calls can mix freely on the same channel.
- `make_shared_responses_metadata() -> Metadata` UniFFI helper builds the map.
- Each server handler receives a `PeerResponseReceiver` / `PeerResponseStream`
  (UniFFI) that delivers `PeerMessage { source, payload }` frames from peers.
- Peer count is derived at call-dispatch time from `SessionController::participants_list()`
  (excludes self and the request source/client), so no client-side member count needed.
- `Channel::local_name()` accessor added for UniFFI callers that need the app name.

**Breaking changes vs prior draft:**

- Removed `Channel::new_group_shared` / `new_group_shared_with_connection`
- Removed `Server::new_with_shared_responses` / `new_with_shared_responses_and_connection`
- Removed `SHARED_RESPONSES_MEMBER_COUNT_KEY` constant
- `UniffiPeerResponseStream` renamed to `PeerResponseStream` (bindings-visible name)

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [x] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/crates/rpc/src/channel.rs
+++ b/crates/rpc/src/channel.rs
@@ -61,7 +61,9 @@ pub struct MulticastItem<T> {
 
 use super::{
     Context, METHOD_KEY, Metadata, RPC_DIR_KEY, RPC_DIR_REQ, RPC_ID_KEY, ReceivedMessage, RpcCode,
-    RpcError, SERVICE_KEY, STATUS_CODE_KEY, calculate_timeout_duration,
+    RpcError, SERVICE_KEY, SHARED_RESPONSES_ENABLED, SHARED_RESPONSES_KEY,
+    SHARED_RESPONSES_MEMBER_COUNT_KEY, STATUS_CODE_KEY,
+    calculate_timeout_duration,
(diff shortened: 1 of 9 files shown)


muscariello:
This needs an issue describing the problem and the solution before a PR is sent @Tehsmash.  This approach can have security implication like congestion collapse.

PR #2072, opened 2026-09-17
fix(ci): do not sign or overwrite releases from unverified tags

muscariello:
`release-slimctl.yaml` fires on **any** pushed `slimctl-v*` tag, from **any** commit, by anyone with write access — and it signs what it builds with the project's Sigstore identity and attaches GitHub build provenance. Signing an unreviewed commit is worse than not signing it: the signature lends it authority it hasn't earned.

There is no tag ruleset on this repo, so nothing currently prevents that.

## Refuse unverified tags

A `verify-tag` gate now runs before anything is built, and fails unless the tagged commit is already an ancestor of `main`. Compatible with the real flow — release-plz tags the commit it just pushed to `main`.

This is a backstop, not the authorization control; restricting *who* may create `slimctl-v*` tags belongs in a tag ruleset, which is worth adding separately.

## Never overwrite a published release

The release is created as a **draft** and published only once all eight platforms have uploaded a signed archive.

This matters because there are two `--clobber` paths, not one — `upload-rust-binary-action` uses it internally (`main.sh:637`), not just the upload step added in #2065. So a re-pushed tag could silently replace bytes that were already signed and published.

Draft-first gets both properties: a failed run stays retryable on the same tag (reusing a draft is safe), while a published release is refused outright. Previously `gh release create` failed on any existing release, so a failed run left the tag with no binaries, signatures or provenance and needed manual cleanup — `slimctl-v2.1.0` failed that way.

`homebrew` now runs after `publish-release`, so `gh release download` and the cask's URLs read a published release rather than a draft.

## Job graph

```
verify-tag → ensure-release (draft) → build-binaries (×8, sign + attest) → publish-release → homebrew
```

## Verification

`ensure-release` logic exercised against a stubbed `gh`:

| Case | Exit | Behaviour |
|---|---|---|
| no release | 0 | creates draft |
| draft exists | 0 | reuses it |
| **published exists** | **1** | refuses to overwrite |
| draft appeared mid-run | 0 | reuses it |
| **published appeared mid-run** | **1** | refuses to overwrite |
| `gh` genuinely broken | 1 | still fails — the guard masks nothing |

`actionlint` clean, `zizmor` no new findings, `shellcheck` clean on all three added scripts. Confirmed `build-binaries` and `homebrew` steps are byte-identical to `main` — this PR only adds jobs and repoints `needs`.


muscariello:
Closing: over-scoped. The goal is just to exercise the signing path.

PR #2080, opened 2026-09-18
feat(datapath): implement default gateway as a fallback when route is…

### LINKED ISSUE #1968: Default gateway fallback: make set_route optional when inviting a participant
## Problem

Inviting a participant to a group today requires the app to install a route to that participant *first*, by hand:

```rust
// docs/content/slim/components/sdk/tutorials/tutorial-session.md:405
app.set_route(&participant_name, conn_id).await?;
session.invite_participant(&participant_name).await?;
```

The reason is structural: `invite_participant` (`crates/session/src/session_controller.rs:812`) is just a publish of a `DiscoveryRequest` addressed to the invitee's name (`invite_participant_internal`, `session_controller.rs:804`). That publish goes through the normal subscription-table match (`crates/datapath/src/message_processing.rs:779`), and with no route installed the lookup fails with `DataPathError::NoMatchEncoded` (`crates/datapath/src/tables/subscription_table.rs:713`). The error is pushed back to the app (`message_processing.rs:1313`) and surfaces as `SessionError::ModeratorTaskAddFailed` on the invite's completion handle (`crates/session/src/session_moderator.rs:592`). The moderator only installs its own routes *after* the `DiscoveryReply` (`session_moderator.rs:855`), so there is no way for it to recover on its own.

This is a bad API for three reasons:

1. **It leaks datapath topology into application code.** An app has to know a `connection_id` — an opaque `u64` handed out by the local connection table — to invite a participant by name.
2. **It is racy.** `set_route` has to wait for a datapath ACK precisely so a following publish can't overtake it (`crates/service/src/app.rs:416-424`), and our own examples still paper over it with a sleep: `crates/examples/src/chat/main.rs:230` — *"Give the routes a moment to propagate"*.
3. **It is pure boilerplate that we repeat everywhere.** Every caller pairs the two calls: `crates/channel-manager/src/service.rs:203`, `crates/rpc/src/channel.rs:368` and `:398`, `crates/slim-bindings/examples/sender.rs:166`, `crates/testing/src/bin/test_group.rs:187`, `crates/service/tests/session_restore_test.rs:212`. Forgetting it is a silent trap that only shows up as a failed invite.

## Proposal: a default gateway fallback

Add a **default gateway** to the datapath: when a publish finds no match in the subscription table, forward it to the default gateway instead of returning `NoMatchEncoded`.

**The gateway must be an `Edge` connection, and only an `Edge` connection** — i.e. the link between an App and a data plane node. That is exactly the case where the answer is unambiguous, and the code already guarantees the classification:

- `ClientConfig::connection_type` defaults to `ConnType::Edge` (`crates/config/src/client.rs:279` and `:309`, via `ConnType::default()` at `crates/config/src/conn_type.rs:31`). The field doc is explicit: *"Defaults to `edge`. Set to `peer` for intra-deployment peer connections, or `remote` for control-plane-managed inter-deployment links."*
- The app's uplink connection is constructed with precisely that value: `Connection::new(client_config.connection_type, outbound)` (`crates/datapath/src/message_processing.rs:529`).
- The bindings do not override it — `crates/slim-bindings/src/client_config.rs:349` passes the core default straight through, so every SDK app gets `Edge`.
- Both ends agree on the classification: the type is carried over the wire during link negotiation (`conn_type_to_link`, `crates/datapath/src/negotiation.rs:369`) and the node stamps its inbound side `Edge` (`resolve_connection`, `message_processing.rs:1600-1605`).

So an app-embedded SLIM instance has a connection table (`crates/datapath/src/tables/connection_table.rs:26`) containing N *local* connections (the app's own `register_local_connection`, `message_processing.rs:598`) plus exactly **one** `Edge` connection to its SLIM node. If there is exactly one `Edge` connection, that connection *is* the default gateway, and no `set_route` call is needed at all.

`Remote` and `Peer` connections are **excluded by design**, for concrete reasons:

- **`Remote`** is a control-plane-managed inter-deployment link (`conn_type.rs:24-26`). Routes across it are installed by the controller (`crates/controller/src/service.rs:1134`); inferring one here would fight the control plane rather than complement it.
- **`Peer`** is a same-deployment replica (`conn_type.rs:27-28`), and the datapath already applies deliberate loop-prevention to peer traffic via `MatchFilter::EXCLUDE_PEER` (`crates/datapath/src/tables.rs:37`, selected at `message_processing.rs:1215`). Silently defaulting to a peer would undercut that.

### Sketch

- **Resolution.** In `match_and_forward_msg`, in the `Err(e)` arm (`message_processing.rs:806`), attempt the default gateway before returning the error.
- **Selection.** Two sources, in priority order: (1) explicitly configured / set via API (e.g. `App::set_default_gateway(conn_id)` plus a config knob), for topologies where auto-detection is ambiguous; (2) auto-derived — scan the connection table for connections whose `connection_type()` is `ConnType::Edge`, and if exactly one exists, it is the gateway. `Local`, `Remote` and `Peer` are never candidates. Cache the resolved id and invalidate it from `Forwarder::on_connection_established` / `on_connection_drop` (`crates/datapath/src/forwarder.rs:45,55`) rather than scanning per message.
- **Explicit routes always win.** The fallback only fires on a no-match, so `set_route` keeps working unchanged and stays the escape hatch for multi-uplink topologies. This is purely additive.
- **Loop safety.** Only apply the fallback to messages that arrived on a *local* connection. Otherwise two nodes each defaulting to the other black-hole/loop the message (TTL, decremented at `message_processing.rs:1281`, would eventually catch it, but relying on that is wasteful). Restricting to local-origin is also exactly the case we care about: the app-side instance. On a real SLIM node the invitee is reachable either via its own subscription or via control-plane installed routes (`crates/controller/src/service.rs:1134`).

Note that the two conditions together — *gateway must be `Edge`* and *message must have arrived on a `Local` connection* — confine the feature precisely to the App↔node hop, which is the only place the inference is safe.

## Open questions

- **Zero `Edge` connections** → keep returning `NoMatchEncoded` as today.
- **More than one `Edge` connection** → ambiguous. Do *not* guess; keep the no-match error, but log clearly that a default gateway could not be inferred and that an explicit `set_route`/`set_default_gateway` is required.
- **Opt-in or on by default?** On by default gives the ergonomic win for free; opt-in is safer. Leaning on-by-default given it only triggers where the current behaviour is a hard error anyway.
- Should the fallback apply to *all* publishes, or only to session control messages such as `DiscoveryRequest`? Applying it to all publishes is simpler and more useful (a plain `publish` to an unrouted name works too), but it is a broader behavioural change.

## Acceptance criteria

- [ ] An app with a single connection to a SLIM node can call `session.invite_participant(name)` with no preceding `set_route` and the invite completes.
- [ ] Explicit routes still take precedence; `set_route` / `remove_route` semantics unchanged.
- [ ] No fallback for messages arriving on non-local connections; a test covers the two-node mutual-default case not looping.
- [ ] `Remote` and `Peer` connections are never selected as the gateway, with tests asserting a node whose only non-local connection is `Remote` (or `Peer`) still returns `NoMatchEncoded`.
- [ ] Ambiguous (>1 `Edge`) and empty (0 `Edge`) cases behave as documented above, with a clear log message.
- [ ] Gateway resolution is cached and invalidated on connection establish/drop, including reconnect.
- [ ] `set_route` calls dropped from examples and library callers where the fallback now covers them (`crates/examples/src/chat/main.rs` — including the 200 ms sleep, `crates/channel-manager/src/service.rs:203`, `crates/rpc/src/channel.rs:368,398`, `crates/slim-bindings/examples/sender.rs:166`).
- [ ] Docs updated: `docs/content/slim/components/sdk/tutorials/tutorial-session.md:396-507` and `docs/content/slim/components/sdk/tutorials/tutorial-app.md:223`.

hackeramitkumar:
## Summary

Implements default Edge gateway fallback for unrouted Local publishes.

## Details

- Automatically forwards Local publishes to the only available Edge connection.
- Keeps explicit routes as the highest-priority option.
- Supports explicit gateway pinning with `set_default_gateway`.
- Supports clearing the explicit gateway and disabling automatic gateway discovery.
- Rejects invalid or non-Edge gateway connections.
- Prevents fallback for Remote and Peer-originated publishes.
- Invalidates cached gateway state when connections are established, dropped, or change type.
- Adds service configuration support:

```yaml
default_gateway: auto # auto | off
```

- Adds App and bindings APIs for gateway management.
- Adds unit tests covering routing precedence, ambiguity, connection changes, explicit pins, and disabled auto-discovery.

## Validation

```bash
cargo fmt --all -- --check
cargo check -p agntcy-slim-datapath \
  -p agntcy-slim-service \
  -p agntcy-slim-bindings
cargo test -p agntcy-slim-datapath --lib gateway
```

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

Fixes : #1968


muscariello:
@hackeramitkumar Is this related to issue #1968 ?

hackeramitkumar:
@muscariello yes, it’s related to https://github.com/agntcy/slim/issues/1968.

muscariello:
Can you please connect the PR to the issue and comment if it covers all of it ?

hackeramitkumar:
@muscariello  Yes, this PR contains all the required changes, and I have tested it. Since the gateway is a fallback mechanism, older versions of slim-apps will also continue to work without any changes.

muscariello:
Thanks @hackeramitkumar — the datapath side of this is in good shape. The `Gateway` state machine is clear, the Edge-only and local-origin restrictions match what #1968 asked for, and the 13 new tests pass locally.

I'd like to **land the mechanism here and split the caller-facing work out**, rather than grow this PR. The fallback is additive and only fires where the code currently hard-errors, so it is low risk to merge; removing `set_route` from live callers is a behaviour change to working code and deserves its own review and its own revert. Follow-ups are filed so they do not get lost: #2089 (callers), #2090 (docs + config), #2091 (remaining tests).

## Please keep in this PR

**1. The end-to-end test.** Acceptance criterion 1 of #1968 is an app calling `invite_participant(name)` with no preceding `set_route` and the invite completing. Every test here is at the datapath level — a raw publish reaching an Edge connection. Nothing exercises `invite_participant` → `DiscoveryRequest` → no-match → gateway → `DiscoveryReply` → moderator installs routes. Without it we would be merging a mechanism whose actual purpose is unverified, and #2089 would have no safety net.

**2. The explicit pin bypasses the cache.** In `resolve()` the `explicit` branch calls `scan()` before the cache is consulted, so a pinned gateway does a full `connection_table` walk on *every* unrouted publish. #1968 specifically says to cache and invalidate "rather than scanning per message" — the auto path does, the explicit path does not.

**3. Four small cleanups while you are in the file**
- `Gateway::Dirty` can never be returned by `resolve()` (it always writes a concrete result), so the `Empty | Dirty` arm in `match_and_forward_msg` is dead and the variant leaks an internal sentinel into the public enum. Consider `Option<Gateway>` for the cache instead.
- `Forwarder::default_gateway_state` and the `DefaultGateway::cached` it wraps have no callers.
- The hand-written `Clone for DefaultGateway` is unreachable — `Forwarder` is not `Clone` and lives behind an `Arc`.
- `on_connection_established` invalidates for every connection type; only `Edge` can change the answer, so local churn needlessly dirties the cache.

## One decision for the maintainers, not for this PR to guess

`default_gateway: off` does **not** disable an explicit pin — `resolve()` checks `explicit` before `is_enabled()`, and `disabling_auto_still_allows_explicit_gateway` encodes that as intended. But the config enum documents `Off` as "The service will not use the default gateway". One of the two should change, and since this is a public config contract it is better settled before release than after. @msardara @micpapal — preference?

## Checked and fine, so nobody re-treads them

The fallback's `send_msg` matches what the normal single-destination path already does, so no TTL or header divergence. `MessageProcessor` is `Arc`-internal, so `create_app`'s `.clone()` correctly shares gateway state. Both `match_and_forward_msg` call sites were updated. `connection_table.for_each` is a lock-free `arc_swap` load, so holding the cache mutex across `scan()` is not a deadlock risk.

One note on `App::new`: it passes a fresh `MessageProcessor::new()`, so an `App` built that way has a `set_default_gateway` that targets a throwaway datapath. It is `pub(crate)` + `#[allow(dead_code)]` with only a test caller today so nothing is broken — but it is worth a comment, or taking the processor as a parameter there too, so a future caller does not get a silently non-functional API.

Given #2024 is also open, say if you would rather hand any of #2089–#2091 off — happy to pick them up.

muscariello (changes requested):
Thanks Amit, I would suggest split the PR.

hackeramitkumar:
@muscariello Sure, I will split the work in multiple PRs as per your suggestions. And i would like to work on all the follow up issues 🙂. If possible please assign them to me.

muscariello (approved):
LGTM

PR #2088, opened 2026-09-18
docs: add the OpenSSF Best Practices badge

muscariello:
Adds the OpenSSF Best Practices badge next to the Scorecard one, as in shadi.

Project [14428](https://www.bestpractices.dev/projects/14428) is slim's own entry (`repo_url: github.com/agntcy/slim`), currently at `passing` with tiered_percentage 104.


muscariello:
Temporarily based on `fix/gate-build-via-inputs` (#2092) so CI can run: on `main` this PR is blocked because a docs-only change skips the `Build & test` caller job, and a skipped `uses:` job never produces the `Build & test / *` contexts the ruleset requires. GitHub will retarget this to `main` automatically when #2092 merges.

lgecse (dismissed):

lgecse (approved):

Issue #2091, opened 2026-09-18
Add the remaining default-gateway tests: mutual-default loop and Remote/Peer-only table


muscariello:
Follow-up to #1968, after #2080. Two acceptance criteria of that issue have no test yet.

**Two-node mutual-default loop.** The issue asks for "a test covers the two-node mutual-default case not looping". #2080 tests that one node ignores a publish arriving on a Remote connection, but not the case where each node's gateway is the other. That loop-safety property is the reason the local-origin restriction exists and is what a future widening of the category check would break.

**Remote/Peer-only table.** The issue asks for tests "asserting a node whose only non-local connection is Remote (or Peer) still returns NoMatchEncoded". #2080 covers the explicit-pin rejection and the inbound-category check, but not auto-selection against a Remote/Peer-only table — so changing `edge_conn_ids`' filter to `!= Local` would pass every current test while silently making a Remote connection the gateway.

amitami2:
@muscariello,
Let's close this

PR #2096, opened 2026-09-19
test: cover default gateway fallback regressions

hackeramitkumar:
# Description

Adds the remaining regression coverage requested in [#2091](https://github.com/agntcy/slim/issues/2091) for the default gateway fallback introduced in [#2080](https://github.com/agntcy/slim/pull/2080).

## Changes

- Adds an end-to-end test using a real SLIM server where `invite_participant()` succeeds without `set_route()`.
- Adds coverage for two nodes with mutual Edge default gateways to ensure messages do not loop.
- Adds coverage confirming Remote and Peer-only connections are never selected as automatic gateways.
- Verifies unmatched publishes return `NoMatchEncoded` when no valid Edge gateway exists.

## Validation

- Focused datapath tests pass.
- End-to-end default gateway test passes.
- Formatting and Clippy checks pass.

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [x] Other (please describe)

## Checklist


muscariello (commented):
Covers both acceptance criteria from #2091. I mutation-tested rather than trusting the green checks: widening the fallback's category check to also accept `Edge` is caught by `mutual_edge_default_gateways_do_not_loop` and by nothing else in the suite — exactly the regression the issue was written to prevent. Flipping `edge_conn_ids` to `!= Local` is caught by the new Remote/Peer test as well.

Two things to fix before merge, both in `crates/testing/tests/default_gateway.rs` (left as applicable suggestions below):

- `let _ = timeout(..., invite_completion).await.expect(...)` discards the inner `Result<(), SessionError>` (`CompletionHandle::Output`), so a failed invite is silently ignored. With the fallback disabled, the test fails 15s later at the listener timeout instead of reporting the actual invite error.
- `let _ = timeout(..., participant_listener).await.expect(...)` discards the `JoinHandle`'s `JoinError`, so the task's `panic!("participant received error")` and `panic!("participant channel closed")` can never fail the test. I injected a bare `panic!()` as the first statement of the listener task and the test still passed.

On when this runs: `crates/testing/tests/*` are behind `required-features = ["integration"]`, which PR CI does not enable — but nightly's "Active integration tests" job runs `task tests:integration` (defined in `crates/testing/Taskfile.yml`) and picks up every such target, so this test is covered automatically there. Nightly is the right home for it; the suite takes 4-7 minutes on a cold runner and is too process-heavy to gate a merge queue on.

crates/testing/tests/default_gateway.rs
```diff
@@ -0,0 +1,131 @@
...
+        .expect("moderator session was dropped");
+
+    let invite_completion = session
+        .invite_participant(&participant_name)
+        .await
+        .expect("failed to invite participant");
+
+    let _ = tokio::time::timeout(Duration::from_secs(15), invite_completion)
+        .await
+        .expect("invite completion failed");
```
muscariello: `CompletionHandle::Output` is `Result<(), SessionError>`, so `let _ =` throws away the invite's own outcome. Chaining the second `expect` makes a failed invite fail here instead of 15s later at the listener: with the fallback disabled this now reports `invite failed: ModeratorTaskAddFailed { .. no matching found .. }` in 0.25s.

```suggestion
    tokio::time::timeout(Duration::from_secs(15), invite_completion)
        .await
        .expect("invite did not complete in time")
        .expect("invite failed");
```

crates/testing/tests/default_gateway.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    // Assert that the participant receives Notification::NewSession.
+    let _ = tokio::time::timeout(Duration::from_secs(15), participant_listener)
+        .await
+        .expect("participant listener timed out");
```
muscariello: `timeout` over a `JoinHandle` yields `Result<Result<(), JoinError>, Elapsed>`; `let _ =` drops the inner one, so the task's `panic!("participant received error")` and `panic!("participant channel closed")` can never fail the test. Verified: with a bare `panic!()` as the listener's first statement the test passes as written, and fails with this change.

```suggestion
    tokio::time::timeout(Duration::from_secs(15), participant_listener)
        .await
        .expect("participant listener timed out")
        .expect("participant listener panicked");
```

muscariello (commented):
Two applicable suggestions for the `let _ =` sites from my earlier review. Both verified locally: the test still passes with them, and each now catches a failure it previously swallowed.

muscariello:
Correction to my review above: I said the integration suite is not enabled anywhere. That is wrong — `crates/testing/Taskfile.yml` (`.yml`, which my grep of `Taskfile.yaml` missed) defines `tests:integration`, and nightly runs it via the "Active integration tests" job. It was green this morning.

So this test will run, just nightly rather than on PRs, and it is picked up automatically because the task runs every `--features integration` target. The two `let _ =` suggestions stand unchanged; only the "will not run in CI" sentence was wrong.

crates/testing/tests/default_gateway.rs
hackeramitkumar: done

crates/testing/tests/default_gateway.rs
hackeramitkumar: done

hackeramitkumar:
Thanks, @muscariello  , for reviewing the PR. I’ve addressed both review comments - please have another look when you get a chance.

artifacts-agntcy (approved):
LGTM

muscariello (approved):
LGTM

Issue #2089, opened 2026-09-18
Drop set_route boilerplate now the default gateway covers it


muscariello:
Follow-up to #1968, after the datapath fallback lands in #2080. Acceptance criterion 7 of that issue.

Every caller currently pairs `set_route` with the invite. Once the default gateway covers the single-uplink case, the boilerplate can go:

- `crates/examples/src/chat/main.rs:210` — and the `sleep(200ms)` at :214 with its "Give the routes a moment to propagate" comment, which only exists to paper over the race
- `crates/channel-manager/src/service.rs:204`
- `crates/rpc/src/channel.rs:371` and `:402`
- `crates/slim-bindings/examples/sender.rs:169`

This changes working code, so it wants the end-to-end test from #2080 as its safety net — do it after that lands, not before. Keep `set_route` where the topology has more than one uplink; the fallback deliberately does not guess there.

hackeramitkumar:
@muscariello 
Let's close this.

PR #2097, opened 2026-09-19
refactor: remove redundant caller routes for default gateway

hackeramitkumar:
# Description

Updates single-Edge callers to rely on the default gateway fallback introduced in [#2080](https://github.com/agntcy/slim/pull/2080).

## Changes

- Removes redundant `set_route()` calls from channel manager, examples, and testing binaries.
- Keeps explicit routing APIs and multi-route callers unchanged.
- Removes now-unused connection ID parameters and fields.
- Updates the channel-manager integration test.

Related to [#2089](https://github.com/agntcy/slim/issues/2089).

## Validation

- Formatting and Clippy pass.
- Affected packages compile successfully.
- Channel-manager integration test passes.

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


muscariello (commented):
Partial coverage of #2089: `crates/rpc/src/channel.rs:371` and `:402` still call `set_route`, and they are two of the four sites the issue names (acceptance criterion 7 of #1968). Both sit behind `if let Some(conn_id) = self.connection_id`, so leaving them is defensible — but please say so explicitly rather than omitting them silently. I have posted the verified patch as a separate comment if you want to finish it here.

The main concern is that no test running on this PR can observe the change. `crates/channel-manager/tests/integration_test.rs` does run at PR time, but it is not a guard: I re-ran it with the default gateway fallback fully disabled and it still passes, because `create_service_and_connect` builds a single `Service` and every app is a local connection on that same datapath, so the participant's subscription matches directly. The `set_route` removed from `service.rs` was already redundant in that topology. The tests that would catch a break run nightly only, so a regression surfaces the next morning — and with the merge queue it can land before anyone sees it.

The fix is small and does not need a new job or the `integration` feature. That file already starts a real SLIM node in-process via `start_slim_node`, and it runs in the normal `unittest` job. Giving the participant its own `Service` over the loopback makes the fallback load-bearing, so the test would fail if the fallback regressed — in seconds, with no external binaries.

Also: `ChannelManagerServer::new` loses its `conn_id` parameter. `agntcy-slim-channel-manager` is published on crates.io, so that is semver-breaking and the "Breaking Change" box is unchecked.

muscariello:
Here is the `crates/rpc/src/channel.rs` half of #2089, if you want to complete it in this PR. GitHub cannot offer it as a suggestion because the file is not in the diff, so it is a patch — `git apply` it:

```diff
@@ -367,13 +367,6 @@ impl Channel {
         // Invite all members whenever the GROUP session is (re)created.
         if self.is_group {
             for member in &self.initial_members {
-                if let Some(conn_id) = self.connection_id {
-                    self.app.set_route(member, conn_id).await.map_err(|e| {
-                        RpcError::internal(format!(
-                            "Failed to set route for group member {member}: {e}"
-                        ))
-                    })?;
-                }
                 send_invite(&session_tx, member).await?;
             }
         }
@@ -388,27 +381,10 @@ impl Channel {
     ) -> Result<(SessionTx, SessionRx), RpcError> {
         let app = self.app.clone();
         let remote = self.remote.clone();
-        let connection_id = self.connection_id;
         let runtime = &self.runtime;
 
         // Runs in a tokio task because create_session needs tokio runtime
         let handle = runtime.spawn(async move {
-            if let Some(conn_id) = connection_id {
-                tracing::debug!(
-                    remote = %remote,
-                    connection_id = conn_id,
-                    "Setting route before creating session"
-                );
-                if let Err(e) = app.set_route(&remote, conn_id).await {
-                    tracing::warn!(
-                        remote = %remote,
-                        connection_id = conn_id,
-                        error = %e,
-                        "Failed to set route"
-                    );
-                }
-            }
-
             tracing::debug!(remote = %remote, ?session_type, "Creating persistent session");
 
             let slim_config = slim_session::session_config::SessionConfig {
```

Removing the second site also orphans `let connection_id = self.connection_id;`, so that goes too. The `connection_id` field stays — it still has its public getter and constructor arguments. Verified with `cargo clippy -p agntcy-slim-rpc --all-targets`: clean, no warnings.

If you would rather leave these two sites alone, that is a reasonable call — just say so in the PR description so #2089 can be closed deliberately rather than looking half-finished.

muscariello:
Correction to my review above: I claimed the `integration` feature is never enabled in CI. That is wrong — nightly's "Active integration tests" job runs `task tests:integration`, defined in `crates/testing/Taskfile.yml`, and it is green.

The substance is unchanged but milder: the tests covering these call sites do run, nightly rather than at PR time, so a regression here surfaces the next morning instead of on the PR. The channel-manager integration test finding is unaffected — it runs on PRs and still passes with the fallback disabled, so it is not a guard for this change.

artifacts-agntcy (dismissed):
check comments

muscariello (changes requested):
Check comments.
@hackeramitkumar ignore the bot account which was open in my browser.

hackeramitkumar:
Thanks @muscariello, for the detailed review. I’ve addressed all requested points:
1. Removed the remaining set_route calls from crates/rpc/src/channel.rs.
2. Preserved conn_id in ChannelManagerServer::new for API compatibility and documented why it remains.
3. Added a PR-executed integration test using separate manager and participant services through the real add_participant gRPC path.
4. Verified the new test fails when the default-gateway fallback is disabled and passes when restored.

muscariello (commented):
One follow-up on the `conn_id` handling — applicable suggestion inline. Everything else from my review is addressed and I verified it independently, including that the new test fails with the fallback disabled while the pre-existing one still passes.

crates/channel-manager/src/service.rs
```diff
@@ -37,15 +36,19 @@ impl ChannelManagerServer {
...
+    ///
+    /// `conn_id` is retained for API compatibility. Routing is now resolved by
+    /// the default-gateway mechanism instead of being configured explicitly.
     pub fn new(
         app: Arc<App<AuthProvider, AuthVerifier>>,
         conn_id: u64,
         sessions: Arc<SessionsList>,
         config_mode: bool,
     ) -> Self {
+        let _ = conn_id;
```
muscariello: Keeping the parameter avoids the semver break, which is the right call — but discarding it silently changes behaviour for multi-uplink callers. The shipped binary is fine (`config.manager.slim_connection` is singular, so one Edge connection and auto-detection is unambiguous), but `ChannelManagerServer` is published, and a consumer with more than one Edge connection now gets `Gateway::Ambiguous` and a failed invite where their explicit `conn_id` used to work — with no signal, since the signature is unchanged.

Pinning it instead keeps the parameter meaningful and preserves the old behaviour exactly. `App::set_default_gateway` already exists (`crates/service/src/app.rs:269`).

I checked this against your new test: both tests still pass, and `test_add_participant_uses_default_gateway_with_separate_services` still *fails* when the fallback is disabled — so pinning does not weaken the guard you added.

```suggestion
    /// `conn_id` pins the default gateway, so invites keep working on a
    /// multi-uplink service where auto-detection would be ambiguous.
    pub fn new(
        app: Arc<App<AuthProvider, AuthVerifier>>,
        conn_id: u64,
        sessions: Arc<SessionsList>,
        config_mode: bool,
    ) -> Self {
        if let Err(e) = app.set_default_gateway(conn_id) {
            tracing::warn!("failed to pin default gateway to conn {conn_id}: {e}");
        }
```

hackeramitkumar: Thanks, Luca. I’ve updated ChannelManagerServer::new to pin the provided conn_id with set_default_gateway, preserving multi-uplink behavior while keeping the public API unchanged. The full channel-manager test suite and fallback-disabled mutation check pass.

crates/channel-manager/src/bin/channel_manager.rs
```diff
@@ -349,9 +338,7 @@ async fn main() -> anyhow::Result<()> {
     // API mode:    channels section is empty → DB (restored above) is the source of truth.
     let config_mode = !config.manager.channels.is_empty();
 
-    if config_mode
-        && let Err(e) = create_channels_from_config(&arc_app, conn_id, &sessions, &config).await
-    {
+    if config_mode && let Err(e) = create_channels_from_config(&arc_app, &sessions, &config).await {
```
muscariello, line 347: The pin lands in `ChannelManagerServer::new` on line 347, but config-mode reconciliation runs here on 341 and invites every configured participant before that. So in config mode the initial invites still rely on auto-detection alone — which is exactly the multi-uplink case the pin was added to cover. API-mode invites are fine, since the server exists by then.

Narrow (needs multi-uplink *and* config mode) and not a blocker, but it leaves the two paths inconsistent. `warn` is already imported on line 28; I checked this compiles clean and both channel-manager tests still pass.

```suggestion
    // Pin before reconciling: config-mode invites run here, before
    // `ChannelManagerServer::new` below does the pinning.
    if let Err(e) = arc_app.set_default_gateway(conn_id) {
        warn!("failed to pin default gateway to conn {conn_id}: {e}");
    }

    if config_mode && let Err(e) = create_channels_from_config(&arc_app, &sessions, &config).await {
```

muscariello (commented):
Pin looks good and I confirmed the guard still bites with it applied. One leftover on the config-mode path, inline.

crates/channel-manager/src/bin/channel_manager.rs
hackeramitkumar: done

muscariello (approved):
LGTM

Issue #2093, opened 2026-09-19
Pin the two unpinned build dependencies (Dockerfile install script, pip install)


muscariello:
OpenSSF Scorecard reports `Pinned-Dependencies` at 9/10. Two sites fetch code that is not pinned by hash:

**`Dockerfile:45` — download-then-run**

```dockerfile
curl -1sLf 'https://dl.cloudsmith.io/public/task/task/setup.deb.sh' | bash
```

A remote script piped straight into a shell during the image build. Whatever that URL serves at build time runs with full privileges in the builder. Options: pin the Taskfile release by checksum and install the binary directly, or vendor the setup script and verify its digest.

**`.github/workflows/reusable-rust-build-and-test.yaml:259` — unpinned pip install**

```bash
/tmp/ws-echo-venv/bin/pip install --quiet websockets
```

Installs whatever `websockets` resolves to at run time. Pin the version and hash, e.g. a small `requirements.txt` next to `.github/scripts/ws_echo_server.py` installed with `--require-hashes`.

Neither is exploited today, but both mean a build can execute code nobody reviewed.

PR #2099, opened 2026-09-19
fix(build): Pin the two unpinned build dependencies (Dockerfile install script, pip install)

vishalsahare:
## Summary

    - Replaced the unpinned Task Cloudsmith setup script with a direct Task v3.53.1 .deb download.
    - Added architecture-specific SHA-256 verification for amd64 and arm64 before installation.
    - Pinned the CI WebSocket test dependency to websockets==16.0.
    - Added package hashes and enabled pip’s --require-hashes mode.
    - Added the required AI change-trace entries.

    ## Validation

    ```bash
    cargo fmt --all -- --check
    cargo check -p agntcy-slim-datapath \
      -p agntcy-slim-service \ -p agntcy-slim-bindings ```

    ## Type of Change

    - [x] Bugfix
    - [ ] New Feature
    - [ ] Breaking Change
    - [ ] Refactor
    - [ ] Documentation
    - [ ] Other (please describe)

    ## Checklist


vishalsahare:
Can anyone look at this PR?

vishalsahare:
Can anyone review this PR, please?

muscariello (commented):
Checked the hashes independently rather than trusting them, and they are all correct:

- Task `amd64`/`arm64` SHA-256 match the official `task_checksums.txt` for v3.53.1, and I re-verified amd64 by downloading the `.deb` and hashing it.
- All seven `websockets==16.0` hashes are real PyPI artifacts — the `py3-none-any` wheel, the sdist, and cp310–cp314 manylinux x86_64. That is the right coverage for GitHub-hosted glibc runners; musllinux is correctly excluded.
- `dpkg --install` is safe here: the `.deb` control file declares no `Depends:`, so dropping `apt-get`'s dependency resolution costs nothing. Worth stating explicitly since it is the one behavioural change that could bite.
- It fails closed — `set -euo pipefail` plus `sha256sum --check` aborts the build on a mismatch.

**The one thing I would address: CI never builds this.** `docker-build` in `ci.yaml` is gated on `github.event_name == 'push' && github.ref == 'refs/heads/main'`, so "Build SLIM docker image" reports *skipping* on this PR. The Dockerfile rewrite therefore merges unverified, and if anything is wrong it breaks the image build on main rather than here. Worth a local `docker build` before merging, or running the docker build once against this branch.

Two small things:

`sha256sum --check --status` suppresses all output, so a hash mismatch fails with no explanation in the log. Dropping `--status` makes it print which file failed, which is worth the one line when this fires.

Pinning means Task is now frozen at 3.53.1 — the Cloudsmith script used to float. That is the point of #2093, but nothing will bump it now, so it wants a renovate rule or it silently ages.

Also, the description mentions "the required AI change-trace entries" — there is no such convention in this repo and nothing matching in the diff, so that line looks like template residue.

muscariello (commented):
One inline suggestion on top of the earlier review — everything else there is prose feedback, not something to fix in the diff.

Dockerfile
```diff
@@ -42,8 +43,26 @@ curl -L -o /tmp/llvm.sh https://apt.llvm.org/llvm.sh
...
+        echo "Unsupported platform: ${TARGETPLATFORM}"
+        exit 1
+        ;;
+esac
+
+TASK_DEB=/tmp/task.deb
+curl --fail --location --silent --show-error \
+    --output "${TASK_DEB}" \
+    "https://github.com/go-task/task/releases/download/v${TASK_VERSION}/task_${TASK_VERSION}_linux_${TARGETARCH}.deb"
+echo "${TASK_SHA256}  ${TASK_DEB}" | sha256sum --check --status
```
muscariello, line 63: `--status` suppresses all output, so a hash mismatch fails the build with nothing in the log saying which file or why. Dropping it costs one line and saves the next person a bisect.

```suggestion
echo "${TASK_SHA256}  ${TASK_DEB}" | sha256sum --check
```

muscariello:
Built this locally against the branch head since CI does not exercise the Dockerfile on PRs (`docker-build` only runs on push to `main`).

`docker build -f Dockerfile --target rust --build-arg TARGETARCH=arm64 --no-cache .` — full success, no cache reuse. The log confirms the new path actually ran: `.deb` downloaded, checksum verified (build would have aborted under `set -euo pipefail` otherwise), `dpkg --install` unpacked and set up `task (3.53.1)`, and the workspace compiled through to a finished image.

So the Dockerfile change is verified working on arm64. Combined with the earlier review (hashes independently verified, `dpkg` has no unmet deps since Task ships no `Depends:`), I don't see a blocker left. Still worth someone confirming amd64 before merge, since I only had arm64 hardware to test on.

Issue #2112, opened 2026-09-20
slim: stop publishing the old OCI path and repoint the docs


muscariello:
Part of #2103 (Phase 4). Do this last.

- Stop publishing `ghcr.io/agntcy/slim/helm/<chart>`
- Repoint every documented install at `ghcr.io/agntcy/slim-staging/helm-charts/<chart>`: `docs/content/slim/deploy/kubernetes.md` (lines 17-19, 67, 89, 116, 172) and `docs/content/slim/deploy/multicluster.md` (96, 102)
- `README.md:327` links `./charts/slim` and `./charts/slim-control-plane`, which will no longer exist — point at `slim-staging`
- `charts/slim-control-plane/README.md:8` documents `oci://ghcr.io/agntcy/helm/slim-control-plane`, a path that 403s and appears never to have existed; it goes with the chart

Keep the old path publishing until this lands, so `helm install` keeps working for anyone following the current docs.

muscariello:
Rescoping: this is now down to one item, covered by #2121.

Most of what this issue described no longer applies. The published path stayed `ghcr.io/agntcy/slim/helm/<chart>` rather than moving, so the 14 install commands across `kubernetes.md` and `multicluster.md` are all still correct and need no repointing. `README.md:327` was handled in #2118, and `charts/slim-control-plane/README.md` went with the charts. slim stopped publishing the moment #2118 removed the Helm workflows, so "stop publishing the old path" happened implicitly — and we are deliberately keeping that path alive, published from slim-staging.

The only genuine leftover was `docs/content/slim/components/data-plane/install.md:72`, which linked into `charts/` at a pinned tag. #2121 repoints it.

Issue #2110, opened 2026-09-20
slim-staging: move the charts in and publish them as OCI


muscariello:
Part of #2103 (Phase 2).

Copy the four charts from `agntcy/slim` — `slim`, `slim-control-plane`, `slim-channel-manager` and `slim-spire` — into `slim-staging`, and add a Helm release workflow publishing them to `ghcr.io/agntcy/slim-staging/helm-charts/<chart>`.

Then point each `applications/<app>/dev/config.json` at the new coordinates and re-verify that the charts still render (`task deps:vuln:images:list` exercises the whole path).

`slim-spire` comes too. It is a fork of upstream `spiffe/helm-charts-hardened`, vendoring `spire-server`, `spire-agent`, `spire-lib` and the CSI drivers as `file://` subcharts, so it moves as-is rather than being replaced by an upstream dependency.

This is the gate for Phase 4 — nothing is deleted from `slim` until the charts publish from here and #2107 is green.

muscariello:
Done. All four charts live in `agntcy/slim-staging/charts/`, the release workflow there publishes them, and the path is proven end to end — `slim-spire` v2.0.1 published to `ghcr.io/agntcy/slim/helm/slim-spire` and pulls cleanly.

One change from the original plan: the published coordinates stayed `ghcr.io/agntcy/slim/helm/<chart>` rather than moving to a slim-staging path, so every documented `helm install` keeps working and "staging" does not appear in a production install path. The `config.json` files therefore needed no change.

Issue #2103, opened 2026-09-20
Move the Helm charts to slim-staging and publish them from there


muscariello:
Move the Helm charts out of the `slim` repo into `agntcy/slim-staging`, and publish them from there, so `slim-staging` is a complete GitOps deliverable the community can fork.

## Today

slim treats Helm charts as a separately released product: four charts under `charts/` (`slim` 2.2.0, `slim-control-plane` 2.2.0, `slim-channel-manager` 2.1.0, `slim-spire` 2.0.0), each independently semver'd with its own `CHANGELOG.md`, released by pushing `helm-<chart>-v<x>` tags, across four dedicated workflows (`helm-charts.yaml`, `release-helm.yaml`, `reusable-helm-lint-test.yaml`, `reusable-helm-release.yaml`), published to `oci://ghcr.io/agntcy/slim/helm/<chart>`.

`agntcy/slim-staging` now exists and holds the ArgoCD deployment: an ApplicationSet over the four components, sync-waved spire → control-plane → data plane → channel-manager. It is the template for a private upstream GitOps repo.

## Where this is going

All four charts, `slim-spire` included, move to `slim-staging`. `slim-staging` publishes them as OCI to `ghcr.io/agntcy/slim-staging/helm-charts/<chart>` and consumes them by coordinate, exactly as it does today. The `slim` repo ends up with no charts and no Helm workflows.

Direct `helm install` has to keep working: `docs/content/slim/deploy/kubernetes.md` and `multicluster.md` both tell users to install from `oci://ghcr.io/agntcy/slim/helm/<chart>`. The old path keeps publishing until the docs are repointed.

The phases below are ordered so nothing is removed from `slim` until the charts are published from their new home and the deployment is green against them.

## Phase 1 — slim-staging exists

- [x] org-admin: repo created with rulesets and `slim-ops` access (#2105)
- [x] GitOps layout, consuming the charts published from `slim` today (#2106)
- [x] Deployment guide (#2108)

## Phase 2 — charts move and publish from slim-staging

- [x] Copy the four charts into `slim-staging`, `slim-spire` included
- [x] Add a Helm release workflow publishing to `ghcr.io/agntcy/slim-staging/helm-charts/<chart>`
- [x] Point the ApplicationSet's `config.json` files at the new coordinates and re-verify

## Phase 3 — prove it

- [ ] ArgoCD syncs a working deployment from `slim-staging` on the new coordinates (#2107)

## Phase 4 — remove from slim

[...]

PR #2117, opened 2026-09-21
ci(helm): publish charts to the helm-charts OCI path

### LINKED ISSUE #2109: slim: publish charts to ghcr.io/agntcy/slim/helm-charts alongside the current path
Part of #2103 (Phase 3).

Publish the charts to `ghcr.io/agntcy/slim/helm-charts/<chart>`, versioned by the component release tag, matching `dir`'s `reusable-release-helm.yaml`.

Publish to the **new path alongside the existing** `ghcr.io/agntcy/helm/<chart>`. Anyone running `helm install ... oci://ghcr.io/agntcy/helm/slim` today must keep working until Phase 4 retires the old path.

This is the only step in the plan that can quietly break external users, so the dual publishing is not optional.

muscariello:
Publishes each chart to `oci://ghcr.io/agntcy/slim/helm-charts/<chart>` in addition to the current `oci://ghcr.io/agntcy/slim/helm/<chart>`, aligning the path with `dir`. Part of #2103, closes #2109.

Both paths publish until #2110 switches slim-staging over, so nothing pulling the current path breaks. #2112 drops the old one.

Worth noting: the real path today is `ghcr.io/agntcy/slim/helm/<chart>`, not the `ghcr.io/agntcy/helm/<chart>` that `charts/slim-control-plane/README.md:8` documents — that one 403s and appears never to have existed. Fixing the docs belongs to #2112.


muscariello:
Closing — superseded. The charts are moving out of this repo into `agntcy/slim-staging` and will be published as OCI from there, so adding a second publishing path here is work in the wrong direction. Tracked in #2103.

